PR #107, opened 2013-10-14
Add XDMCP support as a server/manager

MartinBriza:
Very basic support, works with Xephyr -query

There are possible code style issues, point them out to me, please.

### DIFF
--- a/data/sddm.conf.in
+++ b/data/sddm.conf.in
@@ -89,3 +89,6 @@ MinimumVT=7
 # Valid values: on|off|none
 # If property is set to none, numlock won't be changed
 Numlock=none
+
+# Enables the XDMCP Server
+XDMCPServer=false
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -20,6 +20,9 @@ set(DAEMON_SOURCES
     daemon/Session.cpp
     daemon/SignalHandler.cpp
     daemon/SocketServer.cpp
+    daemon/xdmcp/Packet.cpp
+    daemon/xdmcp/Server.cpp
+    daemon/xdmcp/Utils.cpp
 )
 
 if(USE_QT5)
--- a/src/common/Configuration.cpp
+++ b/src/common/Configuration.cpp
@@ -63,6 +63,7 @@ namespace SDDM {
         bool autoRelogin { false };
 
         Configuration::NumState numlock { Configuration::NUM_NONE };
+        bool xdmcpServerEnabled { false };
     };
 
     Configuration::Configuration(const QString &configPath, QObject *parent) : QObject(parent), d(new ConfigurationPrivate()) {
@@ -122,6 +123,7 @@ namespace SDDM {
         } else {
             d->numlock = Configuration::NUM_NONE;
         }
+        d->xdmcpServerEnabled = settings.value("XDMCPServer", d->xdmcpServerEnabled).toBool();
     }
 
     void Configuration::save() {
@@ -158,6 +160,8 @@ namespace SDDM {
             settings.setValue("Numlock", "on");
         else if (d->numlock == NUM_SET_OFF)
             settings.setValue("Numlock", "off");
+
+        settings.setValue("XDMCPServer", d->xdmcpServerEnabled);
     }
 
     Configuration *Configuration::instance() {
@@ -261,4 +265,9 @@ namespace SDDM {
     const Configuration::NumState Configuration::numlock() const {
         return d->numlock;
     }
+
+    bool Configuration::xdmcpServerEnabled() const {
+        return d->xdmcpServerEnabled;
+    }
+
 }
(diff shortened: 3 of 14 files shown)


plfiorini:
Let's close this until a rebased version is available, anyway I honestly do not like to have XDMCP support.

MartinBriza:
Yes, it's for the best, I think - until somebody comes pushing to do this.
This patch wasn't very good anyway, in my opinion.

2015-06-14 15:33 GMT+02:00 Pier Luigi Fiorini <notifications@github.com>:

emrecio:
What's the status of XDMCP for SDDM, or do I just need to switch DM's?

mlybarger:
i have sddm installed and have been searching for xdmcp usage.  finally landed me here.  no discussion on why sddm doesn't want to support xdmcp?

gocpp:
Would it be possible to re-open this discussion?

For us the exclusion of XDMCP is a deal-breaker. I simply can't understand why a feature-complete desktop environment like KDE would choose SDDM without XDMCP support... Until something better comes along XDMCP is the best way for us to offer graphical terminal services; it's mature, robust and reliable as well as working with minimal configuration.

After years and years of using KDE we've all but abandoned it as we don't want to mix and match at every level of the operating system.

If XDMCP support could be added to SDDM we would be eternally grateful...

frispete:
@plfiorini The reason, that only just a few people asking for XDMCP support is very simple: they simply install kdm and be done. Unfortunately, forces appeared, that try to get rid of Qt4/KDE4 in some advanced distributions (e.g. openSUSE Tumbleweed), which leads to the awful situation, where we need to use gdm as display manager, since no decent Qt5 based display manager is available for us anymore.

You seem to be uncomfortable with that extension, and obviously never used e.g. VNC to log into a remote system, using a display manager. It would be as simple as enabling the Xorg vnc extension, and suppling a vnc password, if only sddm would cooperate. 

Interestingly, wayland is lagging still, and AFAICT, concepts for remote desktop support are still in flux, so there will be another couple of years to pass, where X persists.

Again. With the decay of kdm, we are going to be forced to use gdm to log into Plasma. I'm sorry to say that, and it is not meant to slag you off, but your personal reluctance is hurting the whole KDE eco system to some extend.

I can understand, that you don't want your code to be polluted with concepts from the past decade, but generally refusing such code is not playing with with something, that's called reality. So, if you want to see your code running on enterprise desktops/terminal server solutions, the first step would be a statement, that you consider to accept such code, if merely technical issues could be finally redressed.

emrecio:
yep, I gave up on it a long time ago.

lewellyn:
FWIW, on distros where KDE 4 is going away (such as Tumbleweed) and also therefore kdm, and you need/want XDMCP but not gdm, there's also lightdm available. It has much lighter dependencies than gdm, though the dependency chain on openSUSE (at least, but probably other distros too) will *suggest* gdm (because of the dependency on gdmflexiserver). It works just fine with gdm uninstalled.

I admit lightdm also isn't a perfect option, but the maintained choices for XDMCP support seem to be gdm, lightdm, and xdm (believe it or not) these days. As @plfiorini is against XDMCP in sddm, that leaves no real choice but to pick your poison between "it supports almost anything you can think of, at the expense of pulling in a lot of GNOME", "I hope Canonical doesn't change this thing's configuration format again", and "this looked state of the art in X11R3!" :(

plfiorini:
As I have said elsewhere, if many people need XDMCP, I'm willing to accept it in the code base.
You should also know that we have bounties (https://www.bountysource.com/teams/sddm), a bounty for XDMCP support would attract a developer to do it.
If the patch is good enough I won't block it.

frispete:
@lewellyn Thank you for your wise words. 

@plfiorini 
> if many people need XDMCP, I'm willing to accept it in the code base.

Well, I would express it differently, but thank you for a positive sign. As said before, it simply renders sddm useless in many environments, and you will not notice other than in discussions like this.

> If the patch is good enough I won't block it.

Could you give some hints, what was wrong with the code base last time? 

Yes, I already tried to merge it, but sddm changed a lot since then... any yes, I noticed the few rough corners (error handling..), that was lacking last time, but couldn't locate a real show blocker (admitting that my C++ foo is a bit rusty at best).

dflogeras:
I also would be happy to see this progress.  I've been using lightdm since KDM was phased out of Gentoo, but would be nice to have an option in SDDM.

I'm willing to spend some time on this, I'm a moderately experienced C++ dev however I've got no insight into how a login manager, and/or XDMCP is designed and actually works under the hood.  IMHO the path of least resistance would be if an experienced SDDM dev could at least outline what needs to be done to bring Martin's patch up to date so it will merge, as well as what structural changes are necessary (what prevented it from being merged the first time).

bitstreamout:
Just stumble over missing XDMCP support in sddm ... this is not very pleasant.  I'd like to open on vt8 witth the help of an script which simply does a query via  XDMCP but the sddm within the LAN (not WAN) does simply nothing.  And this I do not understand ... a simply Xaccess file together with the XDMCP protocol is secure and much faster than using an ssh or stunnel

PhilZ-cwm6:
Across LAN, XDMCP is most light and efficient way to go and even more secure once tunneled through ssh
I also had to switch from sddm to use on terminals

PR #136, opened 2014-03-31
Userconfig

Teknocrat:
Userconfigurable userAuthFile  and clientLogFile variables


plfiorini:
@jleclanche Do we still want this? I don't see a valid reason for changing the defaults since they are pretty standard among other login managers.

jleclanche:
Yeah, we still want this to unclutter $HOME. Mostly for xsession-errors; the .Xauthority file is another beast, but I think it's generally good design not to hardcode any path.

plfiorini:
This doesn't apply cleanly, closing...
I will send a new PR.

PR #148, opened 2014-04-18
Add support for PID-file support

GinoM:
Ok this version of pid-support seems to work ok.. although honestly i'm not sure how much deleting the pid file really matters..


majewsky:
This PID file does not appear to serve any actual purpose. The purpose of a PID file is to detect and prevent multiple instances of the same program running concurrently. But in this case, a second instance of SDDM would just overwrite the PID file of the first one. There are no failure cases in there.

Of course, if SDDM uses some other check to detect concurrent instances of itself, then my point is moot. But then again, I would contest the usefulness of a PID file altogether in that case.

eliasp:
@majewsky I agree. PID files are an anachronistic and broken workaround for a problem which is better solved differently.

plfiorini:
Ok rejected

PR #153, opened 2014-04-27
Wayland port

plfiorini:
Details in the commit message.

NOW OPEN FOR TESTING.

It would have been much cooler to user systemd user session, but systemctl --user starts services with a clean environment hence weston doesn't know about the session created by pam_systemd and it cannot continue.

I tried putting weston as a service in default.target.wants but the user session is started before pam_systemd with the same result.

So I use sddm-display-server as a session command, it gets launched by sddm-helper after it opens a sddm greeter session. sddm-display-server runs weston and the greeter and that works.

However I get some errors on VMware when a weston session for plfiorini is launched by the greeter, maybe there are some issues with logind having two weston instances running at the same time with two different users. The error is "[drm:vmw_master_check] *ERROR* Dropped master trying to access ioctl that requires authentication."

To fix this I suppose I should kill sddm-display-server before running the sddm-helper process that performs authentication and run the plfiorini session but that means we are unable to show authentication errors in the greeter because it gets restarted.

Alternative: sddm-display-server only runs weston and it notifies systemd with sd_notify as soon as a Wayland connection can be established. sddm-display-server opens the pam session reusing some of the sddm-helper code. WaylandDisplayServer runs "systemctl --user start sddm-greeter-wayland@0"

-------------------------------------------

OK, it seems that systemctl can be used. Not how I thought but it's possible to have a service that spawns a user session specifying user name and pam service and have other services depend on it, like a weston server that depends on the session and a greeter service that depends on weston.

This doesn't solve the error when two westons are running at the same time but makes things look much nicer (no sddm-display-server).

I'm tempted to make Display a base class and have two implementations: a traditional one like on master and a dedicated class for systemd.
Remove Greeter and have WaylandDisplayServer that runs systemctl start sddm-greeter-session.service and calls stop when asked to stop.
In this scenario sddm-helper is called twice: the first time without a command just to check the password, if the authentication goes well sddm-greeter-session.service is stopped and then sddm-helper it's called a second time with the session command.

-------------------------------------------

Remember that qtwaylandscanner is not yet supported by qtchooser so you might need to build with:

cmake -DCMAKE_INSTALL_PREFIX=/usr -DUSE_QT5=ON -DUSE_WAYLAND=ON -DQTWAYLAND_SCANNER_EXECUTABLE=/usr/lib/qt/bin/qtwaylandscanner ..

mesa > 10.0 make Qt applications on Wayland crash, turns out that this is happening on VMware too, if you are using Arch Linux downgrade your packages: http://www.maui-project.org/news/2014/04/02/how-to-fix-mesa-on-arch-linux/


plfiorini:
https://github.com/sddm/sddm/issues/419 tracks what needs to be done to make Wayland sessions run.
Another issue will be open to track activity on other stuff like making the greeter run on Wayland too.
This PR will now be closed and kept as a reference.

Issue #265, opened 2014-08-13
Dual monitors recognised as single giant screen


qjqqyy:
SDDMappears to not recognise the 2 physical Xmonitors and display things as if it is a single big screen.
The login prompt is placed right in the middle.
theme: breeze

lnxslck:
I use 2 monitors and i can't reproduce this bug. To me SDDM appears in 2 different monitors

brenns10:
I can reproduce it with the breeze login screen, but not with other login screens.  I'm thinking it might be an issue with that theme.

lnxslck:
@brenns10 I also have the breeze login screen

tbleiker:
Same here. With dual monitor and breeze theme, the login screen is placed in the middle of / between the two screens.

ava1ar:
Used to work good, but now also shows the login screen improperly. May be this is theme related, but not only breeze theme shows this weird behavior.

Darktori:
I also see the login screen right in the middle. Which is bad, because the other "screen" is my TV plugged in using HDMI, and always disabled in KDE settings when not used.

eliasp:
For me, the results vary wildly.
One time I get a clone on both screens, where the clone on the larger screen is a horribly scaled up version of the smaller screen.
Next time, it's a clone too, but this time with the proper size/resolution on the larger screen and the smaller screen showing the same picture, but with the not-fitting content moved off the screen.
And I don't know what kind of scenarios I've seen beyond that - I should probably start taking pictures whenever I reboot my laptop every 2 weeks or so ;)

CommanderAlchemy:
I'm also using breeze theme and the login screen creates one big across all my 3 screens. Then the splash screen happens individually on each screen stuttering when showing up and going away.

flyingDavid:
The issue with the breeze theme should be fixed in Plasma-workspace 5.3, there should be a login prompt on every screen.
See also: http://quickgit.kde.org/?p=plasma-workspace.git&a=commit&h=1985e8bf6bbc6044bd5ffac3c1b501536f2d1dd2

This workaround could be revoked as soon as sddm reaches version 1.0 and is widely spread.

eliasp:
Looks now like this here:
![sddm dual-screen, dual-panel :)](https://cloud.githubusercontent.com/assets/48491/7106740/1cf262f2-e14c-11e4-84fa-1bf5dc6a2dfb.jpg).

flyingDavid:
[...]

PR #268, opened 2014-08-17
Create a window per screen

plfiorini:
A big window for all screens may hit hardware limits and it often has
a bad alignment as reported by #265.

Creating a window per screen doesn't suffer from these problems and,
as a nice side effects, it helps the Wayland port.

[ChangeLog][Greeter][Behavioral Change] Create a window per screen.

### DIFF
--- a/src/greeter/GreeterApp.cpp
+++ b/src/greeter/GreeterApp.cpp
@@ -1,4 +1,5 @@
 /***************************************************************************
+* Copyright (c) 2015 Pier Luigi Fiorini <pierluigi.fiorini@gmail.com>
 * Copyright (c) 2013 Abdurrahman AVCI <abdurrahmanavci@gmail.com>
 *
 * This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


plfiorini:
Resuming my work on this soon...

plfiorini:
Updated... It no longer remove the screenModel let's see about that in a follow up patch.
The way it is done now it shouldn't change much for clone mode (which is what I get here).
I'm afraid clone mode might be default in the X world, so if this patch doesn't show any regression I'd like to merge it and take another look with Wayland.

davidedmundson:
Did you have a chance to check this hasn't broken breeze before pushing?

src/greeter/GreeterApp.cpp
```diff
@@ -134,35 +129,74 @@ namespace SDDM {
...
+        connect(view, &QQuickView::screenChanged, this, [view, this](QScreen *screen) {
+            if (screen == Q_NULLPTR)
+                removeViewForScreen(view);
+        });
+#endif
+
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 4, 0))
+        // always resize when the screen geometry changes
+        connect(screen, &QScreen::availableGeometryChanged, this, [view](const QRect &r) {
+            view->setGeometry(QRect(QPoint(0, 0), r.size()));
```
davidedmundson: this isn't right. views geometry is global to the screen (in the X sense of the word) not to the the montitor.
If screen 2 size changes, you'll put it on top of screen 1.

view->setGeometry(screen->geometry()); will do you best

plfiorini:
@davidedmundson yep breeze look fine here too and yes i won't remove the screen model for a while, in fact this PR reintroduces it.

src/greeter/GreeterApp.cpp
plfiorini: ops you are right, don't know why i left it this way

PR #274, opened 2014-08-29
Add config parameter to disable avatars

ypnos:
On startup, the greeter accesses *all* home directories to check for the existence of a `.face.icon` file. If a home directory sits on a remote machine, this can significantly slow down greeter startup; furthermore, if a remote machine is not available, the greeter sits through the default automounter timeout. The behavior is especially undesirable when a theme is used that does not use face files anyway.

1. The current behavior can make SDDM unfit for client systems in larger networks with distributed home directories.
2. As there is no output at all, graphical or in the log, during that period, it is hard to track down the cause of a seemingly "broken" SDDM.

This patch enables the user to change the default behavior of checking for `.face.icon` files in home directory. When opting out via the ReadHomeFaces setting (default: true), SDDM only supports files in the dedicated FacesDir location (which, for example, could be populated via the LDAP jpegPhoto property). Also, the config description mentions this as a possible cause for slow startup.

### DIFF
--- a/data/man/sddm.conf.rst.in
+++ b/data/man/sddm.conf.rst.in
@@ -123,6 +123,12 @@ OPTIONS
 	wont be updated.
 	Default value is true.
 
+`ReadHomeFaces=`
+	If this flag is true, home directories of all users are
+	searched for .face.icon files on startup. This can lead to
+	a slowdown in large NFS environments.
+	Default value is true.
+
 [Autologin] section:
 
 `User=`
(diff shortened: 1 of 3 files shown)


jleclanche:
So here's the plan:

* Add an "EnableAvatars" configuration item
* Allow themes to check for that config item and not show them at all
 * Otherwise, they default to the default `user.face.icon` and `root.face.icon`.

@ypnos are you willing to update your PR to reflect this?

ypnos:
I'm sorry I'm not interested anymore.

jleclanche:
@ypnos Ok, I'll be picking up the PR then, closing this.

PR #319, opened 2014-11-25
Show battery status

Meumeu:
Expose battery info to QML
Display battery level in maui if a battery is present
Add power backends at runtime if they are started after sddm


jleclanche:
@plfiorini @davidedmundson I am catching this super late but on the LXQt side we'd be ok with pulling in Solid. Would that make sense?

jleclanche:
Solid has been approved to use in sddm. @Meumeu I don't suppose you'd like to work on that?

PR #331, opened 2014-12-21
Add rounded corner support to TextBox and PasswordBox

jparvela:
Although the widgets used inside TextBox and PasswordBox support rounded corners, these widgets themselves do not provide it.

By adding the radius property to the TextBox and PasswordBox, these widgets become capable of supporting rounded corners, which are a simple nice ability to give to themes.


plfiorini:
Squashed and merged

PR #332, opened 2014-12-21
Make PasswordBox follow user set height.

jparvela:
The widgets in PasswordBox have a fixed height in them. This means that if some theme applies a custom height to PasswordBox, it is ignored, and the default 30px is used anyway.

This commit makes it so the widgets inside PasswordBox inherit the container's size, so when a custom height is applied, it is followed.

---

I have tested the default themes included and they are unaffected by this change.


jleclanche:
GTM

plfiorini:
Merged

PR #337, opened 2015-01-03
clarify syntax of HideUsers & HideShells in man

jrowens:
Explain that the lists of users/shells should be comma-separated. When I was editing my sddm.conf, I found this quite unobvious, especially since it appeared directly under the colon-separated DefaultPath. I'd also suggest changing the default sddm.conf to reflect this, but that is added by the Fedora packager, in my case. I'll file there, too.

### DIFF
--- a/data/man/sddm.conf.rst.in
+++ b/data/man/sddm.conf.rst.in
@@ -104,11 +104,13 @@ OPTIONS
 	Default value is 65000.
 
 `HideUsers=`
-	Users that shouldn't show up in the user list.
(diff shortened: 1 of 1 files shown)


jrowens:
I should add, my knowledge of C++ & Qt is quite limited; the comma separation is my best guess from a bit of reading about QStringList. If it should be something else, change it appropriately, of course.

plfiorini:
Seems already fixed by 9a9bdd4f355b93e2acccd4c7dfe57f5a631fe5df

PR #340, opened 2015-01-14
Install 'sddm.service' in 'graphical.target'.

eliasp:
sddm didn't start for me anymore on bootup - the explanation was actually rather straight-forward.
I wonder how we missed this so far ;)

### DIFF
--- a/services/sddm.service.in
+++ b/services/sddm.service.in
@@ -11,4 +11,4 @@ Restart=always
 
 [Install]
 Alias=display-manager.service
-
+WantedBy=graphical.target


plfiorini:
Weird... :-/

plfiorini:
Here's a possible explaination: graphical.target already requires display-manager.service so if you had set graphical.target to be the default target and sddm.service was enabled it would start.
I don't know what's changed in systemd. Since when and which systemd version did you have this problem?

eliasp:
@plfiorini Running systemd from git and update usually once or twice a day.
I think I had this problem since around 2.5 weeks, but can't tell definitely.

Will have a look whether I can identify a systemd commit responsible for this.

eliasp:
Most likely caused by:
[8fa844dccff70824ff063bfe5f409c47f5699d73](http://cgit.freedesktop.org/systemd/systemd/commit/?id=8fa844dccff70824ff063bfe5f409c47f5699d73) - units: make graphical.target dependencies more complete and similar to those of multi-user.target

plfiorini:
That commit could be the one causing the problem.
I don't expect your patch to cause issues with older systemd versions but I would like to test it just in case...

jleclanche:
Hold.

plfiorini:
Jerome asked Lennart in person over IRC.
Could be a systemd regression but he needs more information.

[13:55:21] <poettering> jleclanche: well, dunno, logs when booted with systemd.log_level=debug would be good...

eliasp:
Will provide logs with `systemd.log_level=debug` later today. Can't reboot this system right now.

plfiorini:
@eliasp ping :)

plfiorini:
Should we close this?

jleclanche:
Well we still need to figure out the situation i think... but we won't be fixing anything on our side.

Issue #341, opened 2015-01-21
Should support TryExec in session files


martenjj:
Related to issue #305 ("Can't use all available sessions").

Some distros that I have encountered have desktop files for all possible desktops/WMs installed in /usr/share/xsessions, relying on the display manager to filter out the unavailable ones by checking for a TryExec= entry in each file.  SDDM should do this too.

martenjj:
Suggested patch:

diff --git a/src/greeter/SessionModel.cpp b/src/greeter/SessionModel.cpp
index a451b54..81964a2 100644
--- a/src/greeter/SessionModel.cpp
+++ b/src/greeter/SessionModel.cpp
@@ -57,6 +57,7 @@ namespace SDDM {
                 continue;
             SessionPtr si { new Session { session, "", "", "" } };
             QTextStream in(&inputFile);
+            bool execAllowed= true;
             while (!in.atEnd()) {
                 QString line = in.readLine();
                 if (line.startsWith("Name="))
@@ -65,9 +66,15 @@ namespace SDDM {
                     si->exec = line.mid(5);
                 if (line.startsWith("Comment="))
                     si->comment = line.mid(8);
+                if (line.startsWith("TryExec=")) {
+                    QFileInfo fi(line.mid(8));
+                    if (!fi.exists() || !fi.isExecutable())
+                        execAllowed = false;
+                }
             }
             // add to sessions list
-            d->sessions.push_back(si);
+            if (execAllowed)
+                d->sessions.push_back(si);
             // close file
             inputFile.close();
         }

martenjj:
Suggested patch:

<pre>
diff --git a/src/greeter/SessionModel.cpp b/src/greeter/SessionModel.cpp
index a451b54..81964a2 100644
--- a/src/greeter/SessionModel.cpp
+++ b/src/greeter/SessionModel.cpp
@@ -57,6 +57,7 @@ namespace SDDM {
                 continue;
             SessionPtr si { new Session { session, "", "", "" } };
             QTextStream in(&inputFile);
+            bool execAllowed= true;
             while (!in.atEnd()) {
                 QString line = in.readLine();
                 if (line.startsWith("Name="))
@@ -65,9 +66,15 @@ namespace SDDM {
                     si->exec = line.mid(5);
                 if (line.startsWith("Comment="))
                     si->comment = line.mid(8);
+                if (line.startsWith("TryExec=")) {
+                    QFileInfo fi(line.mid(8));
[...]

PR #345, opened 2015-01-23
Honour TryExec in X session desktop files

martenjj:
No point in showing a session as available if it could not be started.  As described in GitHub issue #341.


davidedmundson:
Looks good, thanks.
I'll merge once CI says it's OK.

PR #359, opened 2015-02-06
 	Add Hungarian translation

mortein79:
Just add Hungarian translation.


plfiorini:
Merged in a single commit

PR #367, opened 2015-02-12
Update it.ts

ilpianista:
Better Italian translation for "Login failed"


jleclanche:
gtm, also @plfiorini maybe we should give @ilpianista commit rights

Issue #375, opened 2015-02-22
X Sessions with TryExec not showing up in session list


mystal:
PR #345 filters sessions out when the binary specified in TryExec does not exist. However, some X session files (like [i3.desktop](https://gist.github.com/mystal/e7045159a09c54f18c45)) do not specify an absolute path, just the binary name. This is legal according to the [desktop entry spec](http://standards.freedesktop.org/desktop-entry-spec/latest/ar01s05.html), and the binary should be looked up in the $PATH environment variable.

The fix should be pretty straightforward. I'll probably give it a shot once I have some free time unless someone else gets to it first.

PR #382, opened 2015-03-12
Check for TryExec in $PATH if it is not absolute

mystal:
This should fix issue #375. I've tested it locally on Arch Linux and it indeed shows the i3.desktop session once more.


plfiorini:
Merged

PR #386, opened 2015-03-19
Improve Russian translation

aspotashev:
1. Use action verbs for buttons,
2. Use a better word to translate "session",
etc...


plfiorini:
Can't really comment on the translation because I don't speak Russian so I trust you :)

PR #389, opened 2015-03-23
Use X_PREFIX rather than hardcoded /usr.

trombonehero:
On the BSDs, non-base-system binaries are found in `/usr/local/bin`
and most header files are found in `/usr/local/include`. CMake exposes
this platform detail as `${CMAKE_INSTALL_PREFIX}`, so let's use it to make
SDDM a bit more cross-platform.

### DIFF
--- a/src/common/Configuration.h
+++ b/src/common/Configuration.h
@@ -53,10 +53,10 @@ namespace SDDM {
         );
         // TODO: Not absolutely sure if everything belongs here. Xsessions, VT and probably some more seem universal
         Section(XDisplay,
-            Entry(ServerPath,          QString,     _S("/usr/bin/X"),                           _S("X server path"));
-            Entry(XephyrPath,          QString,     _S("/usr/bin/Xephyr"),                      _S("Xephyr path"));
-            Entry(XauthPath,           QString,     _S("/usr/bin/xauth"),                       _S("Xauth path"));
-            Entry(SessionDir,          QString,     _S("/usr/share/xsessions"),                 _S("Session description directory"));
+            Entry(ServerPath,          QString,     _S(PREFIX "/bin/X"),                           _S("X server path"));
+            Entry(XephyrPath,          QString,     _S(PREFIX "/bin/Xephyr"),                      _S("Xephyr path"));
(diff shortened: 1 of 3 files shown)


src/common/Configuration.h
```diff
@@ -53,10 +53,10 @@ namespace SDDM {
         );
         // TODO: Not absolutely sure if everything belongs here. Xsessions, VT and probably some more seem universal
         Section(XDisplay,
-            Entry(ServerPath,          QString,     _S("/usr/bin/X"),                           _S("X server path"));
-            Entry(XephyrPath,          QString,     _S("/usr/bin/Xephyr"),                      _S("Xephyr path"));
-            Entry(XauthPath,           QString,     _S("/usr/bin/xauth"),                       _S("Xauth path"));
-            Entry(SessionDir,          QString,     _S("/usr/share/xsessions"),                 _S("Session description directory"));
+            Entry(ServerPath,          QString,     _S(PREFIX "/bin/X"),                           _S("X server path"));
```
davidedmundson: won't this mean if a user installs a self compiled SDDM, they can't use it with a system provided X?

trombonehero: You are right, yes: a non-default CMAKE_INSTALL_PREFIX would've broken the search for `xauth`, `xsessions`, etc. I've changed the patch to explicitly look for the `share/xsessions` and use that as an X prefix.

trombonehero:
Ping?

plfiorini:
-1

trombonehero:
When I build on Arch Linux, I see:

```shell
[arch build]$ grep X_PREFIX src/common/*
#define X_PREFIX            "/usr"
```

so it seems to find everything in the correct place. Please note that, as of commit 0f7819b, I use the searched-for `X_PREFIX` rather than `CMAKE_INSTALL_PATH`.

jleclanche:
This isn't the correct approach to fix this. I recommend the maintainer ships a custom configuration file that matches the layout of the distro. If this were to be fixed in sddm, it'd have to be some CMake option for X binaries prefix.

Closing.

plfiorini:
Agree :+1:

Issue #393, opened 2015-04-02
nvidia-prime support at login


jriddell:
Nvidia Prime is a funky setup for users who feel the need for both NVidia and Intel GPUs on their computer.  This needs a program to be run as root at login and logout.  Here's a patch I'm using in Kubuntu to run the program at login.

This would be more elegant with cascading config support as requested in https://github.com/sddm/sddm/issues/217 then the nvidia-prime package could just set a config value like lightdm has:
"display-setup-script is run after the X server starts but before the user session / greeter is run. "

jriddell:
Index: sddm-0.11.0/data/scripts/Xsetup
===================================================================
--- sddm-0.11.0.orig/data/scripts/Xsetup
+++ sddm-0.11.0/data/scripts/Xsetup
@@ -1,3 +1,7 @@
 #!/bin/sh
 # Xsetup - run as root before the login dialog appears
 
+if [ -e /sbin/prime-offload ]; then
+    echo running NVIDIA Prime setup /sbin/prime-offload, you will need to manually run /sbin/prime-switch to shut down
+    /sbin/prime-offload
+fi

tseliot:
We have "display-setup-script" (which seems to be the same as the "DisplayCommand" entry in sddm), and "display-stopped-script" to execute a script immediately after X is stopped. I've just written some code that should add the latter (it's more of a copy and paste).

plfiorini:
Maybe this is something that we would add by default.

@jriddell Is there any change between distros on how NVIDIA Prime is handled?

PR #395, opened 2015-04-02
Add support for the DisplayStopCommand entry

tseliot:
The DisplayStopCommand entry allows executing a script right after stopping X.

This is relevant to the following feature request:
https://github.com/sddm/sddm/issues/393

### DIFF
--- a/src/daemon/XorgDisplayServer.cpp
+++ b/src/daemon/XorgDisplayServer.cpp
@@ -250,6 +250,14 @@ namespace SDDM {
         qDebug() << "Running display stop script " << displayStopCommand;
         displayStopScript->start(displayStopCommand);
 
+        // wait for finished
+        if (!displayStopScript->waitForFinished(5000))
+            displayStopScript->kill();
+
+        // clean up the script process
+        displayStopScript->deleteLater();
+        displayStopScript = nullptr;
+
         // clean up
         process->deleteLater();
         process = nullptr;
@@ -275,6 +283,9 @@ namespace SDDM {
         env.insert("XAUTHORITY", m_authPath);
         env.insert("SHELL", "/bin/sh");
         displayScript->setProcessEnvironment(env);
+
+        // delete displayScript on finish
+        connect(displayScript, SIGNAL(finished(int,QProcess::ExitStatus)), displayScript, SLOT(deleteLater()));
 
         // start display setup script
         qDebug() << "Running display setup script " << displayCommand;
--- a/src/daemon/XorgDisplayServer.cpp
+++ b/src/daemon/XorgDisplayServer.cpp
@@ -233,6 +233,23 @@ namespace SDDM {
         // log message
         qDebug() << "Display server stopped.";
 
+        QString displayStopCommand = mainConfig.XDisplay.DisplayStopCommand.get();
+
+        // create display setup script process
+        QProcess *displayStopScript = new QProcess();
+
+        // set process environment
+        QProcessEnvironment env;
+        env.insert("DISPLAY", m_display);
+        env.insert("HOME", "/");
+        env.insert("PATH", mainConfig.Users.DefaultPath.get());
+        env.insert("SHELL", "/bin/sh");
+        displayStopScript->setProcessEnvironment(env);
+
+        // start display setup script
+        qDebug() << "Running display stop script " << displayStopCommand;
+        displayStopScript->start(displayStopCommand);
+
         // clean up
         process->deleteLater();
         process = nullptr;
--- a/data/man/sddm.conf.rst.in
+++ b/data/man/sddm.conf.rst.in
@@ -85,6 +85,10 @@ OPTIONS
 	Path of script to execute when starting the display server.
 	Default value is "@DATA_INSTALL_DIR@/scripts/Xsetup".
 
+`DisplayStopCommand=`
+	Path of script to execute when stopping the display server.
+	Default value is "@DATA_INSTALL_DIR@/scripts/Xsetup".
+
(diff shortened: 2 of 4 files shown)


src/daemon/XorgDisplayServer.cpp
```diff
@@ -233,6 +233,23 @@ namespace SDDM {
         // log message
         qDebug() << "Display server stopped.";
 
+        QString displayStopCommand = mainConfig.XDisplay.DisplayStopCommand.get();
+
+        // create display setup script process
+        QProcess *displayStopScript = new QProcess();
```
davidedmundson: leaks.

src/daemon/XorgDisplayServer.cpp
```diff
@@ -233,6 +233,23 @@ namespace SDDM {
...
+        QProcessEnvironment env;
+        env.insert("DISPLAY", m_display);
+        env.insert("HOME", "/");
+        env.insert("PATH", mainConfig.Users.DefaultPath.get());
+        env.insert("SHELL", "/bin/sh");
+        displayStopScript->setProcessEnvironment(env);
+
+        // start display setup script
+        qDebug() << "Running display stop script " << displayStopCommand;
+        displayStopScript->start(displayStopCommand);
```
davidedmundson: should we block for it to finish?

davidedmundson:
Thanks for looking into this.

src/daemon/XorgDisplayServer.cpp
tseliot: On 02-04-15 08:55:15, davidedmundson wrote:
>    In [1]src/daemon/XorgDisplayServer.cpp:
> > @@ -233,6 +233,23 @@ namespace SDDM {
> >          // log message
> >          qDebug() << "Display server stopped.";
> >
> > +        QString displayStopCommand = mainConfig.XDisplay.DisplayStopCommand.g
> et();
> > +
> > +        // create display setup script process
> > +        QProcess *displayStopScript = new QProcess();
>
>    leaks.
>

Then we have two leaks, as XorgDisplayServer::setupDisplay() does the
same.

--
Alberto Milone

src/daemon/XorgDisplayServer.cpp
tseliot: On 02-04-15 08:56:57, davidedmundson wrote:
>    In [1]src/daemon/XorgDisplayServer.cpp:
> > +        QString displayStopCommand = mainConfig.XDisplay.DisplayStopCommand.g
> et();
> > +
> > +        // create display setup script process
> > +        QProcess *displayStopScript = new QProcess();
> > +
> > +        // set process environment
> > +        QProcessEnvironment env;
> > +        env.insert("DISPLAY", m_display);
> > +        env.insert("HOME", "/");
> > +        env.insert("PATH", mainConfig.Users.DefaultPath.get());
> > +        env.insert("SHELL", "/bin/sh");
> > +        displayStopScript->setProcessEnvironment(env);
> > +
> > +        // start display setup script
> > +        qDebug() << "Running display stop script " << displayStopCommand;
> > +        displayStopScript->start(displayStopCommand);
>
>    should we block for it to finish?

Yes, that would be a good idea. It's been a while since I last used Qt
and QProcess.

--
Alberto Milone

tseliot: How about something like this?

```patch
From 067beb73065383dd557643e39d7fc859f713a418 Mon Sep 17 00:00:00 2001
From: Alberto Milone <alberto.milone@canonical.com>
Date: Thu, 2 Apr 2015 18:24:18 +0200
Subject: [PATCH 1/1] src/daemon/XorgDisplayServer.cpp: wait for
 DisplayStopCommand to finish

Also get rid of two memory leaks
---
 src/daemon/XorgDisplayServer.cpp | 12 ++++++++++++
 1 file changed, 12 insertions(+)

diff --git a/src/daemon/XorgDisplayServer.cpp b/src/daemon/XorgDisplayServer.cpp
index 4f75be9..8fbfc1f 100644
--- a/src/daemon/XorgDisplayServer.cpp
+++ b/src/daemon/XorgDisplayServer.cpp
@@ -250,6 +250,14 @@ namespace SDDM {
         qDebug() << "Running display stop script " << displayStopCommand;
         displayStopScript->start(displayStopCommand);
 
+        // wait for finished
+        if (!displayStopScript->waitForFinished(5000))
+            displayStopScript->kill();
+
+        // clean up the script process
+        displayStopScript->deleteLater();
+        displayStopScript = nullptr;
+
         // clean up
         process->deleteLater();
         process = nullptr;
@@ -279,6 +287,10 @@ namespace SDDM {
         // start display setup script
         qDebug() << "Running display setup script " << displayCommand;
         displayScript->start(displayCommand);
+
+        // clean up the script process
+        displayScript->deleteLater();
+        displayScript = nullptr;
     }
 
     void XorgDisplayServer::changeOwner(const QString &fileName) {
-- 
1.9.1
```

davidedmundson: that won't work because then you're deleting the displayScript startup script potentially before it's finished.

in this case it kinda leaked for a reason... though a terrible one.

you'd need something like
connect(process, finished, process, deleteLater())); 

or use QProcess::startDetatched which does it all internally.

tseliot:
I got the code to work exactly the way I wanted in the last commit. I connected the signal as you suggested in order to free displayScript. Furthermore we now wait for displayStopScript to finish. As a result, I can easily switch between GPUs in Ubuntu now.

tseliot:
Thanks a lot!

Issue #371, opened 2015-02-16
helper quits silently, when sddm user homedir isn't present


mathieujobin:
Hello,

I am running Kubuntu 15.04 since Alpha 1 at the end of december... SDDM was shipped by default and working... eventually I got a black screen, I believe after some package update, but it could be something else.

investigating the logs, I noticed this error... (flagged as a warning...)

```
[01:12:50.139] (WW) DAEMON: Auth: sddm-helper exited with 2
```

I tried to find it in the source, where this could happen, and I found that
Auth::HELPER_SESSION_ERROR has 2 for value.

so, I am thinking this is where it ends....

https://github.com/sddm/sddm/blob/master/src/helper/HelperApp.cpp#L136

am I right thinking so, or there is something else I am missing?

here is the full log snippet for that run

```
[01:12:49.803] (II) DAEMON: Initializing...
[01:12:49.807] (II) DAEMON: Starting...
[01:12:49.807] (II) DAEMON: Adding new display 0 on vt 1 ...
[01:12:49.807] (II) DAEMON: Adding cookie to "/var/run/sddm/:0"
[01:12:49.810] (II) DAEMON: Display server starting...
[01:12:49.810] (II) DAEMON: Running: /usr/bin/X :0 -auth /var/run/sddm/:0 -nolisten tcp -background none -noreset vt1
[01:12:50.117] (WW) DAEMON: Signal received: SIGUSR1
[01:12:50.117] (II) DAEMON: Running display setup script  "/usr/share/sddm/scripts/Xsetup"
[01:12:50.117] (II) DAEMON: Display server started.
[01:12:50.117] (II) DAEMON: Socket server starting...
[01:12:50.117] (II) DAEMON: Socket server started.
[01:12:50.120] (II) DAEMON: Greeter starting...
[01:12:50.127] (II) HELPER: [PAM] Starting...
[01:12:50.127] (II) HELPER: [PAM] Authenticating...
[01:12:50.127] (II) HELPER: [PAM] returning.
[01:12:50.137] (II) DAEMON: Greeter session started successfully
[01:12:50.138] (II) HELPER: [PAM] Ended.
[01:12:50.139] (WW) DAEMON: Auth: sddm-helper exited with 2
[01:12:50.139] (II) DAEMON: Greeter stopped.

```

I am not sure what could possibly be wrong? missing PAM settings?

thanks

velocity303:
[...]

PR #396, opened 2015-04-04
add comments, should sddm user be hardcoded ?

mathieujobin:
Questions in relation with issue #371 

I wonder if those 3 occurrence of *sddm* should be hardcoded like that?
or fetch from config?

also, getpwnam is probably inexpensive but we are doing it 5 times.
should it get encapsulated?

### DIFF
--- a/src/common/Configuration.h
+++ b/src/common/Configuration.h
@@ -81,6 +81,7 @@ namespace SDDM {
         );
     );
 
+                                                 // is this a default value?
     Config(StateConfig, []()->QString{auto tmp = getpwnam("sddm"); return tmp ? tmp->pw_dir : STATE_DIR;}().append("/state.conf"),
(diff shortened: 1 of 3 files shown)


davidedmundson:
I'm not happy with flooding code with basically feature requests.

If something should be changed, change it.

davidedmundson:
To answer your question about getpwnam.

The memory you get back from getpwnam is owned by that library, and the next call to getpwanything will re-use that same memory.

It's a bit weird.

So the only way to cache the result is to then do a deep copy of the contents on the first call; do-able but not as trivial as things first look.

I think it probably would be "better", but optimising without profiling can mean you over complicate a lot of code with no tangible benefit.

mathieujobin:
thanks for taking the time to read this and reply

PR #397, opened 2015-04-04
should there be a problem with the sddm user, be more verbose instead of normal exit

mathieujobin:
fix in relation to issue #371 

print verbose messages in case there is a  problem with sddm user.
also, do failure exit instead of normal exit

cheers,

### DIFF
--- a/src/helper/UserSession.cpp
+++ b/src/helper/UserSession.cpp
@@ -61,21 +61,26 @@ namespace SDDM {
         return m_path;
     }
 
-    void UserSession::bail(int status) {
-        emit finished(status, QProcess::NormalExit);
-        exit(status);
-    }
-
     void UserSession::setupChildProcess() {
-        struct passwd *pw = getpwnam(qobject_cast<HelperApp*>(parent())->user().toLocal8Bit());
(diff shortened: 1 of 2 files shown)


src/helper/UserSession.cpp
```diff
@@ -61,21 +61,28 @@ namespace SDDM {
...
-        struct passwd *pw = getpwnam(qobject_cast<HelperApp*>(parent())->user().toLocal8Bit());
-        if (setgid(pw->pw_gid) != 0)
-            bail(2);
-        if (initgroups(pw->pw_name, pw->pw_gid) != 0)
-            bail(2);
-        if (setuid(pw->pw_uid) != 0)
-            bail(2);
-        if (chdir(pw->pw_dir) != 0)
-            bail(2);
+        //TODO: why do we need to fetch the info from the passwd file?
```
davidedmundson: Remove this line.
Code is not a bug tracker

davidedmundson:
Looks good. Make that one change I want and I'll merge this.

Thanks.

mathieujobin:
i removed that comment and fix a syntax error I had.

davidedmundson:
Thanks very much.

PR #399, opened 2015-04-10
Install scripts/Xstop now that the script is available

tseliot:
This actually installs the Xstop script that is now part of sddm as per #395. Sorry I didn't spot it before.


tseliot:
Thanks again!

PR #403, opened 2015-04-16
allow changing the dbus config file name

apachelogger:
on some distributions the generic name of the dbus config might be provided
by more than one display manager, so always installing with the generic
name would require distributions to hard-patch the cmake code.
allowing to change it through a cmake cache variable enables distributions
to simply parameterize in their cmake call.

this for example affects Ubuntu where the config would be provided by both
SDDM and LightDM.


plfiorini:
lgtm

PR #411, opened 2015-04-30
pbis-open support and manually login

Hundefutter:
It would be great if sddm would support pbis-open. Also a field to login manually when pbis-open would be supported. Because if there a hundreds of persons it would be terrible to search a personal name.

And also a gui for plasma to configure all this. Last but not least a scrollbar when there a multiple users infront of sddm. It is not funny when you can see four users and when you don't know that you must scroll to see further user logins.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,7 +71,7 @@ else()
 endif()
 
 # systemd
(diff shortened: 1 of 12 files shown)


plfiorini:
what is this? i see a lot of old commits

Issue #419, opened 2015-05-17
Wayland phase 1 (User session)


plfiorini:
Wayland support phase 1.
The greeter keeps running on X11 but we can list Wayland sessions and start them.

Old code to be used as a reference: #153

Phase 2 is #426

- [x] Add wayland-session, wayland-start and wayland-setup scripts
- [x] List Wayland session from the greeter
- [x] Pass session directory from greeter to daemon so that the correct session entry can be picked up
- [x] Set xdg env vars correctly
- [x] Don't call xauth if user session is wayland
- [x] Do not run Wayland sessions with Xsession (run through wayland-session)
- [x] Save user session stderr to .wayland-errors
- [x] Do not set DISPLAY for Wayland sessions
- [x] Do not set PAM_XDISPLAY and PAM_XAUTHDATA
- [x] Set up a new vt for Wayland sessions
- [x] Helper/backend: Set PAM_TTY to VT for Wayland sessions
- [x] Helper: Take control of tty
- [x] Better acquire/release VT
- [x] Fix build errors with Qt 5.6 (dev branch)

plfiorini:
Fork with this work: https://github.com/plfiorini/sddm/tree/wayland-session

PR #427, opened 2015-06-17
Read session desktop entries with Session and pass it upon login

plfiorini:
The greeter knows where the session entry is located (either under the
X11 or Wayland sessions path) and so it must pass this information
to the daemon.

Add a class to represent a session in order to share this information
between daemon and greater allowing to know the exact location of
the session desktop entry.

Since a relative file name can be specified for autologin session,
we need to search the desktop entry and create a Session object
with the right type.

In the long run we might want to deal with absolute desktop entry
paths and do not allow relative paths or names without extention.

Issue: #419

### DIFF
--- a/src/common/Session.cpp
+++ b/src/common/Session.cpp
@@ -1,0 +1,162 @@
+/***************************************************************************
+* Copyright (c) 2015 Pier Luigi Fiorini <pierluigi.fiorini@gmail.com>
+* Copyright (c) 2013 Abdurrahman AVCI <abdurrahmanavci@gmail.com>
+*
+* This program is free software; you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation; either version 2 of the License, or
+* (at your option) any later version.
+*
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with this program; if not, write to the
+* Free Software Foundation, Inc.,
+* 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+***************************************************************************/
+
+#include <QFile>
+#include <QFileInfo>
+#include <QTextStream>
+
+#include "Configuration.h"
+#include "Session.h"
+
+const QString s_entryExtention = QStringLiteral(".desktop");
+
+namespace SDDM {
+    Session::Session()
+        : m_valid(false)
+        , m_type(UnknownSession)
+    {
+    }
+
+    Session::Session(Type type, const QString &fileName)
+    {
+        setTo(type, fileName);
+    }
+
+    bool Session::isValid() const
+    {
+        return m_valid;
+    }
+
+    Session::Type Session::type() const
+    {
+        return m_type;
+    }
+
+    QString Session::xdgSessionType() const
+    {
+        return m_xdgSessionType;
+    }
+
+    QDir Session::directory() const
+    {
+        return m_dir;
+    }
+
+    QString Session::fileName() const
+    {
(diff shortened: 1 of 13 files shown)


plfiorini:
Since it's not a super small commit I'd like a quick review @jleclanche @davidedmundson in case I made some obvious mistake

plfiorini:
BTW, I plan to merge this commit on Sat

jleclanche:
couple of nits but gtm overall

src/common/Session.h
```diff
@@ -0,0 +1,96 @@
...
+
+        QString desktopNames() const;
+
+        void setTo(Type type, const QString &name);
+
+        Session &operator=(const Session &other);
+
+    private:
+        bool m_valid;
+        Type m_type;
```
davidedmundson: not initialised

src/common/Session.cpp
```diff
@@ -0,0 +1,158 @@
...
+    {
+        return m_tryExec;
+    }
+
+    QString Session::desktopNames() const
+    {
+        return m_desktopNames;
+    }
+
+    void Session::setTo(Type type, const QString &name)
```
davidedmundson: so the api is that the name argument should include ".desktop" , but ::name()  won't?

I'd call the arg fileName to be less confusing.

src/common/Session.cpp
```diff
@@ -0,0 +1,158 @@
...
+        case WaylandSession:
+            m_dir = QDir(mainConfig.WaylandDisplay.SessionDir.get());
+            m_xdgSessionType = QStringLiteral("wayland");
+            break;
+        default:
+            m_xdgSessionType = QStringLiteral("");
+            break;
+        }
+
+        m_name = m_name.replace(QRegExp(".desktop$"), QLatin1String(""));
```
davidedmundson: merge with line 100

src/common/Session.h
```diff
@@ -0,0 +1,96 @@
...
+        QString m_comment;
+        QString m_exec;
+        QString m_tryExec;
+        QString m_xdgSessionType;
+        QString m_desktopNames;
+
+        friend class SessionModel;
+    };
+
+    typedef QSharedPointer<Session> SessionPtr;
```
davidedmundson: I don't get where this is used, especially as you have a &operator=

If you do want this, I'd QSharedDataPointer and make it implicit (like QString)

src/daemon/Display.cpp
```diff
@@ -121,9 +121,21 @@ namespace SDDM {
...
+            // determine session type
+            const QString &autologinSession = mainConfig.Autologin.Session.get();
+            Session session;
+            if (findSessionEntry(mainConfig.XDisplay.SessionDir.get(), autologinSession)) {
+                session.setTo(Session::X11Session, autologinSession);
+            } else if (findSessionEntry(mainConfig.WaylandDisplay.SessionDir.get(), autologinSession)) {
+                session.setTo(Session::WaylandSession, autologinSession);
+            } else {
+                qCritical() << "Unable to find autologin session entry" << autologinSession;
+                return;
```
davidedmundson: you don't want to return here, we need login to attempt and fail so that we hanlde it and show the greeter the second time.

This version will just hang (I think)

davidedmundson:
Looks good, is there an email on fd.o about agreeing on /usr/share/wayland-sessions  ?

plfiorini:
@davidedmundson No there's no fd.o email about /usr/share/wayland-sessions to my knowledge, on the other hand this is what GDM do and I don't see any harm in calling it this way

davidedmundson:
Ah cool, that's what I wanted to hear.

src/daemon/Display.cpp
plfiorini: in that case the Session will be invalid and we won't have an actual command to execute (Exec line was not read). maybe i can just emit loginFailed() before return

src/common/Session.h
plfiorini: this was used by SessionModel i just preserved, i think we can just do a new Session instead of a shared pointer.

PR #428, opened 2015-06-19
Pulles in LC_TIME env variable on greeter start

farseerfc:
User may set different LC_TIME apart from LANG, like myself, 
for different datetime format shown in the greeter.
In my case, I am a Chinese speaker living in Japan, 
therefore I have LANG=zh_CN while LC_TIME=ja_JP.
Currently the lock screen of KDE5 set the LC_TIME correctly, 
but the login greeter use only LANG.
This change makes lock screen and login greeter behave the same.


plfiorini:
lgtm

Issue #390, opened 2015-03-23
XorgDisplayServer::addCookie() called before XorgDisplayServer::m_display set


trombonehero:
Hi,

I'm attempting to run SDDM on FreeBSD, and I'm running into an xauth error.

In `XorgDisplayServer::start()` (`src/daemon/XorgDisplayServer.cpp:119`), `addCookie` is called (line 125) before `m_dispaly` is set (line 199). Thus, I see error messages such as:

```console
/usr/local/bin/xauth: (stdin):1:  bad "remove" command line
/usr/local/bin/xauth: (stdin):2:  bad "add" command line
```

Is `m_display` supposed to be set somewhere else before `start()` is called?

trombonehero:
It looks like this isn't a FreeBSD-specific thing: I get the same behaviour on a fresh Arch Linux install.

eliasp:
Saw the same issue several times on my Gentoo box, but then it suddenly/magically worked again.

plfiorini:
#431 fixes the issue here

PR #431, opened 2015-06-22
Create X authorization after the Xorg server is started

plfiorini:
Otherwise m_display is not yet set.

Issue: #390


davidedmundson:
+2

PR #433, opened 2015-06-24
wait for the display script to complete

plfiorini:
Received this from OpenMandriva: https://abf.io/openmandriva/sddm/blob/master/sddm-wait-for-display-script.patch

Not super sure about -1, I get it that we can't run the process async but waiting forever is probably not a good idea.

Thoughs @davidedmundson ?


plfiorini:
OpenMandriva needs -1 because they run programs that require user input and that of course can easily take more than 30s.

So we should really have a timeout here and OMDV will patch with -1

PR #437, opened 2015-07-05
Use Qt 5.4.1 with Travis CI

plfiorini:
Build with a more recent Qt version, this should avoid some
build errors.


jleclanche:
This isn't the way to go anyway as we just discussed on irc. Closing.

PR #438, opened 2015-07-10
Add ConsoleKit2 as a PowerManager backend

EricKoegel:
This optionally adds ConsoleKit2 as a PowerManager backend (via an ENABLE_CONSOLEKIT2 cmake option). It uses the same DBUS API as logind and turns it into a shared seat manager backend.


a17r:
Thanks, I've successfully tested your patches (together with a patched Xsession that adds ck-launch-session).

plfiorini:
At a first glance the code looks good, now the challenge is to test it myself.

plfiorini:
Merged

PR #442, opened 2015-07-17
Basic fix for RTL layout and Update Arabic translation

SafaAlfulaij:
First fix for RTL layout.
There are some issues left that I have to take care of.
I used a workaround to not require Qt5.5 for the RTL layout to take effect.

### DIFF
--- a/data/themes/circles/Main.qml
+++ b/data/themes/circles/Main.qml
@@ -30,6 +30,9 @@ Rectangle {
     width: 640
     height: 480
 
+    LayoutMirroring.enabled: Qt.locale().textDirection == Qt.RightToLeft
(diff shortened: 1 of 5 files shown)


plfiorini:
Ops... I didn't see it before committing what I did today.

I will try to use:

```
Qt.locale().textDirection == Qt.RightToLeft
```

Instead of having to translate QT_LAYOUT_DIRECTION (as the documentation says, although it's ugly).

plfiorini:
Using textDirection seems fine, however it's the ImageButton component that needed the mirror property not its instances.

I'm merging this keeping only the translation update and slightly change the commit title to reflect the fact that.

PR #454, opened 2015-08-17
Handle QtMsgType in Qt5.5 or later correctly

kenya888:
Hi guys:)

As you all know QtInfoMsg was introduced in Qt 5.5. This change stops warnings while compiling as following.

/home/kenya888/devel/sddm/src/common/MessageHandler.h:40:16: warning: enumeration value 'QtInfoMsg' not handled in switch [-Wswitch]
/home/kenya888/devel/sddm/src/common/MessageHandler.h:81:16: warning: enumeration value 'QtInfoMsg' not handled in switch [-Wswitch]

And I think it would be better adding default behavior for new QtMsgType being added in the future.

### DIFF
--- a/src/common/MessageHandler.h
+++ b/src/common/MessageHandler.h
@@ -41,6 +41,11 @@ namespace SDDM {
             case QtDebugMsg:
                 priority = LOG_DEBUG;
             break;
+// QtInfoMsg was added in Qt 5.5
+#if QT_VERSION >= 0x050500
+            case QtInfoMsg:
+            break;
+#endif
             case QtWarningMsg:
                 priority = LOG_WARNING;
             break;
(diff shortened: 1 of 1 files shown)


jleclanche:
The default case is enough. If you want to specifically handle `QtInfoMsg`, you can implement it.

kenya888:
@jleclanche Thanks for your review/comment. At present I have no thought about handling QtInfoMsg especially. I agree all with you.

@peremen Thanks for your review/comment one-by-one. I agree all with you. As you say default case is enough to remove code to handle QtInfoMsg.

I'll update PR for your feedback. thanks:)

plfiorini:
Could you please squash the commits into one?

kenya888:
@plfiorini Oops, I squashed changes and send new PR...

https://github.com/sddm/sddm/pull/456

PR #457, opened 2015-08-17
Honor the screen available geometry

plfiorini:
Using 0,0 coordinates was wrong and caused additional windows to always
cover the first one when a screen was resized.


davidedmundson:
+2

jleclanche:
gtm

Issue #325, opened 2014-12-04
SDDM always starts X with '-nolisten tcp'


RAMeijer:
I would like to use SDDM but had to switch back to LightDM as I could not find a way to turn off the option '-nolisten tcp' and I need to be able to connect to my X from remote machines...

The only places I can find any reference to that option are the binary itself and the log file:

  [root@x-ray ~]# locate sddm|xargs grep -lr nolisten
  /usr/bin/sddm
  /var/log/sddm.log

Am I missing something?

phavx:
Not a developer of sddm, but a quick grep throught the sources for "nolisten" only finds on instance:

> src/daemon/XorgDisplayServer.cpp:                 << "-nolisten" << "tcp"

So a patch is easy, though a compile time option or ideally a config setting would be better, imho.

jleclanche:
We need to get the xorg startup arguments from sddm.conf. I think this should be in 1.0 tbh, but it'll depend if anyone wants to pick it up. Marking

BETLOG:
I've just spent several hours trying to figure out why my desktop wont allow me to '-listen tcp' ...and which files even determine this. But I am no closer to a solution. Did someone revert something in Kubuntu 18.04?
```
uname -a
Linux betlogbrick 4.15.0-70-generic #79-Ubuntu SMP Tue Nov 12 10:36:11 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

PR #459, opened 2015-08-19
Add X args

bjacquot:
add the possibility of adding parameters to the X server


jleclanche:
What do you need this for, specifically?

bjacquot:
I have a multiseat configuration for which I would manage only a seat with sddm. So I have to add an argument "-seat xxx"

bjacquot:
Ok, done. Maybe the argument ServerArguments must be initialized with "-nolisten tcp -background none -noreset" and avoid them in XorgDisplayServer.cpp.

plfiorini:
@jleclanche This might be needed to close #325

plfiorini:
Squashed and merged

PR #461, opened 2015-08-20
Wip/xorgargs

plfiorini:


jleclanche:
you know, we really need to generate at least a commented-out sddm.conf. someone who wants to *add* an argument will have to re-add `nolisten tcp`...

jleclanche:
gtm

plfiorini:
@jleclanche yep, distros could generate a sddm.conf with sddm --example-config but they need a previous sddm build. Having a sddm.conf.in template with defaults expanded by CMake could work, those defaults would be used by the C++ class as well.

Issue #460, opened 2015-08-20
Honor /etc/locale.conf


plfiorini:
Honor the locale from /etc/locale.conf if locale environment variables are not set (LC_*) and /etc/locale.conf exists and contains a LANG variable.

Issue #364, opened 2015-02-09
Desktop doesn't load if $HOME containts non-ASCII character(s)


davispuh:
If I've a user with `$HOME` like `/home/Dāvis/` then after typing password I get black screen with visible cursor but without desktop - it doesn't load.

```
sddm-helper[1253]: Starting: "/usr/share/sddm/scripts/Xsession" "/usr/bin/startkde"
sddm-helper[1258]: Could not open stderr to .xsession-errors file
sddm-helper[1258]: Adding cookie to "/home/D?vis/.Xauthority"
sddm[1223]: Session started
sddm-helper[1253]: [PAM] Ended.
sddm[1223]: Auth: sddm-helper exited with 1
systemd-logind[324]: Removed session c15.
```

It works fine with KDM.

mathieujobin:
this could be very much caused by Qt toLocal8Bit() being used in helper/UserSession.cpp
https://github.com/sddm/sddm/blob/master/src/helper/UserSession.cpp#L65

mathieujobin:
@davispuh note that my system gives a warning when I create an accout with your username

```
mathieu@mathieu-MSI-X6:~$ sudo adduser Dāvis
[sudo] password for mathieu: 
adduser: To avoid problems, the username should consist only of
letters, digits, underscores, periods, at signs and dashes, and not start with
a dash (as defined by IEEE Std 1003.1-2001). For compatibility with Samba
machine accounts $ is also supported at the end of the username
mathieu@mathieu-MSI-X6:~$ 


```

davispuh:
That's not surprising, for me on Arch Linux it complains only about capital letters, `dāvis` don't give any warnings while `Davis` does.
Anyway to be POSIX compliant username should match`^[a-z][-a-z0-9]*$` but technically there aren't any reason why couldn't use any other characters/format. Main reason is for compatibility with other applications which could break with unexpected usernames. I'm intentionality creating it such as just personal preference and have been using it for years without any problems until now SDDM :D For regular home users I don't think anyone really cares about POSIX compliance (I don't) so I would recommend to try work with non-compliant usernames whenever it's possible.

davispuh:
[...]

PR #466, opened 2015-08-21
Implement /etc/locale.conf support.

### LINKED ISSUE #365: Can't login if username contains non-ASCII characters(s)
Username such as `dāvis` can't login. After typing password it shows `Login failed`.
```
sddm-helper[13097]: [PAM] Starting...
sddm-helper[13097]: [PAM] Authenticating...
sddm-helper[13097]: [PAM] Preparing to converse...
sddm-helper[13097]: [PAM] Conversation with 1 messages
sddm-helper[13097]: pam_tally(sddm:auth): pam_get_uid; no such user
sddm-helper[13097]: pam_unix(sddm:auth): check pass; user unknown
sddm-helper[13097]: pam_unix(sddm:auth): authentication failure; logname= uid=0 euid=0 tty= ruser= rhost=
sddm-helper[13097]: [PAM] authenticate: Authentication failure
sddm-helper[13097]: [PAM] returning.
sddm[13078]: Authentication error: "Authentication failure"
sddm-greeter[13087]: Message received from daemon: LoginFailed
sddm-helper[13097]: [PAM] Ended.
sddm[13078]: Auth: sddm-helper exited with 1
```

works with KDM.

davispuh:
This implements #460 (and reverts afd6d1c82f) and fixes #365, #364

### DIFF
--- a/src/auth/Auth.cpp
+++ b/src/auth/Auth.cpp
@@ -113,7 +113,22 @@ namespace SDDM {
             , id(lastId++) {
         SocketServer::instance()->helpers[id] = this;
         QProcessEnvironment env = child->processEnvironment();
-        env.insert("LANG", "C");
+        bool langEmpty = true;
+        QFile localeFile("/etc/locale.conf");
+        if (localeFile.open(QIODevice::ReadOnly | QIODevice::Text)) {
+            QTextStream in(&localeFile);
+            while (!in.atEnd()) {
+                QStringList parts = in.readLine().split('=');
+                if (parts.size() >= 2) {
+                    env.insert(parts[0], parts[1]);
+                    if (parts[0] == QStringLiteral("LANG"))
+                        langEmpty = false;
+                }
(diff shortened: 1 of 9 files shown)


davispuh:
Updated with 2 more commits so that user Icon works too.
Qt uses `LC_*` for locale and if they're set they take precedence over anything else
but sddm would set them to empty string if they weren't set before and that would ruin Qt system's locale causing use of wrong encoding.
For proper i18n support need to ensure that `LANG` is set (from /etc/locale.conf) and then always use QString::fromLocal8Bit for all strings from POSIX API.
I would recommend enabling QT_NO_CAST_FROM_ASCII and don't use `QString(char *)` but for POSIX API use fromLocal8Bit and for other APIs use respective encoding for them.

plfiorini:
:+1: 

Going to merge, I will just make a change to the last commit message adding your notes

Issue #469, opened 2015-08-22
Do not use QString(char*)


plfiorini:
As suggested by @davispuh we should enable QT_NO_CAST_FROM_ASCII and replace all QString(char*) instances using fromLocal8Bit with POSIX API and QStringLiteral for the rest or other encodings.

majewsky:
This can be closed then.

PR #471, opened 2015-08-22
enable QT_NO_CAST_FROM_ASCII

majewsky:
To facilitate that, sprinkle a bunch of QStringLiteral() around string
literals and QLatin1Char() around char literals. In some places, I used
QLatin1String() instead of QStringLiteral() when the expression is used
as an argument to a function that has a QLatin1String overload (this is
advised by the Qt docs).

I also replaced empty strings like

    QString x = "";
    QString x { "" };
    funcWithQStringArg("");
    funcWithQStringArg(QString(""));

by the QString() default constructor, which yields an empty string more
efficiently:

    QString x;
    funcWithQStringArg(QString());

QString::fromLocal8Bit() was used whenever strings were read from C
libraries (e.g. PAM). For SDDM's own configuration files, I used
QString::fromUtf8(), under the assumption that most text editors today
default to UTF-8.

In some places, I also used the chance to optimize single-char string
literals to char literals, e.g.

    str << list.join(",");              //before
    str << list.join(QLatin1Char(',')); //after

**Testing done: It compiles and the ConfigurationTest passes. I don't have a test setup for actually running a compiled SDDM, so if someone could check that I didn't break anything, that would be highly appreciated.**

Closes: #469

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,7 +33,7 @@ option(BUILD_MAN_PAGES "Build man pages" OFF)
 option(ENABLE_JOURNALD "Enable logging to journald" ON)
 
 # Definitions
-add_definitions(-Wall -std=c++11)
+add_definitions(-Wall -std=c++11 -DQT_NO_CAST_FROM_ASCII)
 
 # Default build type
 if(NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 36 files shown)


plfiorini:
Can't log into Weston anymore :)

majewsky:
Oh dear. :)

plfiorini:
My bad, I forgot that this system needs a different pam config, now it works :P

plfiorini:
:+1:

PR #474, opened 2015-08-28
Support external on screen keyboard

hykwok:
Add a feature that allows the greeter to call external keyboard application such as onboard:
- Add a new option in the sddm configuration file to set the on screen keyboard (likes lightdm-gtk-greeter)
- Add a new function for greeter to call external OSK application.

### DIFF
--- a/data/themes/maui/Main.qml
+++ b/data/themes/maui/Main.qml
@@ -256,7 +256,19 @@ Rectangle {
 
                     onClicked: sddm.powerOff()
 
-                    KeyNavigation.backtab: btnReboot; KeyNavigation.tab: prevUser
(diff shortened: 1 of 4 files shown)


EmperorArthur:
Maybe it's just me, but there seem to be some strange spacing issues under files changed.
Also, have you considered rebasing the compiler errors into the previous commit?  It makes running 'git bisect' much easier.

Haven't looked at this project's coding guidelines, so it might be ok.

jleclanche:
You're mixing tabs and spaces everywhere, this needs to be fixed.

davidedmundson:
Qt now has a virtual keyboard built-in that can be run in-process.
I think that's the way forwards here as it removes the need for any window-manager like functionality.

Issue #465, opened 2015-08-21
Support for consolekit


tnut:
I know this feature has been already ask in the past. I agreed it's getting really obsolete. But there are still a few distributions that are not ready to move to systemd for good reasons.

 I had to give a try.  Perhaps if you have a look at my patch, you could reconsider this option. It's seems to be not to complicated to integrate it. 

sddm is really a nice project by the way, I had a chance to learn a bit about qt (still a long way to go for me) Because of my lack of knowledge the patch need a lot of FIXME. Perhaps, some qt specialists could give me some advises. 

You can find the patch at http://downloads.nutyx.org/files/patchs/sddm/sddm-0.11.0-consolekit-1.patch

Here is the interesting part of the /var/log/sddm.log:
 [11:23:48.383] (II) DAEMON: Reading from "/usr/share/xsessions/xfce.desktop"
 [11:23:48.384] (II) DAEMON: Session "/usr/share/xsessions/xfce.desktop" selected, command: "startxfce4"
 [11:23:48.397] (II) DAEMON: XDG_SESSION_COOKIE SET for ConsoleKit Session
 [11:23:48.404] (II) HELPER: [PAM] Starting...
 [11:23:48.404] (II) HELPER: [PAM] Authenticating...
 [11:23:48.404] (WW) HELPER: [PAM] authenticate: Permission denied
 [11:23:48.404] (II) HELPER: [PAM] returning.

As you can see I finally get stuck with pam which is for me another nice mistery 

Best regards

Thierry

plfiorini:
ConsoleKit has been revived by the ConsoleKit 2 project.
We are not going to support ConsoleKit. @EricKoegel do you think any of this is useful to CK2 support, or supporting the logind interface will be enough?

tnut:
Do you have a link of the consolekit2 project ?

plfiorini:
https://github.com/ConsoleKit2/ConsoleKit2

tnut:
I beleave the ck-connector API is not change much (perhaps the boolean parameter less which is not a bad idea). For me it Should works as well for consolekit2. As I said, if somebody could help me for the FIXME issue in the patch

EricKoegel:
The ck-connector API hasn't changed. A patch like this would allow distros to drop the call to ck-launch-session.
[...]

PR #476, opened 2015-08-28
Add the ConsoleKit2 pam connector to the greeter

EricKoegel:
This way a ConsoleKit2 session is setup the same way systemd-logind is.
See: https://github.com/sddm/sddm/issues/465


plfiorini:
Does it work on your system?

jleclanche:
So what do we do with this? Nobody on the team can test it.

EricKoegel:
Sorry, I don't think this pull request is enough for consolekit support. I'll open a new one when I get time to fully test it before hand.

Issue #410, opened 2015-04-29
Loading a theme using QML modules in non-standard import path fails outside of test mode


ace13:
Like the title says, Themes fail to load QML modules if they're outside of the main Qt QML import path. A simple fix for this would be to include the QML/QML2_IMPORT_PATH and LD_LIBRARY_PATH variables into the greeter environment, or using the system environment completely - like when run through --test-mode.

olesalscheider:
This bug breaks KDE's breeze theme on Exherbo.

ace13:
A way to kludge it is to edit `sddm/src/daemon/Greeter.cpp` L124-127

```c++
            // greeter environment
            QProcessEnvironment env;
            QProcessEnvironment sysenv = QProcessEnvironment::systemEnvironment();
            env.insert("LANG", sysenv.value("LANG"));
```
to
```c++
            // greeter environment
            //QProcessEnvironment env;
            QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
            //env.insert("LANG", sysenv.value("LANG"));
```

A very dirty way of solving the issue, but it "works" until there's a proper fix.

plfiorini:
This will kill all locale environment variables breaking another thing we recently fixed.
Since the variable names are known I'd say we should get QML2_IMPORT_PATH and LD_LIBRARY_PATH from sysenv and copy them over env.

ace13:
I did some testing with my current KDE5 install (living in `/opt/kf5`).
To get the Breeze team to load and display correctly - without changing anything in the filesystem - I had to copy over four environment variables to the greeter; `LD_LIBRARY_PATH`, `QML2_IMPORT_PATH`, `QT_PLUGIN_PATH`, and `XDG_DATA_DIRS`

With those four added, it works perfectly.

PR #483, opened 2015-09-06
Inherit path environment variables from parent

plfiorini:
Make themes installed in non-standard locations work.

Issue: #410

[ChangeLog][Daemon][Greater] Inherit path environment variables from
parent so that themes installed in non-standard paths will work.


jleclanche:
+0. This seems like a weird solution, but I won't block it - gtm if it's functional.

PR #488, opened 2015-09-08
Add Serbian translations (all 4 variants)

pejakm:


plfiorini:
LGTM, I'm going to merge but we'll add the following to the commit message:

[ChangeLog][Translations][Added Serbian translation]

So that it will be automatically picked up when creating the change log for the next release.

Issue #196, opened 2014-06-13
Plymouth


plfiorini:
Some notes from IRC:

```
<plfiorini> apachelogger: does plymouth have some sort of protocol that needs to be used by the DM or it's just Xorg not running with -background none and possibly other args plus vt.handoff?
<plfiorini> oh and the sddm.service needs to be after plymouth-quit.service
<apachelogger> plfiorini: the gist of integration: http://bazaar.launchpad.net/~ubuntu-branches/ubuntu/lucid/kdebase-workspace/lucid/view/head:/debian/patches/kubuntu_34_kdm_plymouth_transition.diff
<apachelogger> basically use plymouth binary to talk to daemon, query vt it is using, use that vt, tell plymouth to quit but not repaint the vt before x starts
<apachelogger> as for the systemd units: I am not sure what plymouth quit does, but on upstart from what I remember the respective DM units would call plymouth quit when the DM unit actually is not intended to start (e.g. cmdline contains text argument) and conversively the plymouth-stop upstart unit would call plymouth quit when an unsupported DM is started
<apachelogger> *not sure what plymouth-quit.service does* ^^
<apachelogger> upstart nonesense FTR http://paste.ubuntu.com/7638410/
<apachelogger> plfiorini: I am reasonable certain for the systemd stuff you can just look at any of the other major DMs, I think pretty much every distribution has plymouth transition support
<plfiorini> apachelogger: i will look at gdm code but calling plymout quit will be handled by the systemd unit and that's less code to write :P
<apachelogger> plfiorini: you still need to call quit from inside sddm
<apachelogger> plymouth quit --retain-splash
<apachelogger> that forces plymouth to not repaint the VT
<plfiorini> apachelogger: ok in fact the plymouth-quit.service doesn't call plymouth quit with --retain-splash so that might not work as expected unless it's the default now
<apachelogger> plfiorini: quit without argument will repaint the VT (i.e. will go black), while with retain it will nto repaint and simply preserve whatever the last render run looked like
<apachelogger> the problem is that if a DM doesn't implement plymouth support plymouth may run on VT1 and DM on VT2 and then you'd have the last render of plymouth hanging around on VT1, that's what plymouth-quit.service is supposed to prevent, well, and the fact that you don't want a useless plymouthd binary running all the time ^^
```

xiangzhai:
Hi @plfiorini 

[...]

Issue #78, opened 2013-07-29
Multi-seat support


lbssousa:
I would like to propose multi-seat support in SDDM in a similar way it's proposed in LightDM.

Maybe this bug should be broken in four, namely:
### 1. Basic multi-seat support (as of PR #564)
- Implement steps 3 and 4 of systemd-logind [multi-seat protocol](http://www.freedesktop.org/wiki/Software/systemd/writing-display-managers/):
  - Setting PAM variable `XDG_SEAT=SEAT_ID` is already done in SDDM.
  - Command-line option `-seat SEAT_ID` needs to be passed to Xorg.
  - If SEAT_ID is not seat0, don't allocate a VT for Xorg and don't pass "vtXX" command-line argument. Since release 1.16, non-seat0 Xorg servers don't touch VTs.
    - For Xorg servers older than 1.16, systemd multi-seat-x wrapper is required for starting Xorg on non-seat0 seats (if it's not available, start Xorg with options "-sharevts -config [non-seat0.conf.multi-seat](http://cgit.freedesktop.org/xorg/xserver/tree/config/non-seat0.conf.multi-seat?id=xorg-server-1.15.2)"). **WARNING**: multi-seat-x wrapper was dropped in systemd 217.
### 2. Automatic multi-seat support (as of [LightDM#1190581](https://bugs.launchpad.net/lightdm/+bug/1190581))
- Implement steps 1 and 2 of systemd-logind [multi-seat protocol](http://www.freedesktop.org/wiki/Software/systemd/writing-display-managers/):
  - If systemd-logind is running in the system, do the following:
    - Instantiate a QDBusInterface to `org.freedesktop.login1.Manager`.
    - Call `login1` method `ListSeats()` once to get initial seat list and call `SeatManager` method `createSeat()` for each one.
    - Connect somehow `login1` signals `SeatNew()` and `SeatRemoved()` signals to `SeatManager` slots `createSeat()` and `removeSeat()`, respectively.
    - For each seat added from logind, store its object path in a `Seat` object, in order to get its properties later (`CanGraphical`, `CanMultiSession`, etc.).
  - Otherwise, add statically a single seat named "seat0", as SDDM currently does.
### 3. Complete multi-seat support
- Prevent user switching on seats that have property `CanMultiSession=no`.
  - **CAUTION**: do NOT prevent switching to greeter, or you can have problems when a user logs out from a non-seat0 seat.
[...]

PR #494, opened 2015-09-15
Close the pipe fd passed to X server.

wengxt:
If X server fails to write the display number to the pipe (E.G. crash), sddm may freeze on the next readLine call because it is still possible to read from it, which will freeze sddm.

### DIFF
--- a/src/daemon/XorgDisplayServer.cpp
+++ b/src/daemon/XorgDisplayServer.cpp
@@ -175,6 +175,10 @@ namespace SDDM {
                 return false;
             }
 
+            // close the other side of pipe in our process, otherwise reading
+            // from it may stuck even X server exit.
+            close(pipeFds[1]);
+
             QFile readPipe;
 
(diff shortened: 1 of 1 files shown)


xiangzhai:
Hi csslayer, and [it](https://github.com/sddm/sddm/blob/master/src/daemon/XorgDisplayServer.cpp#L169) might be blocked 30000 msecs in very abnormal situitation.

wengxt:
@xiangzhai That's unrelated. This PR only tries to solve the inproper usage of pipe.

xiangzhai:
yup, but there are some issues in sddm daemon need to fix or implement:
1. plymouth smooth transition https://github.com/sddm/sddm/issues/196
2. multiseat https://github.com/sddm/sddm/issues/78

so it is better to re-architecture some code ;-)

Regards,
Leslie Zhai - [a KDE developer](https://identity.kde.org/index.php?r=people/view&uid=lesliezhai)

plfiorini:
Merged

Issue #489, opened 2015-09-10
getpwnam can fail


davispuh:
I don't really know why it failed, but it happened.

In `src/helper/UserSession.cpp`

``` c++
void UserSession::setupChildProcess() {
...
        const char  *username = qobject_cast<HelperApp*>(parent())->user().toLocal8Bit();
        struct passwd *pw = getpwnam(username);
        if (setgid(pw->pw_gid) != 0) {
```

`getpwnam` returned NULL causing segfault on `pw->pw_gid`

```
#0  0x00000000004313d1 in SDDM::UserSession::setupChildProcess (this=0x1d19220) at /mnt/GitHub/sddm/src/helper/UserSession.cpp:115
115             if (setgid(pw->pw_gid) != 0) {
#1  0x00007ff55f286db9 in ?? () from /usr/lib/libQt5Core.so.5
#2  0x00007ff55f2895cf in ?? () from /usr/lib/libQt5Core.so.5
#3  0x00007ff55f22a7c2 in QProcess::start(QString const&, QStringList const&, QFlags<QIODevice::OpenModeFlag>) () from /usr/lib/libQt5Core.so.5
#4  0x00007ff55f22aa17 in QProcess::start(QString const&, QFlags<QIODevice::OpenModeFlag>) () from /usr/lib/libQt5Core.so.5
#5  0x00000000004305ae in SDDM::UserSession::start (this=0x1d19220) at /mnt/GitHub/sddm/src/helper/UserSession.cpp:47
#6  0x000000000042bff9 in SDDM::Backend::openSession (this=0x1d19100) at /mnt/GitHub/sddm/src/helper/Backend.cpp:70
#7  0x0000000000435a49 in SDDM::PamBackend::openSession (this=0x1d19100) at /mnt/GitHub/sddm/src/helper/backend/PamBackend.cpp:276
#8  0x000000000042ded3 in SDDM::HelperApp::doAuth (this=0x7ffc16f1bca0) at /mnt/GitHub/sddm/src/helper/HelperApp.cpp:134
#9  0x0000000000437a5a in SDDM::HelperApp::qt_static_metacall (_o=0x7ffc16f1bca0, _c=QMetaObject::InvokeMetaMethod, _id=6, _a=0x7ffc16f1b590) at /mnt/GitHub/sddm/build/src/src/helper/moc_HelperApp.cpp:111
[...]

PR #503, opened 2015-09-24
Don't cast QByteArray to (char *)

davispuh:
Don't cast `QByteArray` to `char *`, they're [obsolete members](http://doc.qt.io/qt-5/qbytearray-obsolete.html) and caused #489


plfiorini:
Merged, with a better commit message that would help track the issue.

PR #511, opened 2015-10-15
A couple of multi monitor fixes

plfiorini:
Trying to fix multi monitor support.

### DIFF
--- a/src/greeter/GreeterApp.cpp
+++ b/src/greeter/GreeterApp.cpp
@@ -109,7 +109,6 @@ namespace SDDM {
         // create models
 
         m_sessionModel = new SessionModel();
-        m_screenModel = new ScreenModel();
         m_userModel = new UserModel();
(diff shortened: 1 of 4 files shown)


plfiorini:
For some reason on all my systems, mirror mode kicks in and that doesn't help testing.
For future reference, I had to do this on Xsetup:
xrandr --output LVDS1 --auto --rotate normal --pos 0x0 --output VGA1 --auto --right-of LVDS1

CMCDragonkai:
Does this show a login screen of every display? For some reason my primary screen is auto detected to not be my laptop screen, which is annoying.

plfiorini:
Primary screen is detected by Xorg, SDDM uses whatever the display server tells it to use.

CMCDragonkai:
Yea, so does it clone the login prompt to every display? That would be cool.

PR #513, opened 2015-10-16
microoptimizations

shaforostoff:
* qstring-related (qstringref, qlatin1string), see http://catchchallenger.first-world.info//wiki/Benchmark_for_conception#String
* use qvector for storing pointers, see https://marcmutz.wordpress.com/effective-qt/containers/
* use iterators with qmap to avoid duplicate lookups


plfiorini:
Merged

PR #517, opened 2015-10-20
Remove the failsafe session entry

### LINKED ISSUE #515: failsafe session fails
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=800158

Let`s review the source code

* add failsafe session https://github.com/sddm/sddm/blob/master/src/greeter/SessionModel.cpp#L133
* but failsafe session is invalid! https://github.com/sddm/sddm/blob/master/src/daemon/Display.cpp#L244
* there is ***NO*** failsafe.desktop https://github.com/sddm/sddm/blob/master/src/common/Session.cpp#L128

Regards,
Leslie Zhai

plfiorini:
The failsafe session can indeed fail if xterm is missing.
For this reason it doesn't hold much value and is a source of bugs.

Closes #515

### DIFF
--- a/data/scripts/Xsession
+++ b/data/scripts/Xsession
@@ -89,9 +89,6 @@ case $session in
   "")
     exec xmessage -center -buttons OK:0 -default OK "Sorry, $DESKTOP_SESSION is no valid session."
     ;;
-  failsafe)
-    exec xterm -geometry 80x24-0-0
-    ;;
   *)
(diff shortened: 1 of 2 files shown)


jleclanche:
gtm

davidedmundson:
+1

xiangzhai:
:+1:

PR #528, opened 2015-11-05
Improvements to faces

jleclanche:


plfiorini:
i'm not saying i'm against this just yet but could it be possible to improve performance of model population instead of adding yet another setting?

plfiorini:
like concurrent population

jleclanche:
@plfiorini See #274. I was against adding another setting, and I also think avatar loading should be done async, but I *also* think a setting to disable avatars altogether is reasonable which makes it a good compromise.

jleclanche:
BTW the idea is to make the enableAvatars config public to themes so that they can use it to not show avatars at all. This is less urgent though.

plfiorini:
ok gtm

Seegras:
Trying to read the users home dir fails miserably if its encrypted (and automounted upon login of the user). So yes, there needs to be an option to inhibit sddm from trying to read /home and everything below.

PR #529, opened 2015-11-05
Generate and install the default sddm.conf after sddm is built

jleclanche:


plfiorini:
<jleclanche> [18:43:20] <ngladitz> jleclanche: for the command itself use the target name not a path
<jleclanche> [18:43:43] <ngladitz> if the command is a target cmake will itself substitute the path and it will implicitly add the order requirement

plfiorini:
if that's the case and the patch works (i can't test myself right now) then lgtm

plfiorini:
it can easily be checked if cmake prints the command and the executable is indeed expanded with the right path

jleclanche:
Just tested by moving /usr/bin/sddm out of the path and yes it works. Landing.

Issue #14, opened 2013-02-16
SDDM doesn't start in Ubuntu


reypg87:
Hi, I have compiled sddm on Ubuntu 12.10 with qt4 and installed via make install. Then I have modified /etc/X11/default-display-manager setting /usr/bin/sddm. So when I reboot the system keeps forever on boot screen and X server crached, it doesnt find any display manager. The system boot fine by any other DM.
[   106.110](EE) intel(0): [drm] Failed to open DRM device for pci:0000:00:02.0: No such file or directory
[   106.110](EE) intel(0): Failed to become DRM master.
[   106.110](II) UnloadModule: "intel"
[   106.110](EE) Screen(s) found, but none have a usable configuration.
[   106.110]
Fatal server error:
[   106.110] no screens found
[   106.110](EE)
Please consult the The X.Org Foundation support
         at http://wiki.x.org
 for help.
[   106.110](EE) Please also check the log file at "/var/log/Xorg.0.log" for additional information.
[   106.110](EE)
[   106.129] Server terminated with error (1). Closing log file.

PR #531, opened 2015-11-05
Fall back to regular login when autologin fails

### LINKED ISSUE #316: problems with autologin
I use sddm 0.10 under Arch Linux. There was no sddm.conf file installed, so I created one according to the man page. I added
```ini
[Autologin]
User=gerhard
Session=Cinnamon
```
Using this I received the 
> sddm-helper exited with 127

which is also mentioned in another post. After some more reading I had the idea that the correct session line might be 
`Session=Cinnamon.desktop`

This hang the complete system. The error message from the log is
> Failed to find command for session: "/usr/share/xsessions/Cinnamon.desktop

Arch Linux uses systemd and pam. If there is need for some config please drop a line. Manual login works without problems so I will stick with sddm.

jleclanche:
Fixes #316

### DIFF
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -103,6 +103,29 @@ namespace SDDM {
         m_displayServer->start();
     }
 
+    bool Display::attemptAutologin() {
+        Session::Type sessionType = Session::X11Session;
+
+        // determine session type
+        const QString &autologinSession = mainConfig.Autologin.Session.get();
+        if (findSessionEntry(mainConfig.XDisplay.SessionDir.get(), autologinSession)) {
+            sessionType = Session::X11Session;
+        } else if (findSessionEntry(mainConfig.WaylandDisplay.SessionDir.get(), autologinSession)) {
+            sessionType = Session::WaylandSession;
+        } else {
+            qCritical() << "Unable to find autologin session entry" << autologinSession;
+            return false;
+        }
+
+        Session session;
+        session.setTo(sessionType, autologinSession);
+
+        m_auth->setAutologin(true);
+        startAuth(mainConfig.Autologin.User.get(), QString(), session);
+
+        return true;
+    }
+
     void Display::displayServerStarted() {
         // check flag
         if (m_started)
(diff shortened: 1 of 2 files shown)


plfiorini:
loginFailed(m_socket) sends the message to the greeter which emits loginFailed(), but apparently it doesn't go anywhere from there.
m_socket is initially nullptr, a pointer is assigned in ::login() which is invoked when the greeter sends a Login command through the socket to the daemon.

I don't really think that with autologin we'll ever receive the Login command hence m_socket is probably still a nullptr causing a crash.

plfiorini:
We need to emit loginFailed() to communicate with the greeter but I don't see why this must be the case for autologin, hence the loginFailed(m_socket) call is wrong and you got rid of it with this patch.

plfiorini:
Stack trace, courtesy of coredumpctl:

``
           PID: 14600 (sddm)
           UID: 0 (root)
           GID: 0 (root)
        Signal: 11 (SEGV)
     Timestamp: gio 2015-11-05 22:51:54 CET (3min 20s ago)
  Command Line: /usr/bin/sddm
    Executable: /usr/bin/sddm
 Control Group: /system.slice/sddm.service
          Unit: sddm.service
         Slice: system.slice
       Boot ID: a97015dd27cc4e1383ab27d683b81b69
    Machine ID: 1ef22b25243c4982987cb64cd5880329
      Hostname: coruscant
       Message: Process 14600 (sddm) of user 0 dumped core.
                
                Stack trace of thread 14600:
                #0  0x00007f831ea9618f _ZN9QIODevice5writeEPKcx (libQt5Core.so.5)
                #1  0x00005571e2829fcd _ZN4SDDM12SocketWriterD1Ev (sddm)
                #2  0x00005571e28460b0 _ZN4SDDM12SocketServer11loginFailedEP12QLocalSocket (sddm)
                #3  0x00007f831eb9b15a _ZN11QMetaObject8activateEP7QObjectiiPPv (libQt5Core.so.5)
                #4  0x00005571e284ad32 _ZN4SDDM7Display11loginFailedEP12QLocalSocket (sddm)
                #5  0x00005571e28343c8 _ZN4SDDM7Display20displayServerStartedEv (sddm)
                #6  0x00007f831eb9b15a _ZN11QMetaObject8activateEP7QObjectiiPPv (libQt5Core.so.5)
                #7  0x00005571e2839e42 _ZN4SDDM17XorgDisplayServer5startEv (sddm)
                #8  0x00005571e284236d _ZN4SDDM4Seat13createDisplayEi (sddm)
                #9  0x00005571e2842587 _ZN4SDDM4SeatC2ERK7QStringP7QObject (sddm)
                #10 0x00005571e2842f4f _ZN4SDDM11SeatManager10createSeatERK7QString (sddm)
                #11 0x00005571e2830905 _ZN4SDDM9DaemonAppC1ERiPPc (sddm)
                #12 0x00005571e28155c1 main (sddm)
                #13 0x00007f831dcad580 __libc_start_main (libc.so.6)
                #14 0x00005571e2815699 _start (sddm)
``

plfiorini:
LGTM

PR #533, opened 2015-11-06
Execute /etc/X11/Xsession if available

### LINKED ISSUE #449: SDDM never executes xrdb -merge $HOME/.Xresources on KDE/Plasma/Kubuntu
Running Kubuntu Willy with SDDM 0.11, I noticed that SDDM never actually uses my .Xresource file. After further investigation, it became clear that the part of the Xsession script responsible for merging it is never actually executed since execution stops on 

```
Loading X session script /etc/X11/Xsession.d/99x11-common_start
```

Explained further on the Xsession snippet below:

```
echo "OK so far"  # This will execute just fine


if [ -d "$xsessionddir" ]; then
    for i in `ls $xsessionddir`; do
        script="$xsessionddir/$i"
        echo "Loading X session script $script"
        if [ -r "$script"  -a -f "$script" ] && expr "$i" : '^[[:alnum:]_-]\+$' > /dev/null; then

            . "$script"  # Executing 99x11-common_start blocks further execution of the script 

        fi
    done
fi


echo "Not executed at all"  # Never get's here


if [ -d /etc/X11/Xresources ]; then
  for i in /etc/X11/Xresources/*; do
    [ -f $i ] && xrdb -merge $i
  done
elif [ -f /etc/X11/Xresources ]; then
  xrdb -merge /etc/X11/Xresources
fi
[ -f $HOME/.Xresources ] && xrdb -merge $HOME/.Xresources
```

plfiorini:
Ubuntu already provides a /etc/X11/Xsession script which executes
the /etc/X11/Xsession.d scripts.

Those scripts already do the Xresource merge and other actions.

The last script from /etc/X11/Xsession.d executes the session
command specified by $STARTUP, if that is empty it will try
to determine it by itself with the risk of executing something
else.

With this patch we let distros shipping /etc/X11/Xsession to
execute the designated session command and handle Xresource
merge and stuff like that by itself.

At least two other major distributions such as Fedora and
ArchLinux do not have /etc/X11/Xsession, in that case we
will keep the Xresource merge code path and run the session
command ourselves.

Fixes #449

### DIFF
--- a/src/helper/UserSession.cpp
+++ b/src/helper/UserSession.cpp
@@ -48,6 +48,7 @@ namespace SDDM {
         } else if (env.value(QStringLiteral("XDG_SESSION_TYPE")) == QStringLiteral("x11")) {
             qDebug() << "Starting:" << mainConfig.X11.SessionCommand.get()
                      << m_path;
+            env.insert(QStringLiteral("STARTUP"), m_path);
             QProcess::start(mainConfig.X11.SessionCommand.get(),
                             QStringList() << m_path);
         } else if (env.value(QStringLiteral("XDG_SESSION_TYPE")) == QStringLiteral("wayland")) {


jleclanche:
This is a half-measure. You know what I think about half-measures :smiling_imp: 

We're in a position where we can decide whether shipping /etc/X11/Xsession is right, or not. If we think it is, we tell maintainers to change sddm's SessionCommand. If not, we tell a different set of maintainers to start shipping their own Xsession. What do you think?

plfiorini:
We still need to set $STARTUP though

plfiorini:
Updated, now I just set STARTUP.
This is of course totally untested. Debian/Ubuntu users can give some feedback here?

plfiorini:
If this is not interesting for Debian and Ubuntu users I won't certainly care more than they do. This goes to 0.15 now.

Vdragon:
I'm a Ubuntu user, what feedback shall I provide?

PR #541, opened 2015-11-08
Wip/clientlog

plfiorini:


plfiorini:
i thought i have already sent this once and i remember you commenting it but i cannot find it anywhere.
so i'm just sending it another time.

plfiorini:
just force pushed the last revision, seems to work now

tuzonice:
Thanks. I’ll check that commit out, as soon as I can, though a Gentoo recompile is in the mean time (setting -fPIC globally on non-hardened is a very_bad_idea ;-) ).

jleclanche:
Please add a ChangeLog entry, specify the new default values and gtm

plfiorini:
@jleclanche commits have the ChangeLog tag just like Qt so we can automatically create change log files. We should always do so for important commits.

jleclanche:
@plfiorini It's easier to generate the changelog as we go along, that way progress is visible for those looking at the git repo.

plfiorini:
@jleclanche it can be done with another commit after, mixing code with other files seems bad and will screw up in case of a revert.

PR #553, opened 2015-11-11
Add UsersThresholdToDisableAvatars (and more)

antlarr:
This pull request:
- Adds a count property to UserModel with the number of users in the model.
- Renames ConfigReader::isDefault() to matchesDefault() since that's what it really do.
- Adds a real ConfigReader::isDefault() so we can check if a value was explictly set, or its value is just the default (so we can check if EnableAvatars was explicitly set to true or it's true just because that's the default and so, it can be overriden by usersThresholdToDisableAvatars)
- Adds a new config setting UsersThresholdToDisableAvatars. This is the number of users to use as threshold above which avatars are disabled automatically unless explicitly enabled with EnableAvatars.
- Adds a usersThresholdToDisableAvatars to UserModel

I've also changed plasma's breeze theme to use all this to show either a user list (with avatars) or username/password editlines depending on the number of users in the system, the value of usersThresholdToDisableAvatars.

This fixes the sddm problems in systems with a large number of users and NFS mounted homes as in https://bugzilla.opensuse.org/show_bug.cgi?id=953778

### DIFF
--- a/src/common/ConfigReader.cpp
+++ b/src/common/ConfigReader.cpp
@@ -186,17 +186,17 @@ namespace SDDM {
          * Initialization of the map of nondefault values to be saved
          */
         if (section) {
-            if (entry && !entry->isDefault())
+            if (entry && !entry->matchesDefault())
                 remainingEntries.insert(section, entry);
             else
                 for (const ConfigEntryBase *b : section->entries().values())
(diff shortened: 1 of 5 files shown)


plfiorini:
Merged

PR #563, opened 2015-11-26
Only allocate VTs for seat0. Preparing for multi-seat support (#78).

lbssousa:
Only systemd-logind special seat named "seat0" can handle VTs, so there's no sense in allocating VTs for seats other than seat0.

This is a preparation step for multi-seat support (see issue #78). The missing parts for full multi-seat support (specially those regarding DBus interaction with systemd-logind) will come later.


plfiorini:
I wonder how I can test it with other seats since I only have seat0 here.

PR #564, opened 2015-11-26
Only allocate VTs for seat0

lbssousa:
Currently only systemd-logind special seat named "seat0" can
handle VTs. Allocating a VT and setting XDG_VTNR for non-seat0
seats is currently pointless. Moreover, Xorg server 1.16 and
newer doesn't touch VTs if -seat option is passed with a
non-seat0 value, so there's no need to pass vt?? argument to
a non-seat0 Xorg server.

This is a preparation step for multi-seat support. The missing
parts for a complete support (specially those regarding DBus
interactions with systemd-logind) will be commited in a second
moment.

Issue: #78


plfiorini:
You could that without closing #563  :)
With ``git rebase -i HEAD~3`` and mark the last 2 commits with squash or fixup and then force push to your fork with ``git push -f``.

lbssousa:
@plfiorini Thanks for the tip! I didn't know git-push has a --force option :)

lbssousa:
Implementing full multi-seat support may require an external Qt library for wrapping systemd-logind DBus API. There's currently no one in official Qt upstream, but [libsystemd-qt](../../../ascarpino/libsystemd-qt) is an actively developed one.

plfiorini:
Yeah I know the library, but actually @davidedmundson wrote a very good wrapper too with proper DBus property support, though I don't know its current status.

http://blog.davidedmundson.co.uk/node/84

I know at some point David and Andrea discussed on a possible merge but don't know what was decided.

plfiorini:
@lbssousa I never touched multi-seat before, do you know how I can setup a seat1 to test this patch?
I guess it will just work with my setup because I only have seat0 which means I can only test whether it has regressions.

lbssousa:
@plfiorini you can set up a second seat for testing purposes with the following steps:
* Pick up an extra USB hub, keyboard or mouse.
* Write a udev rule for that device which will tag it as a seat master device. Name it `/etc/udev/rules.d/71-seat-testing.rules`. Example (replace the `[...]` placeholders with your device's properties):
```
SUBSYSTEM=="usb", ATTR{idVendor}=="[...]", ATTR{idProduct}=="[...]", TAG+="master-of-seat", ENV{ID_AUTOSEAT}="1", ENV{ID_FOR_SEAT}="testing"
```
* Apply the new rule by running `sudo udevadm trigger`. If everything goes fine, you should see a new entry `seat-testing` in the output of command `loginctl list-seats`, whenever your master device is connected.
* If you have any custom settings for your current video card in `/etc/X11/xorg.conf*`, edit them, putting a `MatchSeat "seat0"` entry in all `"Device"`, `"Screen"`, and `"ServerLayout"` sections. It will prevent these settings from interfering with other seats (on the other hand, it requires that `-seat seat0` option be passed to Xorg, which this PR already does).
* Write a new file `/etc/X11/xorg.conf.d/99-seat-testing.conf` with the needed settings for starting Xorg on your second seat. If you don't have a second graphics device, you can use `xf86-video-dummy` driver for this one. Example:
```
Section "Device"
    MatchSeat "seat-testing"
    Identifier "Dummy device"
    Driver "dummy"
    # Any other options you may want to set
    (...)
EndSection

# You can omit this section
Section "Screen"
    MatchSeat "seat-testing"
    Identifier "Dummy screen"
    Device "Dummy device"
    # Any other options for setting screen modes, depth, etc.
    (...)
EndSection
```
* Hack current SDDM code to explictly add a new seat for `seat-testing` while proper multi-seat implementation is not yet written.

lbssousa:
@davidedmundson 's QLogind seems a more ambitious project for me, since it ends up extending current `QDBusAbstractInterface` class. I'll try @ascarpino 's libQtSystemd first, and let all those `PropertyChanged` handling for a second moment.

ascarpino:
Hi there, I'm currently working at implementing asynchronous calls in libsystem-qt "stealing" David code. Any suggestion is appreciated!

plfiorini:
@lbssousa Thanks for multi seat setup instructions, I will try tonight

davidedmundson:
TBH my logind library is more complex than SDDM needs.

It goes to a lot of effort to be all async, and handle all state changes properly, but SDDM doesn't actually need that. 
SDDM will be just idling till we get the response, so we may as well just use the Qt blocking property calls. That simplifies things greatly.

I'd just generate the xml files to cpp inside SDDM rather than using any external lib.

lbssousa:
I'm testing SDDM with two seats here. It can spawn a `Xorg` server for each seat correctly, but `sddm-greeter` crashes on start for the second seat.

Moreover, the second seat's `Xorg` server is not writing in its own log file, but it's overwriting `seat0`'s one. See this bug: https://bugs.freedesktop.org/show_bug.cgi?id=93212

I believe a possible workaround here is making SDDM pass an explicit display number along with `-displayfd` to `Xorg`.

lbssousa:
I've just updated this VR against latest code in `develop` branch.

src/daemon/Display.cpp
```diff
@@ -309,32 +309,26 @@ namespace SDDM {
...
-            if (session.xdgSessionType() == QLatin1String("wayland")) {
+        if (seat()->name() == QLatin1String("seat0")) {
+            int vt = terminalId();
+            if (session.xdgSessionType() == QLatin1String("wayland"))
                 vt = VirtualTerminal::setUpNewVt();
-                VirtualTerminal::jumpToVt(vt);
-            }
-        }
+            m_lastSession.setVt(vt);
+            env.insert(QStringLiteral("XDG_VTNR"), QString::number(vt));
```
davidedmundson: How does this work? 

env isn't created till line 320

lbssousa: Just fixed it!

lbssousa:
@davidedmundson I've fixed that error. Could you please review it again?

lbssousa:
Closing this PR in favour of https://github.com/sddm/sddm/pull/1182,  since this branch has become too outdated.

PR #573, opened 2015-12-13
Set the focus on the primary screen

### LINKED ISSUE #501: sddm 0.12.0 does not focus the password field and the keyboard does nothing
When booted up SDDM comes up with the last used username selected and the password field apparently "focused" but in reality it is not, as when I try to type it does nothing. Also the keyboard keys (e.g. TAB) don't do anything at all at that moment until I use the mouse to click on the password field which then I can type in the password and also use TAB to move around the fields.

benjarobin:
Fix #501

Call requestActivate() on the view associated with the primary screen

### DIFF
--- a/src/greeter/GreeterApp.cpp
+++ b/src/greeter/GreeterApp.cpp
@@ -37,6 +37,7 @@
 #include <QQmlEngine>
 #include <QDebug>
 #include <QTranslator>
+#include <QScreen>
 
 #include <iostream>
 
@@ -133,6 +134,9 @@ namespace SDDM {
         Q_FOREACH (QScreen *screen, screens)
             addViewForScreen(screen);
 
+        // set focus on the primary screen
+        setFocusOnPrimaryScreen();
+
         // handle screens
         connect(this, &GreeterApp::screenAdded, this, &GreeterApp::addViewForScreen);
     }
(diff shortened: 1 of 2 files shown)


benjarobin:
Why ? Because the code handle all case, there is a function (better than duplicate code), the for loop cost nothing compare to handling graphical things. And this is a for loop on 2 or 3 elements...
In your pull request https://github.com/sddm/sddm/pull/602 you do not handle the case of no primary screen (The Qt doc is not clear on this case)

plfiorini:
QGuiApplication::primaryScreen is null when there are no screens connected hence there is no QtQuick view.

plfiorini:
Also: when a screen is removed the window is assigned to the primary screen and we know when the primary screen changes since Qt 5.6 so the for loop in the remove screen slot should be used only for Qt < 5.6

plfiorini:
Why you try to find the "mainView"?
There is one view assigned to each screen, if the primary screen changes most likely is because the previous primary was removed. In that case you still have one view assigned to each of the screens left, one of them becomes primary which means all you have to do is request activate if it's primary

benjarobin:
I did not say that my code was perfect, not better that yours. My code is just more generic, you can call this function from anywhere. Anyway I only want this bug fixed. It fine to close this pull request...

PR #574, opened 2015-12-13
Theme Maui: Prevent losing focus

benjarobin:
And also word wrap the user text message
=> the message is no longer centered, but at least it now fit on almost every resolution


plfiorini:
merged

Issue #527, opened 2015-11-05
Delete state.conf Last section saved User when RememberLastUser set to false


openizy:
I use openSuSE Leap 42.1. I add RememberLastUser=false to /etc/sddm.conf. But when next restart my computer. Last login user is still fill to use name textbox. I think sddm can modify theme config, but I can not good at Main.qml file. And I think better is clear /var/lib/sddm/state.conf file, delete Last section saved User data when sddm success login and when RememberLastUser set to false.

I delete /var/lib/sddm/state.conf Last section saved User data by vi. And sddm worked OK!!!
Thanks.

PR #580, opened 2015-12-30
Honor RememberLastUser and RememberLastSession settings. [fix issue #527]

mitchnull:
Revert to default if RememberLastUser or RememberLastSession is
set to false.

Fix issue #527.


plfiorini:
lgtm

PR #581, opened 2015-12-30
Simplify reading lists from configuration

mitchnull:
Remove unnecessary loop when parsing lists. The issue fixed in
56f64173991204a1eea4bf6789d68555b94e3eae was caused by a dangling-reference
error.  Store readLine() so that it doesn't go out-of-scope while we
process splitRef().


jleclanche:
@plfiorini This is your regression :) the code lgtm

plfiorini:
It was just an unnecessary while :)
Anyway we already have this fix so I'm gonna close the PR.

PR #583, opened 2016-01-02
Run .xsessionrc

Myrl:


jleclanche:
@plfiorini Can you review? I have no idea if this is correct.

plfiorini:
.xsessionrc might not be executable, in that case this line will fail.
Besides it should probably be sourced because according to [1] it is used to set session wide env vars

[1] https://wiki.ubuntu.com/X/Config/SessionStartup

plfiorini:
So in its current form :-1:

PR #589, opened 2016-01-24
Make Spam Safe

GinoM:
* This Should help with annoying spam
* I changed my email address


jleclanche:
I can change your own email but please don't change everybody else's.

PR #593, opened 2016-01-28
Add instructions on how to change DPI

### LINKED ISSUE #402: Add an option to set DPI value
Hello,

Currently, SDDM does not offer any option to set a DPI value. It seems that the value used may be 96 DPI, perhaps hard-coded somewhere.

On "retina"-like  screen (Apple products, but also some Lenovo, Dells, etc.), the size of everything is way too small, but would appear just right with a DPI value of 192.

Would it be possible to have the possibility to set the DPI value in sddm.conf?

plfiorini:
Closes #402

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,3 +78,19 @@ ACLs if your filesystem does support it.
 setfacl -m u:sddm:x /home/username
 setfacl -m u:sddm:r /home/username/.face.icon
 ```
+
+### Custom DPI
+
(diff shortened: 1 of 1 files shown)


jleclanche:
What do you think about adding a config variable for DPI instead? We'll need it if sddm is to run the same way on x11 and wayland.

plfiorini:
No we don't, Wayland doesn't need it :P

PR #596, opened 2016-02-03
Allow to start several sessions

chrisbra:
Set the VTCount configuration option to specify how many sessions
should be started. Default value is 1.

There have been discussions on IRC about it before and the result is this small change.


jleclanche:
-1

jleclanche:
This isn't the right way to address the use case that was talked about on IRC at all. We want to be able to open new sessions at will, not mess around in conf files.

plfiorini:
-1 also from me

PR #602, opened 2016-02-29
Activate window for the primary screen

### LINKED ISSUE #501: sddm 0.12.0 does not focus the password field and the keyboard does nothing

plfiorini:
This gives focus to text fields.
While we're here, remove the unused m_views.

Closes #501

[ChangeLog][Greeter] Fix text field focus (issue #501)


plfiorini:
I'm going to retest this shortly

PR #603, opened 2016-02-29
Embed "maui" into the resources

plfiorini:
Fallback to "maui" every time the them has an error so we'll always
let the user log in.

The "maui" theme is now embedded into the resources to be sure
it exists.

[ChangeLog][Greeter] Embed the "maui" theme and fallback to it
if there are issues loading the QtQuick view.


jleclanche:
Landed in f7d3f67.

PR #616, opened 2016-04-03
WIP: Wayland support for the greeter

plfiorini:


plfiorini:
Status as of today: almost working

The socket presence check sucks and needs to be better written, maybe a QEventLoop and QFileSystemWatcher. Doing asynchronous code in a synchronous call is weird.

Managed to get the compositor up and running once but got an error when taking vt control. There are problems with weston too, so it's not the compositor's fault rather sddm.

plfiorini:
Update: compositor and greeter start but libinput is not working, it's weird because it works with Wayland user sessions (even tested with Hawaii that uses the same compositor infrastructure). There is definitely something missing between sddm and logind.

Greeter needs a keyboard implementation for Wayland using xkbcommon.

plfiorini:
Needs #673 first otherwise we won't be able to run X11 sessions from a Wayland greeter

DeX77:
Any news on this? Seems like #673 is done so far?

SparK-Cruz:
Still no wayland greeter?

ericschdt:
It's on the TODO list of KDE: 

> sddm without X

 https://phabricator.kde.org/T11640 :)

bam80:
Do we really need even Wayland here? I mean - can't the DM draw directly?

soredake:
Any progress?

plfiorini:
Contiuing on #1367

PR #617, opened 2016-04-03
Input Method settings

plfiorini:
This series only add a setting to configure the input method for the greeter, useful nevertheless.
Still need to decide where and how put the Qt virtual keyboard integration.


jleclanche:
Squash? :)

plfiorini:
guess so

also this can go in 0.14 as is

PR #638, opened 2016-05-25
Fix potential tiny memory leak

davidedmundson:


plfiorini:
+1 - makes sense but haven't actually tried

PR #648, opened 2016-06-18
Source $USERXSESSION if file exists

### LINKED ISSUE #647: Source ~/.xsession
Looks like we don't source it:
https://github.com/sddm/sddm/blob/master/data/scripts/Xsession#L65

jokogr:
If there is ~/.xsession, then it is being sourced instead of checking
the system sessions.

Fixes #647


jokogr:
There are also some other unused variables, such as `OPTIONFILE`, `USERXSESSIONRC` and `ALTUSERXSESSION` which are not used anymore, let me know if it is ok to remove those.

PR #684, opened 2016-09-01
Fix display of user avatars.

bastianbeischer:
QFile::exists("...") does not understand file:// URLs, at least in Qt
5.7.0 and Qt 4.8.7.


plfiorini:
Oops

PR #689, opened 2016-09-04
Add Hindi translations

spremi:
Translations are based on usage (noun/ verb) in the default (English) greeter.
However, wasn't able to test in either production/ test mode on my Fedora 23 desktop.

### DIFF
--- a/data/translations/CMakeLists.txt
+++ b/data/translations/CMakeLists.txt
@@ -7,6 +7,7 @@ set(TRANSLATION_FILES
     et.ts
     fi.ts
     fr.ts
+    hi_IN.ts
     hu.ts
     it.ts
     ja.ts
--- a/data/translations/hi_IN.ts
+++ b/data/translations/hi_IN.ts
@@ -1,0 +1,59 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.1" language="hi_IN">
+<context>
+    <name>TextConstants</name>
+    <message>
+        <source>Welcome to %1</source>
+        <translation>%1 में आपका स्वागत है।</translation>
+    </message>
+    <message>
+        <source>Warning, Caps Lock is ON!</source>
+        <translation>सावधान! कैप्स लॉक सक्रिय है।</translation>
+    </message>
+    <message>
+        <source>Layout</source>
+        <translation>विन्यास</translation>
+    </message>
+    <message>
+        <source>Login</source>
+        <translation>सत्यापन कीजिए</translation>
+    </message>
+    <message>
+        <source>Login failed</source>
+        <translation>सत्यापन विफल रहा।</translation>
+    </message>
+    <message>
+        <source>Login succeeded</source>
+        <translation>सत्यापन सफल हुआ।</translation>
+    </message>
+    <message>
+        <source>Password</source>
+        <translation>कूटशब्द</translation>
+    </message>
+    <message>
+        <source>Enter your username and password</source>
+        <translation>अपनी पहचान और संबंधित कूटशब्द डालिए।</translation>
+    </message>
+    <message>
+        <source>Reboot</source>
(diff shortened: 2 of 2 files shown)


spremi:
Here are steps I followed for testing:

In production mode:
1) Copied `hi_IN.qm` to `/usr/share/sddm/translations/`
2) Changed `/etc/locale.conf` to
```
LANG="hi_IN.UTF-8"
LC_MESSAGES="hi_IN.UTF-8"
```
3) Added to `/etc/sddm.conf`
```
[XDisplay]
DisplayCommand=/etc/sddm/Xsetup
```
4) Added to `/etc/sddm/Xsetup`
```
export LANG="hi_IN.UTF-8"
export LC_MESSAGES="hi_IN.UTF-8"
```
5) Reboot
Observation: Date is localized, but messages are still in English.

In test mode:
```
$ cd build/src/greeter
$ LANG=hi_IN.UTF-8 && LC_MESSAGES=hi_IN.UTF-8 && ./sddm-greeter --test-mode --theme ./theme
```
Observation: Date is localized, but messages are still in English.

Any suggestions? Esp. if I missed something.

plfiorini:
@spremi Do you have your language enabled in /etc/locale.gen?
Uncomment hi_IN UTF-8 in that file and run locale-gen as root, restart the sddm service with systemctl restart sddm and see if the strings are now displayer correctly.

spremi:
@plfiorini Fedora doesn't have /etc/locale.gen.
I have enabled `hi_IN UTF-8`, but for some reasons `LANG` reverts back to `en_IN.UTF8` after reboot.
Still looking for reasons for it... Will be able to spend time tomorrow.

Here are few things that I already tried:
```
$ locale
LANG=en_IN.UTF-8
LC_CTYPE=hi_IN.UTF-8
LC_NUMERIC=hi_IN.UTF-8
LC_TIME=hi_IN.UTF-8
LC_COLLATE=hi_IN.UTF-8
LC_MONETARY=hi_IN.UTF-8
LC_MESSAGES=hi_IN.UTF-8
LC_PAPER=hi_IN.UTF-8
LC_NAME=hi_IN.UTF-8
LC_ADDRESS=hi_IN.UTF-8
LC_TELEPHONE=hi_IN.UTF-8
LC_MEASUREMENT=hi_IN.UTF-8
LC_IDENTIFICATION=hi_IN.UTF-8
LC_ALL=
$
```
With these settings, I see Hindi used almost in all applications (and the date widget in greeter).
Font is definitely different in the greeter than my usual with `LANG=en_US.UTF-8`.

Also tried setting explicit font in the breeze theme as well:
```
[General]
background=components/artwork/background.png
displayFont="Lohit Devanagari"
```

spremi:
Rebased with master...

spremi:
Turns out that Fedora builds use different directories:
```
DATA_INSTALL_DIR
COMPONENTS_TRANSLATION_DIR
```
Therefore, the translations were not available when I had copied new binaries.
Re-compiled after changing `build/src/common/Constants.h`
```
#define DATA_INSTALL_DIR            "/usr/share/sddm"
#define COMPONENTS_TRANSLATION_DIR  "/usr/share/sddm/translations"
```
Now, I can see the translations in work, but for the login button, password widget - which is still in English.
Will try with proper CMake override of these variables...

spremi:
After changing theme to '02-fedora', I could verify more strings appearing correctly.
The text on 'Reboot' and 'Shutdown' Buttons is correct as well.
However, with 'breeze' the text on buttons (Login) and edit box (Password) are not translated.
Could be a problem with the themes themselves...

spremi:
Another few hours of debug indicate that 'breeze' uses many strings from Plasma components that don't use the strings in SDDM. Hence, they don't show the translations.
I have _almost_ created a simple theme for the testing - where, i have used most strings. No surprises so far... Hope it should verify all translations.

PR #705, opened 2016-09-22
Make the default cursor themed

Vogtinator:
The X root window uses an "X" as the default cursor. To use a themed one,
it has to be set explicitly, e.g. by QML MouseAreas. For themes which do
not have a MouseArea that spans the entire screen, the "X" cursor would
be visible. By setting a cursor in the root QML object, the themed cursor
will be used by default. (https://bugs.kde.org/show_bug.cgi?id=337083)

[ChangeLog][Greeter] Fix default cursor appearance


lbeltrame:
Tested here, confirmed working.

PR #713, opened 2016-10-02
add fish shell specific commands to Xsession and wayland-session

florianjacob:
to get the fish-specific login environment instead of the catchall /etc/profile environment.

Prior to this, users with fish shell as default logging in via sddm got an environment as specified by `/etc/profile` but without user-specific configs that only execute on the login shell. In contrast, when logging in via virtual terminals or SSH, they got an environment as specified by `/usr/share/fish/config.fish` and with user-specific configs that execute on login shell.

With this path, an sddm user gets a consistent environment regardless of login method.

### DIFF
--- a/data/scripts/Xsession
+++ b/data/scripts/Xsession
@@ -38,6 +38,12 @@ case $SHELL in
     . $xsess_tmp
     rm -f $xsess_tmp
     ;;
+  */fish)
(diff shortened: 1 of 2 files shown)


florianjacob:
Thanks! I agree, but I wasn't able to come up with something less messy han the csh solution that actually worked, despite my best efforts. :disappointed_relieved:

PR #714, opened 2016-10-03
add kwallet5 support

naund:
Add kwallet-pam support for the plasma5 desktop


davidedmundson:
Can you do sddm-autologin.pam too please

PR #724, opened 2016-10-18
Multiseat support

davidedmundson:
Drop outdated DisplayManager iface - implement logind/consolekit2 API

Fixed reimplementation of the multi seat code I had but without the qlogind dependency.

Tested with a docking station with two seats. 

Debug shows everythign working properly with regards to seat detection, but there's sometimes a break somewhere in helper code if we try and do start two xorgs at once, but that's sort of unrelated to reviewing this.

---
I've made everything asynchronous - but on reflection, that's rather pointless, if logind was blocked then we'd just block in PAM, so this doesn't acheive much and it's not going to deadlock.. I'm tempted to make it blocking and then I can use the autogenerated interface which makes the code cleaner. Thoughts?


davidedmundson:
> but there's sometimes a break somewhere in helper code

Found it. If an Xorg fails to start SDDM crashes, my second session failed to start due to a bad driver setup. I'll fix that crash, but it shows it's unrelated to here.

PR #726, opened 2016-10-20
Logind Multi Seat Support

davidedmundson:
Edit.

It turns out we can't drop the legacy support. logind provides 95% of features, but no implementation of switchToGreeter().

We can't drop it for the forseeable future. However, it does seems it's possible to both track automatic seats through logind, and expose the old FDO interface.

### DIFF
--- a/src/auth/Auth.cpp
+++ b/src/auth/Auth.cpp
@@ -52,6 +52,7 @@ namespace SDDM {
         Q_OBJECT
     public:
         Private(Auth *parent);
+        ~Private();
         void setSocket(QLocalSocket *socket);
     public slots:
         void dataPending();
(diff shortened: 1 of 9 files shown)


AdamWill:
I'm a bit confused - why was #724 merged instead of this? Doesn't it remove user switching support? Am I missing something?

davidedmundson:
#724 isn't merged..

No idea why github says it is. Github is weird.

timknab:
Has anyone tried @davidedmundson branch of SDDM to include multiseat support recently? I've tested both the [`logind2`](https://github.com/davidedmundson/sddm/tree/logind2) and [`logind2_alt1`](https://github.com/davidedmundson/sddm/tree/logind2_alt1) branches both on their own and merged into `SDDM/develop`.

In all cases the second seat is detected and the SDDM greeter appears on the second display. I can login on seat1 to a variety of desktop environments (Gnome, Gnome on Wayland, Plasma, Plasma on Wayland etc). However, regardless of which seat I choose on Seat2 SDDM freeze and crashes.

PR #735, opened 2016-11-02
Switch to another managed VT when the current one quits

### LINKED ISSUE #653: Fast user switching drops to text console
## System information

- Running in Virtual Box
- KDE Neon 5.7.1
- sddm 0.13.0

## Steps to reproduce

I tried to play trough a simple fast user switching scenario with two users. 

1. User A logges in
2. User A locks screen
3. Create a new session
4. User B loggs in
5. User B locks screen
6. Switch Session to User A
7. User A unlocks screen
8. User A logs out

## Expected Behaviour

The session is switched to User B (locked screen)

## Observed Behaviour

After User A is logged out, the system drops to a text console (tty1).

## Further information 

In KDE Neon sddm starts on tty7. When I configure sddm to start on tty1 (MinumumVT=1 in /etc/sddm.conf) then it drops to text console on step 6.

MartinBriza:
Fixes #653 

Not a high quality patch. I guess it should check if the current VT is actually active and probably use stuff from the VirtualTerminal namespace.

There was a change that makes stopping X server on a VT not switch to a previously active VT (if you have more than one of them). This is pretty annoying for users who don't know how to switch VTs manually because the current behavior just leaves you looking at the text terminal under where your X session was a while ago.

### DIFF
--- a/src/common/Configuration.h
+++ b/src/common/Configuration.h
@@ -42,6 +42,7 @@ namespace SDDM {
         Entry(Numlock,             NumState,    NUM_NONE,                                       _S("Initial NumLock state. Can be on, off or none.\n"
                                                                                                    "If property is set to none, numlock won't be changed\n"
                                                                                                    "NOTE: Currently ignored if autologin is enabled."));
+        Entry(EnableHiDPI,         bool,        true,                                           _S("Enable Qt's automatic high-DPI scaling"));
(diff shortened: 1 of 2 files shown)


oliverhenshaw:
I think the xserver does switch to the VT that was active before it started. I think it's just that this doesn't work out if the old active VT has also gone away before you logged out of this X server, or you're stopping the original X server. So checking whether the current VT is active is probably good.

Issue #217, opened 2014-07-01
Offer a way to override default config via /usr/share/sddm/config.d


shadeslayer:
From a downstream perspective, when shipping multiple themes, it would be awesome to have themes install a config file to /usr/share/sddm/config.d to override the default theme from /etc/sddm.conf

Alternatively, maybe implementing a fall back mechanism would make sense as well. For eg. Try to load theme from /etc/sddm.conf , if theme doesn't exist check /usr/share/sddm/conf.d

jriddell:
This is needed to make nvidia-prime support elegant
https://github.com/sddm/sddm/issues/393

plfiorini:
:+1: except for minor details regarding the location

I would follow the systemd convention here so SDDM should:

* Load /etc/sddm.conf (for compatibility)
* Load /etc/sddm/sddm.conf (new main config file name)
* Load /usr/lib/sddm/sddm.conf.d/*.conf sorting in alphabetica order
* Load /etc/sddm.conf.d/*.conf sorting in alphabetical order

/etc/sddm.conf.d/*.conf is where the system administrator puts his stuff
/usr/lib/sddm/sddm.conf.d/*.conf is where the distro puts its stuff

majewsky:
I think the last one in your list should be `/etc/sddm/sddm.conf.d/*.conf`. The convention is to append `.d/*.conf` to the filename of the singular config file, in this case `/etc/sddm/sddm.conf`.

Compare, for example, `/etc/systemd/journald.conf` and `/etc/systemd/journald.conf.d/*.conf`.

plfiorini:
Fixed thanks, it was a typo

jleclanche:
@plfiorini We could do it like supervisor does it:

```ini
[include]
files = /etc/supervisor.d/*.ini
```

jleclanche:
@plfiorini btw this would take backwards compatibility out of our hands and into the distros', then :)

plfiorini:
@jleclanche i'd like to have more predictable file locations and copy the systemd pattern because it is likely to be more familiar to our user base

jleclanche:
@plfiorini with an include directive, we could do whatever we want.

majewsky:
Indeed, but what's the added value compared to the cost of inconsistency between distributions?

jleclanche:
Frankly sddm's not here to solve how inconsistent distros are between each other.

majewsky:
[...]

PR #737, opened 2016-11-03
    Load configuration information from config directories

davidedmundson:
Load configuration information from config directories
    
    Our configuration generally consists of user set config options and
    distro defined options.
    
    For example, a distro might want to specify a custom theme, but the
    user's autostart name can't come from the distro.
    
    Using the same config file leads to problems when upgrading.
    
    This loads configuration from all files in a directory
    (/etc/sddm/conf.d) as well as the main config file /etc/sddm.conf
    The latter has priority and all writes occur in that file.


shadeslayer:
Ping?

plfiorini:
@davidedmundson It should follow systemd convention of loading vendor config from /usr/lib/sddm/sddm.conf.d and site config from /etc/sddm/conf.d

The reason is that site config can be removed with a factory reset and still have a working system with vendor defaults.

See #217

PR #739, opened 2016-11-10
autologin: try last successful session if no autologin session is set

schnitzeltony:
* By this the user can decide which session to select for autologin without
  touching priviledged configuration file.
* If session is set in configuration, the behaviour is same as before

Signed-off-by: Andreas Müller <schnitzeltony@googlemail.com>


plfiorini:
Merged

PR #746, opened 2016-11-24
Support elogind as an alterantive session tracker for systemd-logind.

Yamakuzure:
For users who do not want to run the full systemd suite, but do not
want to patch sddm to use ck-launch-session either, elogind can be
used as an alternative session tracker via its PAM module.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,7 +109,7 @@ if ("${ECM_VERSION}" VERSION_LESS "1.7.0")
 endif()
 
 # systemd
-if(NOT NO_SYSTEMD)
+if(NOT NO_SYSTEMD AND NOT USE_ELOGIND)
     pkg_check_modules(SYSTEMD "systemd")
 endif()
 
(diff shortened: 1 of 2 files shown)


dreaming-augustin:
sddm reliance on systemd is a disturbing trend.
Let me know if you need me to test your patch. Kindly provide steps on how to do so.

Yamakuzure:
@dreaming-augustin : sddm has no disturbing reliance on systemd, it just needs a session tracker for its full functionality. It currently works with systemd-logind and ConsoleKit2. Although the latter can only be used if the exec line in data/scripts/Xsession is changed from
`    exec $@`
to
`    exec ck-launch-session $@`

However, this patch is just about supporting elogind as an alternative to systemd-logind, so users who do not wish to use systemd as  their init, but are uncomfortable with ConsoleKit2 as well, can have the benefits offered by a session tracker nevertheless.

Currently sddm-0.14.0 works pretty well on my system, run by OpenRC, using elogind-226 (dev_v226 branch) as a session tracker.

a17r:
I've been using this patch for many weeks without issues. Plasma Wayland sessions, suspend and restore, elogind/openrc does it all for me.

Yamakuzure:
It works pretty well to the day, could someone review it, please?

CMakeLists.txt
```diff
@@ -143,12 +143,34 @@ if(SYSTEMD_FOUND)
...
+
+if(ELOGIND_FOUND)
+    add_definitions(-DHAVE_ELOGIND)
+    set(CMAKE_AUTOMOC_MOC_OPTIONS -DHAVE_ELOGIND)
+
+    set(MINIMUM_VT 7)
+    set(HALT_COMMAND "/usr/bin/loginctl poweroff")
+    set(REBOOT_COMMAND "/usr/bin/loginctl reboot")
+else()
+    set(ELOGIND_FOUND 0)
```
davidedmundson: why?
It's already not set, not set is better than 0.

Yamakuzure: Oh? I am no CMake guru, I thought it must be explicitly set...

I'll change and test that. Give me a few minutes, please...

Yamakuzure:
set(ELOGIND_FOUND 0) has been removed.

a17r:
elogind support in all Plasma/Frameworks dependencies has been progressing well, with upstreams such as accountsservice, networkmanager, udisks having accepted the change already. Consequently it will soon be our preferred session tracker for non-systemd systems, since it allows to support Plasma Wayland...

Issue #480, opened 2015-09-03
SDDM themes: Installed Main.qml files are not configured by CMake


bastianbeischer:
When using SDDM d42700a31375d5c82359df63fafe03a5eaa824f6 after running:

```bash
cmake "$srcdir/$_pkgbase" \
                -DCMAKE_INSTALL_PREFIX=/usr \
                -DCMAKE_INSTALL_LIBEXECDIR=/usr/lib/sddm \
                -DCMAKE_BUILD_TYPE=Release \
                -DBUILD_MAN_PAGES=ON
make
make DESTDIR="$pkgdir" install
```

(standard Arch Linux script to create the sddm package, based on sddm master from github)

I find that all the 'Main.qml' files belonging to themes are not 'configured' by CMake, i.e. the `${COMPONENTS_VERSION}` lines are not replaced with the actual value.

This causes a whitescreen when starting SDDM.

Things start to work when manually editing the the Main.qml file and replacing `${COMPONENTS_VERSION}` with `2.0`

plfiorini:
this apparently doesn't happen with the official package [*] I wonder what's different with sddm-git

[*] = https://projects.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/sddm

bastianbeischer:
Sorry I had forgotton about my report. I'm not sure the issue was real.

I think the problem went away when I cleaned my build directory and started from scratch. I had been doing incremental builds when the problem happened. Does that sound an explanation? Anyway, I can't reproduce this anymore.

plfiorini:
That explains :)

Thanks

PR #747, opened 2016-11-28
Don’t install themes Main.qml/metadata.desktop twice, in some conditions they don’t get configured

Coucouf:
I’ve run into what I feel is #480 while working on themes for Debian.
The cmake debug logs were mentioning twice "…/Main.qml" and for some themes the second one was just considered "Up-to-date:" and would not replace the unconfigured file.
This patch excludes Main.qml from the install(…) directive to just keep the one provided by configure_file(…).


davidedmundson:
If this is the case for Main.qml surely the same is also true for Metadata.desktop  which has the same code path?

Coucouf:
Yes, I had missed that.
Branch is now updated to also fix metadata.desktop double installation.

Issue #781, opened 2017-02-26
Improve Customizability


preisi:
Hello,
as part of a university linux project we (a team of 2 students) want to contribute to SDDM.

One of our goals is to improve the customizability of SDDM by adding a few more configuration options for various things:
- custom keybindings for navigation, etc.
- custom timeouts (how long shall the greeter wait until all input fields get cleared again)
- external functionality + appearance (add the option to easily add buttons to e.g. start scripts like a registration-mask, etc.)

What do you think about this planned functionality?

Cheers,
Thomas and Milan

P.S.: Of course we will develop this feature in the open on GitHub and give updates here.

adocampo:
Is this project still alive? Customization of SDDM is important for corporate environments and I have some ideas.

PR #794, opened 2017-03-11
Reset the login-form after idle-timeout

milanstephan:
Useful for multi-user-environments. Resets the form after a configurable idle-timeout to always present users clean forms.
#781

### DIFF
--- a/src/greeter/GreeterApp.cpp
+++ b/src/greeter/GreeterApp.cpp
@@ -39,6 +39,7 @@
 #include <QDebug>
 #include <QTimer>
 #include <QTranslator>
+#include <QEvent>
 
 #include <iostream>
 
@@ -58,6 +59,16 @@ namespace SDDM {
     }
 
     GreeterApp *GreeterApp::self = nullptr;
+
+    // catch key-press or mouse-click to restart the timer that resets the form
+    bool GreeterApp::notify(QObject *receiver, QEvent *event) {
+        if (event->type() == QEvent::KeyPress || event->type() == QEvent::MouseButtonPress) {
+            m_formResetTimer.stop();
+            m_formResetTimer.start();
+        }
+
+        return QGuiApplication::notify(receiver, event);
+    }
 
     GreeterApp::GreeterApp(int &argc, char **argv) : QGuiApplication(argc, argv) {
         // point instance to this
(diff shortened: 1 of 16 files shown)


src/greeter/GreeterApp.cpp
```diff
@@ -59,6 +60,16 @@ namespace SDDM {
 
     GreeterApp *GreeterApp::self = nullptr;
 
+    // catch key-press or mouse-click to restart the timer that resets the form
+    bool GreeterApp::notify(QObject *receiver, QEvent *event) {
+        if (event->type() == QEvent::KeyPress || event->type() == QEvent::MouseButtonPress) {
+            m_formResetTimer.stop();
```
davidedmundson: just m_formResetTimer.start()
see docs of QTimer

but more importantly, there's a better way to do event filtering.

Currently you're watching *every* single QObject in your entire application. When you move the mouse, you're restarting the timer a bunch of times.

What you want is to run installEventFilter(this) when we create the QQuickViews.
Then change this method to override eventFilter instead of notify

src/greeter/SessionModel.cpp
```diff
@@ -148,11 +148,21 @@ namespace SDDM {
...
         }
+
+        QString defaultSession = SDDM::mainConfig.DefaultSession.get();
+
         // find out index of the last session
         for (int i = 0; i < d->sessions.size(); ++i) {
-            if (d->sessions.at(i)->fileName() == stateConfig.Last.Session.get()) {
-                d->lastIndex = i;
-                break;
+            if (defaultSession.length() > 0) {
```
davidedmundson: if (!defaultSession.isEmpty()) {
}

src/greeter/GreeterApp.cpp
```diff
@@ -134,6 +145,17 @@ namespace SDDM {
         connect(this, &GreeterApp::primaryScreenChanged, this, [this](QScreen *) {
             activatePrimary();
         });
+
+        // start timer to reset the form if user is idle
+        int formResetTime = SDDM::mainConfig.FormResetTime.get();
+        if (formResetTime > 0) {
+            if (m_metadata->formResetSupport()) {
+                connect(&m_formResetTimer, &QTimer::timeout, m_proxy, &GreeterProxy::formResetAction);
```
davidedmundson: you can connect a signal to a signal directly

i.e

 +                connect(&m_formResetTimer, &QTimer::timeout, m_proxy, &GreeterProxy::formReset);

plfiorini:
None of @davidedmundson has been addressed in 11 months so I'm closing this PR, please send it again if you address those comments.

Issue #783, opened 2017-02-26
Administrative Features


preisi:
Hello,
as part of a university linux project we (a team of 2 students) want to contribute to SDDM.

One of our goals is to add the possibility for administrators to diplay global hints/messages.
(e.g. maintenance times announcements)

What do you think about this planned functionality?

Cheers,
Thomas and Milan

P.S.: Of course we will develop this feature in the open on GitHub and give updates here.

Kendos-Kenlen:
I use SDDM in a personal context, so without any company network or anything, but I think it could be a great feature.
A simple feature could be a notice if someone tried to connect to the computer but failed (and keep the notice after a reboot for example). A notice about available updates would be also nice, or displaying messages of any type.

preisi:
Hey, thanks for the feedback.

The basic idea is to add a configuration-entry for a path to an html-file containing the (already formatted) information which would then be displayed in a previously in the theme defined area on the greeter.

So technically your request of having notifications about failed connection/authentification attemps or even available updates should be easily possible by running a script to generate such an html file.

PR #797, opened 2017-03-13
Implement AdminInfo-Interface for displaying administrative informati…

preisi:
…on on the greeter

Adds a config entry for an "AdminInfoPath". This file can then be
displayed by the theme (both plaintext and basic HTML). Any file
modifications are tracked and the theme is being notified.

In order for themes to support this, their metadata needs to contain
"AdminInfoSupport=true".

Signed-off-by: Thomas Preisner <thomas.preisner@fau.de>
Signed-off-by: Milan Stepahn <milan.stephan@fau.de>


preisi:
#783

src/greeter/GreeterApp.cpp
```diff
@@ -134,6 +155,39 @@ namespace SDDM {
         connect(this, &GreeterApp::primaryScreenChanged, this, [this](QScreen *) {
             activatePrimary();
         });
+
+        // create filewatcher for adminInfo-file
+        QString adminInfoPath = SDDM::mainConfig.AdminInfoPath.get();
+
+        // check if config has been set
+        if (adminInfoPath.size() > 0) {
```
Zopieux: [`!adminInfoPath.isEmpty()`](http://doc.qt.io/qt-5/qstring.html#isEmpty)

preisi:
Seems like the Travis CI is broken. (Take a closer look at the log!)
Apparently Arch is/was broken at the time I updated the pull request (see https://www.archlinux.org/news/ca-certificates-utils-20170307-1-upgrade-requires-manual-intervention/).
Is it possible to rerun the test?

src/greeter/GreeterApp.cpp
```diff
@@ -134,6 +155,39 @@ namespace SDDM {
...
+        QString adminInfoPath = SDDM::mainConfig.AdminInfoPath.get();
+
+        // check if config has been set
+        if (!adminInfoPath.isEmpty()) {
+            // check for theme-compatibility
+            if (m_metadata->adminInfoSupport()) {
+                m_fileAgent = new QFileSystemWatcher(this);
+
+                // connect signals
+                connect(m_fileAgent, &QFileSystemWatcher::fileChanged, this, &GreeterApp::fileChanged);
```
preisi: No, you are right. It does not work in this specific scenario.
But keep in mind that your solution (if I understood it correctly) won't solve it either:

1. our code already "monitors" the directory  since `absoluteFilePath()` does the same as `absoluteDir()` and simply returns it in a different format.
 2. directoryChanged signals will only be sent when a file gets created or deleted, but not if it gets modified.

So the solution for this bug is probably way more complicated than necessary?

preisi: Also: yes, we have a theme updated with support for this feature. But keep in mind that we are not artistic when it comes to design. (It was only for testing purposes after all)

https://github.com/preisi/cip-theme/tree/adminInfo

Furthermore, the HTML is just a feature provided by the "Text QML Type". So no need to use it, normal text works, too.

plfiorini:
Closing the PR because the review comment of @davidedmundson wasn't resolved as far as I can tell, plus the patch is very old and needs a rebase.   You can continue working on it and propose it again later.

Issue #782, opened 2017-02-26
PAM Integration for user-specific messages on login


preisi:
Hello,
as part of a university linux project we (a team of 2 students) want to contribute to SDDM.

One of our goals is to add PAM Integration for user-specific messages on login.
(e.g. account suspension notifications or simply disk quota warnings)

What do you think about this planned functionality?

Cheers,
Thomas and Milan

P.S.: Of course we will develop this feature in the open on GitHub and give updates here.

preisi:
I'm currently looking at the integration of pam to display the messages on login.
As of now the greeter will be notified via signal that a message is available.

The only question right now would be: How to present the pam message?
I've created a component called PamMessage.qml which is some sort of dialog with an Ok-Button containing the message. But at the moment it will not be displayed if the user successfully logged on.

How should I solve this? should I simply use something else instead? E.g. a QMessageBox. Or should I send some sort of signal when "Ok" has been triggered and only then continue the login-process?

PR #802, opened 2017-03-26
PAM-Message Integration

milanstephan:
Displays PAM-messages to the user on successful login or failed login attempt.
Greeter stores multiple PAM-messages and displays one after another using "QMessageBox"es.

Signed-off-by: Milan Stephan <milan.stephan@fau.de>
Signed-off-by: Thomas Preisner <thomas.preisner@fau.de>

#782

### DIFF
--- a/src/greeter/GreeterApp.cpp
+++ b/src/greeter/GreeterApp.cpp
@@ -31,7 +31,7 @@
 
 #include "MessageHandler.h"
 
-#include <QGuiApplication>
+#include <QApplication>
 #include <QQuickItem>
 #include <QQuickView>
 #include <QQmlContext>
@@ -39,6 +39,7 @@
 #include <QDebug>
 #include <QTimer>
 #include <QTranslator>
+#include <QMessageBox>
 
 #include <iostream>
 
@@ -59,7 +60,7 @@ namespace SDDM {
 
     GreeterApp *GreeterApp::self = nullptr;
 
-    GreeterApp::GreeterApp(int &argc, char **argv) : QGuiApplication(argc, argv) {
+    GreeterApp::GreeterApp(int &argc, char **argv) : QApplication(argc, argv) {
         // point instance to this
         self = this;
 
@@ -161,7 +162,9 @@ namespace SDDM {
         view->engine()->addImportPath(QStringLiteral(IMPORTS_INSTALL_DIR));
 
         // connect proxy signals
-        connect(m_proxy, SIGNAL(loginSucceeded()), view, SLOT(close()));
+        connect(m_proxy, &GreeterProxy::loginSucceeded, this, &GreeterApp::pamShowCloseGreeter);
+        connect(m_proxy, &GreeterProxy::loginFailed, this, &GreeterApp::pamShow);
+        connect(m_proxy, &GreeterProxy::storePamMessage, this, &GreeterApp::pamStoreMessage);
 
         // we used to have only one window as big as the virtual desktop,
         // QML took care of creating an item for each screen by iterating on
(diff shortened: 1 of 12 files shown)


src/greeter/GreeterApp.h
```diff
@@ -65,6 +68,7 @@ namespace SDDM {
         UserModel *m_userModel { nullptr };
         GreeterProxy *m_proxy { nullptr };
         KeyboardModel *m_keyboard { nullptr };
+        QStringList m_pamMessages { };
```
davidedmundson: just QStringList
The others need it to initialise a pointer, that's not relevant here.

src/greeter/GreeterApp.h
```diff
@@ -50,6 +50,9 @@ namespace SDDM {
     private slots:
         void addViewForScreen(QScreen *screen);
         void removeViewForScreen(QQuickView *view);
+        void pamShow();
+        void pamShowCloseGreeter();
+        void pamStoreMessage(const QString message);
```
davidedmundson: (const QString &message)

otherwise we make a copy of the QString object when we pass it to the method.
It's a pretty cheap shallow-copy, but still, may as well avoid it whilst we can.

src/greeter/GreeterApp.cpp
```diff
@@ -231,6 +234,29 @@ namespace SDDM {
         view->deleteLater();
     }
 
+    void GreeterApp::pamStoreMessage(const QString message) {
+        m_pamMessages.append(message);
+    }
+
+    void GreeterApp::pamShow() {
+        while (!m_pamMessages.isEmpty()) {
+            QMessageBox *m = new QMessageBox(nullptr);
```
davidedmundson: this leaks

src/greeter/GreeterApp.cpp
```diff
@@ -231,6 +234,29 @@ namespace SDDM {
...
+    }
+
+    void GreeterApp::pamShowCloseGreeter() {
+        // closing greeter-views to show user's window manager
+        while (!m_views.isEmpty()) {
+            m_views.takeFirst()->close();
+        }
+
+        // display PAM messages
+        pamShow();
```
davidedmundson: so the timeline is:

 - we close the greeter
 - we spawn the user's session
 - we spawn dialogs as the SDDM user in the user's session

src/greeter/GreeterApp.h
```diff
@@ -38,7 +38,7 @@ namespace SDDM {
     class KeyboardModel;
 
 
-    class GreeterApp : public QGuiApplication
+    class GreeterApp : public QApplication
```
davidedmundson: I see why you're doing this, but I'd really like not to. 

QApplication is huge compared to QGuiApplication, it'll add half a second to bootup time easily.
and because you don't have a QtPlatformTheme loaded, it'll look hideous too.

Don't you have another patch that exposes messages from the sysadmin into the UI? Can't we share that?

We'll need a solution to stop the greeter from closing, lightDM used to solve that by not logging in when authentication was complete but instead waiting for a signal from the greeter.

src/greeter/GreeterApp.h
milanstephan: Oh, yes, fixed

src/greeter/GreeterApp.cpp
milanstephan: Thanks! Creating the (blocking) QMessageBox on the stack works as well, fixed.

src/greeter/GreeterApp.cpp
milanstephan: The user's session is spawned first (in the helper module) and will be visible as soon as the greeter is closed by Qt-signal.

src/greeter/GreeterApp.h
milanstephan: Simply delaying the closing of the greeter on the greeter's side (until all PAM messages are confirmed) works not as expected and causes strange looking behaviour like fade-in effects which will also affect the greeter when the user's session is started in the background (already tried that).

So the way to go would be to tell the helper to start the user's session only when all PAM messages are confirmed, which requires additional communication from the greeter to the helper, like lightDM does.

I don't think the Admin-Info box is suited that well, as these are two different components and we would have to implement a confirmation-button and reset-functionality for only one of them (and remove it for the other) and disable dynamic updates while PAM-messages are displayed to not overwrite the PAM-messages, which will affect readability of the code and I'm not sure whether the user will notice it as good this way (and not be confused as the Admin-Info box contains important information as well).

src/greeter/GreeterApp.h
milanstephan: fixed

src/greeter/GreeterApp.h
arwarw: Is there a way to "upgrade" a QGuiApplication to a QApplication only as soon as its needed? Or maybe create a separate Process for the MessageBoxes and wait for Termination of that?

SolarAquarion:
I'm pretty sure this needs to be rebased especially with auth.cpp

plfiorini:
Seems like there are still unresolved issues with the patch, but that is 2 years old now, it needs a rebase too.
Closing to clean up the queue.
Feel free to post new iteration of the patch if you feel like finishing it.

PR #814, opened 2017-04-19
Feature/hidpidisabled

plfiorini:


src/greeter/GreeterApp.cpp
```diff
@@ -247,11 +247,16 @@ int main(int argc, char **argv) {
     qInstallMessageHandler(SDDM::GreeterMessageHandler);
 
     // HiDPI
-    if (SDDM::mainConfig.EnableHiDPI.get()) {
+    bool hiDpiEnabled = false;
+    if (QGuiApplication::platformName() == QLatin1String("xcb"))
```
jleclanche: Do we really not have a helper to get the current platform?

plfiorini: an alternative is to make assumptions based on env vars but this way we are sure to check the correct platform

PR #819, opened 2017-05-05
Also theme the default cursor for the root window

Vogtinator:
The root window is still accessible in some cases, like context menu overlays.

### DIFF
--- a/src/daemon/XorgDisplayServer.cpp
+++ b/src/daemon/XorgDisplayServer.cpp
@@ -249,7 +249,7 @@ namespace SDDM {
         env.insert(QStringLiteral("SHELL"), QStringLiteral("/bin/sh"));
         displayStopScript->setProcessEnvironment(env);
 
-        // start display setup script
+        // start display stop script
         qDebug() << "Running display stop script " << displayStopCommand;
         displayStopScript->start(displayStopCommand);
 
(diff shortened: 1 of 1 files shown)


Vogtinator:
I replaced the patch with a version that does not require Xsetup changes.

src/daemon/XorgDisplayServer.cpp
```diff
@@ -285,11 +285,21 @@ namespace SDDM {
         env.insert(QStringLiteral("PATH"), mainConfig.Users.DefaultPath.get());
         env.insert(QStringLiteral("XAUTHORITY"), m_authPath);
         env.insert(QStringLiteral("SHELL"), QStringLiteral("/bin/sh"));
+        env.insert(QStringLiteral("XCURSOR_THEME"), mainConfig.Theme.CursorTheme.get());
         displayScript->setProcessEnvironment(env);
 
         // delete displayScript on finish
         connect(displayScript, SIGNAL(finished(int,QProcess::ExitStatus)), displayScript, SLOT(deleteLater()));
```
davidedmundson: This can't be safe.

We're calling deleteLater() when we finish xsetroot  and then starting X with that QProcess

Vogtinator: Should be fixed now

PR #834, opened 2017-06-23
Honor PAM's ambient supplemental groups (pam_group.so)

### LINKED ISSUE #414: Kerberos/LDAP; how to respect /etc/security/group.conf ?
Question:
What has to be adjusted to make SDDM respect /etc/security/group.conf?

Using Kerberos/LDAP on a Kubuntu install: Login in via SDDM seems to ignore the /etc/security/group.conf file. Local groups (like "audio", "cdrom") need to be added on-the-fly when logging-in.

Does SDDM support this?

jktr:
This patch adds support for PAM's pam_group.so(8) module and fixes #414.

SDDM currently uses initgroups(3) in a PAM context. This is problematic when dynamically adding groups to the local host's session: while PAM correctly loads the groups into a session from /etc/security/groups.conf via pam_groups.so, SDDM's initgroups call then throws these away just before the setguid step.

This fix adds PAM's ambient groups to the set of user groups in a session by explicitly dealing with each set of groups. The old behavior is retained when compiling without USE_PAM, due to ambiguous group context when not running under PAM.

Disclosure:
The patch was originally developed by me for Paderborn University;  I have been given permission to contribute it under GPLv2+ to this project. We are currently using this patch in production.

### DIFF
--- a/src/helper/UserSession.cpp
+++ b/src/helper/UserSession.cpp
@@ -152,11 +152,70 @@ namespace SDDM {
             free(buffer);
             exit(Auth::HELPER_OTHER_ERROR);
         }
+
+#ifdef USE_PAM
+
+        // fetch ambient groups from PAM's environment;
+        // these are set by modules such as pam_groups.so
+        int n_pam_groups = getgroups(0, NULL);
+        gid_t *pam_groups = NULL;
+        if (n_pam_groups > 0) {
+            pam_groups = new gid_t[n_pam_groups];
+            if ((n_pam_groups = getgroups(n_pam_groups, pam_groups)) == -1) {
+                qCritical() << "getgroups() failed to fetch supplemental"
+                            << "PAM groups for user:" << username;
+                exit(Auth::HELPER_OTHER_ERROR);
+            }
+        } else {
+            n_pam_groups = 0;
+        }
+
+        // fetch session's user's groups
+        int n_user_groups = 0;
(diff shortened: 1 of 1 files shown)


jktr:
The original patch had erroneously made the `getgrouplist` part dead code…
which I didn't catch due to our peculiar production setup.
This version should fix that part of the logic. Sorry about that.

tbrethom:
We've also fixed this issue myself in my environment by just moving the initgroups call before pam_open_session. See the attached patch.

[0004-Fix-setgroups-call-after-setcred.patch.txt](https://github.com/sddm/sddm/files/1376027/0004-Fix-setgroups-call-after-setcred.patch.txt)

jktr:
Hi again! It's been some time since I've last touched this, and some merge conflicts had accumulated. I've rebased the patch onto the current state of `sddm/sddm:develop`, as I'm still interested in seeing this merged, and it seems to apply cleanly again now. (The previous patch is still available as `jktr/sddm:fix/pamgroup-v2`)

That said, I'll run some fresh builds in our pam_group.so-dependent environment at work (some time soon) to make sure this still does what it's supposed to.

On another note, while this compiles fine on my archlinux dev box using gcc, with and without PAM, I'm not quite sure why the travis build fails. (XCB package missing from CI env!?).
I'd appreciate some pointers if it's something I need to change on my end.

Issue #606, opened 2016-03-12
RFE: store xauthority in system location


rdieter:
In some circumstances, storing .Xauthority in $HOME is problematic (like NFS $HOME), so consider implementing storing this in some system location instead, perhaps utilizing sddm's RUNTIME_DIR or $XDG_RUNTIME_DIR

I know UserAuthFile= has been implemented, but unfortunately, that's relative to $HOME, so wouldn't help this use-case.

PR #841, opened 2017-07-07
Do not truncate XAUTHORITY file on login

Vogtinator:
xauth gets used to edit the file, so do not truncate the file with QIODevice::WriteOnly beforehand.

Otherwise the ```remove %1``` command to xauth would be useless.


jktr:
Thanks for the patch!
We have NFS-mounted homes at our installation, so truncating XAUTHORITY was causing X sessions to hang when logging into a second computer with the same home directory mounted.
I'll try this out and see if this works as a workaround for #606.

jktr:
I got around to testing this out, and it does indeed fix the NFS issues mentioned in #606.

PR #842, opened 2017-07-09
Add LIBXCB_INCLUDE_DIR to the list of include directories.

rakuco:
The daemon and the greeter include xcb/xcb.h, but they were relying on xcb's
include directory being implicitly added by the compiler instead of making
sure it was always passed to the compiler.


rakuco:
@davidedmundson could you take a look at this one too?

rakuco:
Oops, you're right. Patch updated.

rakuco:
Thanks for reviewing the PR. I don't have write access to the repository, so I'd appreciate it if you could merge the PR.

PR #846, opened 2017-07-19
Update .ts files and CMakeLists.txt

m4sk1n:


data/translations/pt_PT.ts
```diff
@@ -2,6 +2,20 @@
 <!DOCTYPE TS>
 <TS version="2.1" language="pt_PT">
 <context>
+    <name>PictureBox</name>
+    <message>
+        <source>Press to login</source>
+        <translation type="unfinished"></translation>
```
davidedmundson: what does this accomplish?

m4sk1n: I just updated files to current sources, translators can update them without having to use lupdate.

plfiorini: seems like it just adds a new (untranslated) string to all translation files so nothing tangible will happen. probably best if translators will rebase their translations separately?

data/translations/CMakeLists.txt
```diff
@@ -12,22 +12,26 @@ set(TRANSLATION_FILES
...
+    ru.ts
+    sk.ts
     sr.ts
     sr@ijekavian.ts
     sr@ijekavianlatin.ts
     sr@latin.ts
     sv.ts
-    ru.ts
     tr.ts
+    uk.ts
```
plfiorini: we want this part here for sure since the files were not added to cmake

data/translations/pt_PT.ts
m4sk1n: @plfiorini, I also fixed Ukrainian translation and added some translation files to CMakeFiles (yes, they were committed without change in CMakeFiles…)

plfiorini: there is a build failure now:

https://travis-ci.org/sddm/sddm/jobs/255215522#L1208

m4sk1n: @plfiorini, already fixed

Issue #733, opened 2016-10-31
RFE: support xauth/FamilyWild and/or xhost/SeverInterpreted


rdieter:
See also downstream report,
https://bugzilla.redhat.com/show_bug.cgi?id=1370222#c25

In fedora, we've had issues where a machine changing hostname (during or after boot) would sometimes invalidate xauth data, causing sessions and applications to fail to start.  The following is a suggestion from a fedora/gnome gdm developer:

------------------------------------------

sddm should be creating FamilyWild type xauth entries if it isn't already. They're immune to hostname changes.  Also sddm could set up xhost ServerInterpeted entries for the localuser, so the xauth file isn't required.

Here's some code to follow for setting up FamilyWild:

https://git.gnome.org/browse/gdm/tree/daemon/gdm-x-session.c#n107

Note the auth entry is written twice, but one line 145 the family is changed to FamilyWild.

Here's some code for setting up xhost ServerInterpreted entries:

https://git.gnome.org/browse/gdm/tree/daemon/gdm-display.c#n1695

plfiorini:
@rdieter I'll try to do something over the week end unless I have to attend to my day job as it often happens, so no promises...

gbcox:
In the meantime, people who encounter this issue most likely have their /etc/hostname file using the default of:  localhost.localdomain
and have their dhcp server assigning a different value... i.e. my_host
The workaround would be to (as root) edit the value in /etc/hostname to match.  In other words, remove the value of localhost.localdomain in /etc/hostname and repflace it with:  my_host

CRCinAU:
Just wanted to give this a ping.....

PR #863, opened 2017-08-15
Use libXau instead of xauth

MartinBriza:
Mostly work in progress - daemon parts are not done yet.
Should potentially fix issues with hostname changes on bootup when SDDM refused to start (need to test if that's fixed already or not).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,6 +92,9 @@ find_package(XCB REQUIRED)
 # XKB
 find_package(XKB REQUIRED)
 
+# XKB
+find_package(XAU REQUIRED)
+
 # Qt 5
 find_package(Qt5 5.6.0 CONFIG REQUIRED Core DBus Gui Qml Quick LinguistTools)
 
--- a/src/daemon/XorgDisplayServer.cpp
+++ b/src/daemon/XorgDisplayServer.cpp
@@ -94,11 +94,6 @@ namespace SDDM {
 
         qDebug() << "Adding cookie to" << file;
 
-        // Touch file
-        QFile file_handler(file);
-        file_handler.open(QIODevice::Append);
-        file_handler.close();
-
         if (gethostname(localhost, HOST_NAME_MAX) < 0) {
             strcpy(localhost, "localhost");
         }
--- a/src/daemon/XorgDisplayServer.cpp
+++ b/src/daemon/XorgDisplayServer.cpp
@@ -34,6 +34,7 @@
 #include <random>
 
 #include <xcb/xcb.h>
+#include <X11/Xauth.h>
 
 #include <pwd.h>
 #include <unistd.h>
@@ -88,7 +89,9 @@ namespace SDDM {
     }
 
     void XorgDisplayServer::addCookie(const QString &file) {
-        // log message
+        Xauth auth = { 0 };
+        char localhost[HOST_NAME_MAX + 1] = { 0 };
+
         qDebug() << "Adding cookie to" << file;
 
         // Touch file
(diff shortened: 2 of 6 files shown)


MartinBriza:
Related issue: #733

Vogtinator:
With these patches ```m_display``` and ```display``` aren't used anymore - intentional?

tpgxyz:
@MartinBriza  Hi, does this resolves issue:

```
Oct 23 17:05:16 openmandriva sddm[3991]: /usr/bin/xauth: (stdin):1:  bad "remove" command line
Oct 23 17:05:16 openmandriva sddm[3991]: /usr/bin/xauth: (stdin):2:  bad "add" command line
Oct 23 17:05:16 openmandriva sddm[3991]: QProcess: Destroyed while process ("/usr/libexec/sddm-helper") is still running.
Oct 23 17:05:16 openmandriva acpid[3838]: client 7257[0:0] has disconnected
```

This happens when updating systemd :(

src/daemon/XorgDisplayServer.cpp
```diff
@@ -88,28 +89,55 @@ namespace SDDM {
     }
 
     void XorgDisplayServer::addCookie(const QString &file) {
-        // log message
+        Xauth auth = { 0 };
```
Vogtinator: This will only initialize the first member.

src/daemon/XorgDisplayServer.cpp
```diff
@@ -88,28 +89,55 @@ namespace SDDM {
...
+        // libXau expects binary data, not a string
+        QByteArray cookieBinary = QByteArray::fromHex(m_cookie.toLatin1());
+
+        // set up the auth entry
+        char cookieName[] = "MIT-MAGIC-COOKIE-1";
+        auth.family = FamilyLocal;
+        auth.address = localhost;
+        auth.address_length = strlen(auth.address);
+        auth.name = cookieName;
+        auth.name_length = sizeof(cookieName);
```
Vogtinator: This doesn't work from what I can tell - you'll have to use `strlen` or `sizeof(...) - 1` to exclude the nullbyte at the end.

src/daemon/XorgDisplayServer.cpp
```diff
@@ -88,28 +89,55 @@ namespace SDDM {
...
+        auth.data_length = cookieBinary.count();
+        auth.data = cookieBinary.data();
+
```
Vogtinator: Missing initialization of `auth.number{,_length}`, intentional? It should be filled with `m_display` (excluding the `:`).

src/daemon/XorgDisplayServer.cpp
```diff
@@ -88,28 +89,55 @@ namespace SDDM {
...
+        // create the path
+        QFileInfo finfo(file);
+        QDir().mkpath(finfo.absolutePath());
+
+        // open the file
+        FILE *fp = fopen(qPrintable(file), "w");
```
Vogtinator: Shouldn't technically be necessary, `"wb"` for binary mode.

CMakeLists.txt
```diff
@@ -92,6 +92,9 @@ find_package(XCB REQUIRED)
 # XKB
 find_package(XKB REQUIRED)
 
+# XKB
```
Vogtinator: Wrong comment

plfiorini:
Ping @MartinBriza

src/daemon/XorgDisplayServer.cpp
aaronpuchert: It will value-initalize the remaining members. This is an [aggregate initializer](https://en.cppreference.com/w/cpp/language/aggregate_initialization): "If the number of initializer clauses is less than the number of members [...] or initializer list is completely empty, the remaining members [...] are initialized [...] by empty lists, in accordance with the usual list-initialization rules (which performs value-initialization for non-class types and non-aggregate classes with default constructors, and aggregate initialization for aggregates)." [Value initialization](https://en.cppreference.com/w/cpp/language/value_initialization) means "if T is an array type, each element of the array is value-initialized; otherwise, the object is zero-initialized."

The same effect could be achieved with `Xauth auth{};` which is not the same as `Xauth auth;` — the latter does [default initialization](https://en.cppreference.com/w/cpp/language/default_initialization): when no default constructor is available, "nothing is done: the objects with automatic storage duration (and their subobjects) are initialized to indeterminate values."

Vogtinator: C++ has too many initialization methods - the "unified initialization" just made everything more complicated...

I'd very much prefer `Xauth auth{};` here.

Vogtinator:
@MartinBriza Still planning to work on this?
Currently the code won't actually work.

rdieter:
I (still) argue the effects of not fixing issue #733 is worse than the possible side effects mentioned.

That said, those other issues mentioned are worth fixing too (but imo) should not be considered a blocker here.

Vogtinator:
Superseded by #1230

PR #870, opened 2017-08-26
Stop assuming shadow(5) support is always available.

rakuco:
shadow and getspnam(3) are Linux-specific. Add a check for getspnam(3) to
CMake and only use the shadow functions if they exist, otherwise just use
getpwnam(), which on systems such as FreeBSD already performs the same
things shadow does on Linux.


rakuco:
I've rebased the patch now that #869 has been merged.

PR #877, opened 2017-09-02
Explicity set XDG_SEAT env var when starting the user session.

davidedmundson:
Removed during the logind seat tracking. I think I did it deliberately as we passed the seat to X..but it was so long ago I can't remember.

In any case, it seems to be needed. If I login, loginctl list-sessions has the right stuff.


Vogtinator:
Better, but not quite there yet, the tty is missing from the loginctl output still.

Vogtinator:
Nope, that's actually only required for wayland sessions where this works fine. So this looks alright.

PR #880, opened 2017-09-04
Added possibility to change border width of ComboBox widget.

kscd:
I added the possibility to change the border width of the ComboBox widget to an arbitrary integer. The default value is set to 1, the former standard value.

### DIFF
--- a/components/2.0/ComboBox.qml
+++ b/components/2.0/ComboBox.qml
@@ -35,6 +35,7 @@ FocusScope {
     property color menuColor: "white"
     property color textColor: "black"
 
+    property int borderWidth: 1
     property font font
     property alias model: listView.model
     property int index: 0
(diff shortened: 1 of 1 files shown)


components/2.0/ComboBox.qml
```diff
@@ -148,7 +151,7 @@ FocusScope {
         id: dropDown
         width: container.width; height: 0
         anchors.top: container.bottom
-        anchors.topMargin: -1
+        anchors.topMargin: 0
```
davidedmundson: This seems unrelated to borders.

Seems somewhat sensible anyway; make sure you test that with a widget with border 1.

kscd: I tested the change with a borderwidth of 1 and it looks how it should, for comparison I included a picture with a borderwidth of 5. The entries have smaller space because the overall height remains constant in both cases.

The 'anchors.topMargin: -1' prevented the line between the uppermost text and the list view from having a width of 2 pixel, but due to my code modifications this isn't an issue anymore.

borderwidth_5
borderwidth_1

Issue #868, opened 2017-08-24
Not switching to awesome WM even after selecting it


amitavmohanty01:
I had only KDE Plasma window manager which was getting selected when I was logging in. I installed awesome WM. There is an entry for it at ```/usr/share/xsessions/awesome.desktop```. The option shows up in SDDM login screen too. However, when I select the awesome option and try logging in, I get logged into Plasma. There are no errors in ```~/.xsession-errors```. Also, ```/etc/sddm.conf``` does not have any entry for Plasma so it should not be default.

kscd:
Which sddm-Version are you using on which distribution? I just tested to login into awesome under Debian 9 (stretch) with sddm 0.14.0-4 and it works perfectly.

Regarding Plasma being default: I believe, the session which was used the last time is pre-selected per default.

amitavmohanty01:
I am using 0.14.0-3 on Arch linux. Is there any log that I can check and provide for debugging my issue better?

amitavmohanty01:
I did a fresh reboot. When I reached SDDM screen, I selected awesome but again I was in Plasma. The log line is as follows:

``` Session "/usr/share/xsessions/plasma.desktop" selected, command: "/usr/bin/startkde"```

kscd:
I have currently a theory of what might happen when you try to log in in awesome: maybe your log in theme is broken so that it can't tell sddm which WM to use, causing sddm to fall back to Plasma (happened to me, while I was writing my own theme). Which log in theme are you currently using? It is stated in the file /etc/sddm.conf. If this file does not exist, the default theme 'maui' will be used.

To be very sure that awesome has nothing to do with it: could you try to install a third WM and try to log in to it?

amitavmohanty01:
Yeah, it is not a problem with awesome WM. I switched to LightDM and I am able to login to awesome WM from that.

I am not on my machine currently. I will update you soon.

amitavmohanty01:
The theme selected is elarun. I think I switched to this one because some time back 'maui' was broken.

kscd:
Turned out I was partly wrong. The 'elarun' theme uses both a Menu widget and a ComboBox widget for the same purpose, which makes no sense. The fix should be easy enough, but I can't test it right now.

PR #888, opened 2017-09-11
Remove Menu widget from elarun theme

kscd:
The elarun theme is currently broken. When the user selects a session in the ComboBox widget the selected entry will be ignored, but the selected entry from a not-visible Menu widget will be used.

In this patch the Menu widget gets removed, while the ComboBox widget is used correctly.

This patch fixes issue #868.


davidedmundson:
This is now redundant right? (someone else made a similar patch at the same time as you)

kscd:
Yes, I agree with you that this pull request is now redundant and therefore useless (although the other PR was opened 7 months ago).

PR #923, opened 2017-10-30
[WIP] Better support for ConsoleKit2 interfaces

awilfox:
When ConsoleKit2/ConsoleKit2#102 is merged, ListSeats will be implemented in a way that SDDM expects.  However, SDDM's support is lacking in a few other ways:

* Random / character at the start of the interface class name
* SeatNew vs SeatAdded signal name (the interfaces are the same)

What works with this PR (and CK2 patch):

* Seats
* Input devices
* Reboot/sleep/shutdown buttons on the login screen

Problems still present in this PR:

* It doesn't seem to be opening a new session, though it is correctly using the seats given.
* The canGraphical hack I had to write is kind of ugly; ConsoleKit does not have such a property on seats.
* Maybe something else?  Needs more testing.

### DIFF
--- a/src/daemon/SeatManager.cpp
+++ b/src/daemon/SeatManager.cpp
@@ -26,6 +26,7 @@
 #include <QDBusMessage>
 #include <QDBusPendingReply>
 #include <QDBusContext>
+#include <QDebug>
 
 #include "LogindDBusTypes.h"
 
@@ -59,6 +60,12 @@ namespace SDDM {
         QDBusPendingCallWatcher *watcher = new QDBusPendingCallWatcher(reply);
         connect(watcher, &QDBusPendingCallWatcher::finished, this, [=]() {
             watcher->deleteLater();
+            if (Logind::serviceName().contains(QStringLiteral("ConsoleKit"))) {
+                m_canGraphical = true;
+                emit canGraphicalChanged(m_canGraphical);
+                return;
+            }
+
             if (!reply.isValid())
                 return;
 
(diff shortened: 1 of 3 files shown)


src/daemon/SeatManager.cpp
```diff
@@ -59,6 +60,12 @@ namespace SDDM {
         QDBusPendingCallWatcher *watcher = new QDBusPendingCallWatcher(reply);
         connect(watcher, &QDBusPendingCallWatcher::finished, this, [=]() {
             watcher->deleteLater();
+            if (Logind::serviceName().contains(QStringLiteral("ConsoleKit"))) {
```
davidedmundson: If we're going to do this, we should move it earlier and skip
 QDBusPendingReply<QVariant> reply = QDBusConnection::systemBus().asyncCall(canGraphicalMsg);

awilfox: I tried that earlier, and the problem is that it needs time for `logindSeatAdded` to watch the property.  Otherwise, the main thread never sees the property change as there is nothing listening for that signal's emission.

Look around [src/daemon/SeatManager.cpp:155](https://github.com/sddm/sddm/blob/develop/src/daemon/SeatManager.cpp#L155).

davidedmundson: Then change SeatManager.cpp

travankor:
@davidedmundson @awilfox 

Any progress on this? Void Linux is on an outdated sddm, pending a better consolekit2 interface.

awilfox:
There has been no progress on this from our side.  I still have interest and intend to work on this at some point, but I've been more focused on kernel-level stuff with amdgpu before I can continue with hacking on sddm.

Might just go for a test KVM with Qemu/Cirrus graphics so I can get something out the door, if there really is a need here…

Hoshpak:
I am currently running sddm 0.18.0 with your changes applied before compilation and it works at least as well as 0.15.0 did before 0.16.0 broke with ConsoleKit2. sddm starts, the shutdown and reboot buttons work and I can log in and start a lxqt session. No ConsoleKit session is created but you already mentioned that in the persisting problems.

From my point of view there definitly is a need since there are people out there who don't want to use (e)logind for one reason or another and ConsoleKit2 currently is a working alternative. As @travankor already mentioned, Void Linux is currently stuck on 0.15.0 since updating sddm would break existing user setups.

a17r:
Any update here?

awilfox:
Sorry, not yet.  I'm still just one fox doing the role of 10 people.  Display manager is unfortunately very very low on the priority list for our distro since most of our users just use `startx` anyway.  I do still intend to fix this, but it will probably not see any work from me until after our 1.0 release.  Currently, that's projected for December.

awilfox:
I definitely don't mind if others pick this up, but as of 2020 we are 100% elogind and I no longer have any way to test or update this.

Vogtinator:
> I definitely don't mind if others pick this up, but as of 2020 we are 100% elogind and I no longer have any way to test or update this.

So let's close this - if someone else wants to pick this up, a new PR can be based off of this.

PR #934, opened 2017-11-10
Switch to last display on logout if there is still a session running on some display

roberthoffmann:
This fixes the following bug:
1. Login user A
2. Switch user
3. New session: login user B
4. Switch to user A
5. Logout user A
=> Virtual terminal is not switched to session of user B,
   but shows a text terminal login or is blank.

VirtualTerminal::chVt code inspired by chvt utility.

This code works with Xorg sessions. Wayland sessions seem
to have other problems when logging out, this fix does not
address these.

Tested on Ubuntu bionic.


src/daemon/VirtualTerminal.cpp
```diff
@@ -173,5 +173,45 @@ namespace SDDM {
 
             close(activeVtFd);
         }
+
+        void chVt(int vt) {
```
davidedmundson: how is this different from jumpToVt?

roberthoffmann: I tried jumpToVt before doing the chVt, but jumpToVt freezes the display if called on logout from Xorg sessions.

jumpToVt seems to be called only at the start of Wayland sessions and does some graphics mode setting, perhaps that's the reason for the freeze.

roberthoffmann:
I removed chVt() and integrated the VT switching fix into the existing jumpToVt() method.

The problem with jumpToVt() was that it set the VT switching mode to manually (VT_PROCESS). But after the end of a session there is no VT controlling process anymore on that ended VT which could receive the kernel signal and release the switch.

Now I call jumpToVt() with an additional boolean parameter telling it to use (default) automatic mode (VT_AUTO).

The only other already existing call to jumpToVt() (creating a display with Wayland session) is done with 'false', so it uses manual mode like before.

PR #935, opened 2017-11-10
Added method 'executeCommand' to GreeterProxy.

roberthoffmann:
This allows themes e.g. to provide ActionButtons which run
arbitrary configured commands when clicked:

    ActionButton {
        iconSource: "configure"
        text: i18nd("plasma_lookandfeel_org.kde.lookandfeel","Change Password")
        onClicked: sddm.executeCommand("/usr/share/myclient/change_password")
        enabled: true
        visible: !inputPanel.keyboardActive
    }

Tested on Ubuntu bionic.

### DIFF
--- a/src/greeter/GreeterProxy.cpp
+++ b/src/greeter/GreeterProxy.cpp
@@ -108,6 +108,13 @@ namespace SDDM {
         SocketWriter(d->socket) << quint32(GreeterMessages::HybridSleep);
     }
 
+    void GreeterProxy::executeCommand(const QString &cmd) {
+        QByteArray byteArray = cmd.toUtf8();
+        const char *charCmd = byteArray.constData();
+
+        system(charCmd);
+    }
+
     void GreeterProxy::login(const QString &user, const QString &password, const int sessionIndex) const {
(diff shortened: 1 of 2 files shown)


davidedmundson:
I don't get really get your example.
You'd be running that as the SDDM user, and you can't go round opening windows with prompts and stuff as there's no window manager running. 

I'm not sure about the concept. It's a difficult balance between needing to support sysadmins needing to extend the UI in every manner necessary, and the fact that we have desktop users grabbing themes that are supposedly pure UI off the internet which shouldn't have the scope to introduce security holes. 

BTW, you can do it without any SDDM changes by just shipping your own QML imports.

roberthoffmann:
It allows us to open a dialog (some kind of minimalistic browser) from the greeter where users can change their LDAP passwords without having to log in. A window manager isn't needed for that.

plfiorini:
Yes it is needed to control the window position and decorate the client.
I'm with @davidedmundson, this doesn't feel secure to me nor appropriate for a login manager.

plfiorini:
Rejected.

PR #937, opened 2017-11-11
[RFC] Fix connection of PropertiesChanged event for LogindSeat

h4nn35:
SDDM fails to start properly because the seat0 changes the property
canGraphical after SDDM is started and the connection to watch the
PropertiesChanged event didn't work.

### DIFF
--- a/src/daemon/SeatManager.cpp
+++ b/src/daemon/SeatManager.cpp
@@ -40,7 +40,7 @@ namespace SDDM {
     Q_SIGNALS:
         void canGraphicalChanged(bool);
     private Q_SLOTS:
-        void propertiesChanged(const QString &interface, QVariantMap &changedProperties , const QStringList &invalidatedProperties);
+        void propertiesChanged(const QString &interface, const QVariantMap &changedProperties , const QStringList &invalidatedProperties);
     private:
         QString m_name;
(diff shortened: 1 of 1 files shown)


davidedmundson:
I don't see how this changes anything?

h4nn35:
For me this makes the binding src/daemon/SeatManager.cpp:53
`QDBusConnection::systemBus().connect(Logind::serviceName(), objectPath.path(), QStringLiteral("org.freedesktop.DBus.Properties"), QStringLiteral("PropertiesChanged"), this, SLOT(propertiesChanged(QString,QVariantMap,QStringList)));`
 return true, which otherwise failes.
After adding this my PC with a radeon graphics can but up properly again.

davidedmundson:
That definitely shouldn't return false. You might be onto something.

Can you check if changing changedProperties to 

const &changedProperties
(added const)

also works?

h4nn35:
Yes, this works too.

davidedmundson:
Urgh, thanks for finding this. 
If you want to submit that, go for it, otherwise I'll do it

h4nn35:
I have updated the commit to add const instead of removing the reference.

davidedmundson:
Thanks

PR #940, opened 2017-11-16
Set a default InputMethod

aleixpol:
Otherwise we choke by default when importing QtVirtualKeyboard

### DIFF
--- a/src/common/Configuration.h
+++ b/src/common/Configuration.h
@@ -42,7 +42,7 @@ namespace SDDM {
         Entry(Numlock,             NumState,    NUM_NONE,                                       _S("Initial NumLock state. Can be on, off or none.\n"
                                                                                                    "If property is set to none, numlock won't be changed\n"
                                                                                                    "NOTE: Currently ignored if autologin is enabled."));
-        Entry(InputMethod,         QString,     QString(),                                      _S("Input method module"));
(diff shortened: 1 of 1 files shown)


plfiorini:
@aleixpol I'm not sure, with this change the virtual keyboard will always be in use by themes that implement it but that is not what you'd expect on desktop systems

davidedmundson:
> but that is not what you'd expect on desktop systems

At which point the user has to change the default, which is the same as before this patch.

plfiorini:
Before this patch the user had to explicitely configure SDDM to enable the virtual keyboard which is safe on desktop systems (the vast majority of systems). On a mobile system I assume the distro would provide its own SDDM settings to enable the virtual keyboard.

davidedmundson:
How is it not safe on a desktop system?
1) all we're doing is export an env var. 
2) on a desktop system with a touch screen we want this anyway.

aleixpol:
We are not talking about mobile specifically, on a convertible laptop you want to be able to log in without fetching the keyboard. And I haven't seen "Debian for convertibles" just yet...

In fact, the motivation of this PR is Angel's issue that consists of systems from a city council that want to have the same distro for every system, and some of them have touch screens, because today touch screens are normal on "desktop" systems.

plfiorini:
@davidedmundson safe => i meant that virtual keyboards on traditional desktop systems are annoying

@aleixpol Fair enough I didn't take convertibles and today's touch screens into account.

Issue #944, opened 2017-11-27
SDDM needs to use a different location for Xauthority


Vogtinator:
The hostname of a system can change, for instance if it's automatically configured with DHCP and a connection changes. If it changes during a running X11 session, started clients try to connect using the new hostname and get rejected by the server.
By exporting XAUTHLOCALHOSTNAME to the hostname that was written into the auth cookie, clients always use the correct hostname for connecting.

Vogtinator:
There's a second bug that prevents this from being as useful as it should be - entries for old hostnames are still kept in XAUTHORITY (since https://github.com/sddm/sddm/commit/72175056a0af1b523e92861e75d602fb740dc238). Unfortunately those get preferred over the new cookie and auth fails.

Vogtinator:
I talked with some X11 experts on this and the best solution is to:

* Truncate the Xauthority file again on login
* Use a different location for Xauthority by default (which won't be written to outside of sddm)

As it's not possible to login twice as the same user, the file truncation then does not matter anymore.

Any thoughts/opionions about that?

davidedmundson:
>As it's not possible to login twice as the same user,

it is possible

Vogtinator:
Shouldn't it reuse the existing session instead?

davidedmundson:
I had bugs on KDM which didn't saying it should.
I had bugs on LightDM which did saying it shouldn't.

SDDM didn't till last release which made it a config option.

Vogtinator:
Current SDDM releases (which don't include my "Do not truncate .Xauthority" commit) would nuke other X sessions' cookies instantly during login, so I would say that logging in twice never worked on any released sddm.

Vogtinator:
What would work in all cases: Create a Xauthority file in /run/sddm and make it readable by the user only. Then that file can be used as $XAUTHORITY.

mpictor:
@Vogtinator since you closed #947, should this be closed as well?

Vogtinator:
Having .Xauthority in .local/share/sddm is not the best option (it should be either .Xauthority or somewhere in /run) but I don't think it causes any issues anymore, so yes.

PR #947, opened 2017-11-30
Move Xauthority to a different location and truncate it again

Vogtinator:
When the hostname changes, started X applications try the new hostname
first. If a cookie is found for that new hostname, they try to authenticate
(which fails) and exit. So .Xauthority must not contain old cookies.
Truncating ~/.Xauthority breaks ssh forwarding though, so the default
location is changed to something sddm specific.
As it's not possible to login twice as the same user simultaneously,
a fixed path at a fixed location is enough.

Issue #944


Vogtinator:
For reference, downstream bugreport about truncating .Xsession breaking ssh -X: https://bugzilla.suse.com/show_bug.cgi?id=1043221
Downstream bugreport about hostname changes breaking the authentication: https://bugzilla.suse.com/show_bug.cgi?id=1069498

src/common/Configuration.h
```diff
@@ -65,7 +65,7 @@ namespace SDDM {
             Entry(SessionDir,          QString,     _S("/usr/share/xsessions"),                 _S("Directory containing available X sessions"));
             Entry(SessionCommand,      QString,     _S(SESSION_COMMAND),                        _S("Path to a script to execute when starting the desktop session"));
 	    Entry(SessionLogFile,      QString,     _S(".local/share/sddm/xorg-session.log"),   _S("Path to the user session log file"));
-	    Entry(UserAuthFile,        QString,     _S(".Xauthority"),                          _S("Path to the Xauthority file"));
+	    Entry(UserAuthFile,        QString,     _S(".local/share/sddm/.Xauthority"),        _S("Path to the Xauthority file"));
```
eliasp: Instead of hardcoding `~/.local/share/` - shouldn't this be determined through `QStandardPaths::DataLocation`?

Vogtinator: It's the same directory as the logfile location entry, which is hardcoded as well.

Vogtinator: I mean that the default value is hardcoded.

Vogtinator:
I got some positive user feedback, so I removed the WIP flag.

rdieter:
fyi, pull request 863 (Use libXau instead of xauth), will probably solve the problem outlined here too

Vogtinator:
> fyi, pull request 863 (Use libXau instead of xauth), will probably solve the problem outlined here too

It doesn't, there's no difference between xauth and libXau in that regard. Whether you call an external program to assmeble the cookie or a lib doesn't matter.

To really fix the hostname issue, libxcb needs to be patches like done in openSUSE: https://build.opensuse.org/package/view_file/home:michalsrb:branches:X11:XOrg/libxcb/n_If-auth-with-credentials-for-hostname-fails-retry-with-XAUTHLOCALHOSTNAME.patch?expand=1

rdieter:
libXau's support for FamilyWild should be (more) resilient to hostname changes (it is for me).  If that's not what you're talking about, pardon my comments

Vogtinator:
> If that's not what you're talking about, pardon my comments

It kind of is.

This needs to be merged or FamilyWild will just break as more than one FamilyWild entry does not make sense.

Vogtinator:
I'll close this now - this was only a workaround for a bug in libxcb.

PR #984, opened 2018-02-28
Add support for theme supplied default avatars.

Sogatori:
This patch adds support for custom default avatars under $ThemeDir/themeName/faces.

This fixes the issue that by default sddm will use non-breeze avatar icons, which look quite different compared to breeze icons.

Without the patch:
Login screen before the patch
compared to the lockscreen:
Lockscreen

After the patch with a custom icon in breeze/faces:
Login screen after the patch

### DIFF
--- a/src/greeter/UserModel.cpp
+++ b/src/greeter/UserModel.cpp
@@ -95,7 +95,7 @@ namespace SDDM {
 
             // search for face icon
             if (QFile::exists(themeDefaultFace)) 
-                user->icon =QStringLiteral("file://%1").arg(themeDefaultFace);
+                user->icon = QStringLiteral("file://%1").arg(themeDefaultFace);
             else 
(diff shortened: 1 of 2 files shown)


Sogatori:
Oops, I didn't want to do that merge. I am new to github. I thought it would just check if it was still mergable. Sorry :<

src/greeter/UserModel.cpp
```diff
@@ -91,7 +94,10 @@ namespace SDDM {
             user->needsPassword = strcmp(current_pw->pw_passwd, "") != 0;
 
             // search for face icon
-            user->icon = defaultFace;
+            if (QFile::exists(themeDefaultFace)) 
+                user->icon =QStringLiteral("file://%1").arg(themeDefaultFace);
```
plfiorini: nitpick: put a space after equal sign

plfiorini:
Don't worry I will squash before merging

PR #989, opened 2018-03-06
[WIP] Add optional sanitization of screen scale factors

Vogtinator:
This is inspired by KScreens detection of the proper screen scale factors.
It should catch common errors in monitor EDIDs while still allowing most HiDPI
displays to use the appropriate scale.

Pros:
- Should allow `EnableHiDPI` to be enabled by default
- Normally no manual editing of `ServerArguments` for `-dpi 192` or similiar required

Cons:
- Uses private API :-/

### DIFF
--- a/src/common/Configuration.h
+++ b/src/common/Configuration.h
@@ -71,6 +71,7 @@ namespace SDDM {
             Entry(DisplayStopCommand,  QString,     _S(DATA_INSTALL_DIR "/scripts/Xstop"),      _S("Path to a script to execute when stopping the display server"));
             Entry(MinimumVT,           int,         MINIMUM_VT,                                 _S("The lowest virtual terminal number that will be used."));
             Entry(EnableHiDPI,         bool,        false,                                      _S("Enable Qt's automatic high-DPI scaling"));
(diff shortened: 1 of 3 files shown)


plfiorini:
I like what you are doing here but we should rather patch the `xcb` platform plugin to do that.

Vogtinator:
> I like what you are doing here but we should rather patch the xcb platform plugin to do that.

Yes, it would be ideal if Qt would try to detect lying monitors, but I doubt that that's going to happen, at least not with this many "safety nets" and most likely not enabled by default.

plfiorini:
What is preventing you from implementing the exact same thing inside the platform plugin?

Vogtinator:
That such changes would (if ever) only be available from Qt 5.12 on.

plfiorini:
Still better than having to use the private API.

Vogtinator:
It might be possible to avoid the private API by using the implementation-internal `_q_scaleFactor` property of `QScreens`. If you prefer that, I can try out that approach.

plfiorini:
No it's a half measure that might explode in our hands at some point and yet another excuse not to fix the issue upstream where it belongs.

Vogtinator:
I'll abandon this then.

PR #992, opened 2018-03-08
Fix platform detection for EnableHiDPI

### LINKED ISSUE #894: sddm login dialog is super small on my HiDPI laprop display after upgrading to 0.15
Everything worked fine for me with `sddm-0.14` when the login dialog was scaled according to my laptop display size parameters specified in the `/etc/X11/xorg.conf.d/90-monitor.conf`
After upgrading to `sddm-0.15` I always get super small dialog with tiny not readable font.
I thought maybe the cause is in the new `EnableHiDPI` parameter in `sddm.conf` file which is set to `false` by default and for some reason specified in two sections: `[Wayland]` and `[X11]`. But setting this parameter to `true` doesn't have any affect.

It might be that changes in some other package cause the problem, as I upgraded quite a few of them, but since this is sddm which is not functioning properly now, I think it's fair to report this here.

P.S. this might be related to another issue #692

Vogtinator:
We can't use QGuiApplication before it's constructed, so find out which
platform is requested ourselves.

Fixes #894

### DIFF
--- a/src/greeter/GreeterApp.cpp
+++ b/src/greeter/GreeterApp.cpp
@@ -278,11 +278,27 @@ int main(int argc, char **argv)
     // Install message handler
     qInstallMessageHandler(SDDM::GreeterMessageHandler);
 
+    // We set an attribute based on the platform we run on.
+    // We only know the platform after we constructed QGuiApplication
+    // though, so we need to find it out ourselves.
+    QLatin1String platform;
+    for (int i = 1; i < argc - 1; ++i) {
+        if(qstrcmp(argv[i], "-platform") == 0) {
+            platform = QLatin1String(argv[i + 1]);
+        }
+    }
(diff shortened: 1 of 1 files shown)


akontsevich:
So we may hope it will be fixed in 0.18?

Vogtinator:
If the issue was sddm being too small despite having `EnableHiDPI=true` enabled and the display reports the correct size, yes!

peterhoeg:
I've tried this patch applied to 0.17 and it doesn't seem to be enough to fix it - are other patches needed?

Vogtinator:
> I've tried this patch applied to 0.17 and it doesn't seem to be enough to fix it - are other patches needed?

If `ServerArguments= -dpi`... is needed, that's https://bugreports.qt.io/browse/QTBUG-67928

PR #996, opened 2018-03-20
Do not install sddm.conf by default

plfiorini:
sddm.conf overrides the configuration from the conf.d
directories and we don't want that.

SDDM still reads it if users keep the file after an
upgrade, but since the introduction of conf.d directories
we shouldn't install sddm.conf by default.

[ChangeLog][Daemon] Do not install /etc/sddm.conf anymore

### DIFF
--- a/src/daemon/CMakeLists.txt
+++ b/src/daemon/CMakeLists.txt
@@ -69,12 +69,3 @@ if(JOURNALD_FOUND)
 endif()
 
 install(TARGETS sddm DESTINATION "${CMAKE_INSTALL_BINDIR}")
-
-# Generate and install the default config file
-add_custom_command(COMMAND sddm --example-config > "${CMAKE_BINARY_DIR}/sddm.conf"
-    OUTPUT "${CMAKE_BINARY_DIR}/sddm.conf"
-    COMMENT "Generate default sddm.conf file"
-    DEPENDS sddm
-)
-add_custom_target(generate_config_file ALL DEPENDS "${CMAKE_BINARY_DIR}/sddm.conf")
-install(FILES "${CMAKE_BINARY_DIR}/sddm.conf" DESTINATION "${CMAKE_INSTALL_FULL_SYSCONFDIR}")


CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ add_custom_target(dist COMMAND ${CMAKE_MAKE_PROGRAM} package_source)
 
 # Options
 option(BUILD_MAN_PAGES "Build man pages" OFF)
+option(INSTALL_SDDM_CONF "Install sddm.conf" OFF)
```
Vogtinator: I'd add a slightly longer description - `"Install sddm.conf containing current default values for all configuration options"` or similiar.

Vogtinator:
I never understood the purpose of this in the first place - it overrides the internal defaults with the ones defined during installation (as it can either overwrite all options or keep all) which is not what you'd want.

I'd even suggest to remove this option altogether and repurpose ``--example-config`` as option to print "internal default values".

plfiorini:
It was needed before conf.d to have an easier way to change default values and them all in one file instead constantly reading the man page.

Keeping compatibility with the previous behavior with the `INSTALL_SDDM_CONF` option feels too much, in fact we should not even give that option. If someone wants install sddm.conf they can run `sddm --example-config > /etc/sddm.conf`. So in a future revision I will completely remove sddm.conf.

I also agree to repurpose `--example-config`, should I rename the argument?

Vogtinator:
> It was needed before conf.d to have an easier way to change default values and them all in one file instead constantly reading the man page.

Even then it overwrote the defaults, which means that updated internal defaults wouldn't have any effect anymore.

> Keeping compatibility with the previous behavior with the INSTALL_SDDM_CONF option feels too much, in fact we should not even give that option. If someone wants install sddm.conf they can run sddm --example-config > /etc/sddm.conf. So in a future revision I will completely remove sddm.conf.
> 
> I also agree to repurpose --example-config, should I rename the argument?

I'd say so - compatibility shouldn't be that important there.

plfiorini:
> > It was needed before conf.d to have an easier way to change default values and them all in one file instead constantly reading the man page.

> Even then it overwrote the defaults, which means that updated internal defaults wouldn't have any effect anymore.

sddm.conf was generated with defaults and every time the packages was updated a new configuration file was installed so depending on what the user did a she had a different outcome:

- overwrite old sddm.conf with the new one: loose custom settings
- keep old sddm.conf: loose internal defaults that may have changed with the new sddm version

Very bad.

plfiorini:
Updated

davidedmundson:
FWIW, +1 thanks

PR #997, opened 2018-03-20
Remove suffix for Wayland session

plfiorini:
Some desktops like GNOME specify which windowing system is in use 
with the Name entry of their desktop file.

For Wayland-only desktops such as Liri this information is
redundant and so is for X11-only window managers.

Do not append the Wayland suffix and let desktops handle it
themeselves.

[ChangeLog][Greeter] Remove suffix for Wayland sessions


Vogtinator:
I agree, but keeping backwards-compat in mind: How hard would it be to still apply the suffix if a conflicting session already exists?

plfiorini:
@Vogtinator What's a conflicting session?

Vogtinator:
> @Vogtinator What's a conflicting session?

Current Plasma releases have plasma.desktop and plasmawayland.desktop which both have `Name=Plasma`. It relies on this suffix for distinction.

See also https://phabricator.kde.org/D22210

Pointedstick:
This seems like the right approach to me, and then the session itself becomes responsible for having an accurate name.

What's blocking this?

davidedmundson:
>What's blocking this?

Me. We need to fix Plasma first. 

We should have done it for 5.17, that's arguably my fault, sorry. I'll make sure we do for Plasma 5.18.

a17r:
I think it is enough to make distro packagers aware through mail/changelog, no need for workarounds. We patch stuff now, we'll patch stuff in the meantime until it fits by default.

plfiorini:
Quick fix to deduplicate name, don't have time to even test this right now.
We'll do it later, but if you have time to test it please report what's wrong.
Basically it still appends " (Wayland)" if more than one session has the same name.

Vogtinator:
Basically what I requested a few comments above, yay. Code looks good, but I didn't test it yet either.

plfiorini:
Exactly that

Conan-Kudo:
@plfiorini @davidedmundson ping? Could this be rebased and merged? With Plasma 5.18 and 5.19, we now have this ugly `Plasma (Wayland) (Wayland)` descriptor now for the Wayland session.

Vogtinator:
@Conan-Kudo

That was already taken care of with https://github.com/sddm/sddm/pull/1202 though.

Conan-Kudo:
Welp... Thanks @Vogtinator!

plfiorini:
This patch is different: it adds the Wayland suffix only when there are more than one sessions with the same name.
If I recall the problem was that Plasma had the same name under both Xorg and Wayland.
Adding the Wayland suffix in SDDM is a hack, desktops that are available in multiple windowing systems should be named accordingly upstream.

Vogtinator:
> This patch is different: it adds the Wayland suffix only when there are more than one sessions with the same name.
If I recall the problem was that Plasma had the same name under both Xorg and Wayland.

With this patch, it won't be `GNOME (Wayland)` anymore, but for the `Plasma (Wayland) (Wayland)` case it won't actually make a difference at all.

plfiorini:
I was under the impression that Plasma had changed their desktop entry name removing Wayland.
I still believe that SDDM should stop resorting to this hack and desktops should get their name right.

Vogtinator:
> I was under the impression that Plasma had changed their desktop entry name removing Wayland.
> I still believe that SDDM should stop resorting to this hack and desktops should get their name right.

https://phabricator.kde.org/D22210 got merged, so `plasmawayland.desktop` has `Name=Plasma (Wayland)` since 5.18. The code here for deduplicating makes sure that even older `plasmawayland.desktop` appears separate. The only issue is with new Plasma and sddm without this PR or #1202, but that's nothing we can fix.

PR #1000, opened 2018-03-21
UserModel: lazily load user list

dtor:
In large environments loading user list can take some time and the list
is not usable anyway. The theme may opt not to show list at all, however
we create the model unconditionally. Let's lazily load data into the
model so that users disabling it do not have to pay the price.


dtor:
I see, OK, I'll create a new pull request if I ever get to implement the requested changes.

Issue #824, opened 2017-06-06
D-Bus SwitchToGreeter does not always work


lheckemann:
Sending the `org.freedesktop.DisplayManager.Seat.SwitchToGreeter` message will not always cause an actual switch, instead doing nothing. This seems to happen if the reply to the message can't be sent, i.e. if the sender of the original message drops off the bus (as I understand it).

This can be reproduced using `dbus-send`:
```
$ dbus-send --system --dest=org.freedesktop.DisplayManager /org/freedesktop/DisplayManager/Seat0 org.freedesktop.DisplayManager.Seat.SwitchToGreeter
(nothing happens)
$ dbus-send --print-reply --system --dest=org.freedesktop.DisplayManager /org/freedesktop/DisplayManager/Seat0 org.freedesktop.DisplayManager.Seat.SwitchToGreeter
(switch happens correctly)
```

i.e. the switch only happens if dbus-send is still there when SDDM tries to send the reply. This is not how it is meant to behave:

```
sphalerite> dbus-send --system --dest=org.freedesktop.DisplayManager /org/freedesktop/DisplayManager/Seat0 org.freedesktop.DisplayManager.Seat.SwitchToGreeter    does nothing unless I also add the option --print-reply. Why would this be?
sphalerite> Given that the side effect of the call is what I want, and I don't really care about the response, surely --print-reply shouldn't be necessary?
thiago> sphalerite: what do you mean by "does nothing"
thiago> are you saying that the called service does not take the action you asked of it?
sphalerite> thiago: doesn't switch to the greeter
sphalerite> yes
thiago> not a D-Bus bug then
thiago> please contact the people who made that service
sphalerite> I don't really understand what's going on there — why would the message's recipient be able to distinguish between the reply being wanted and the reply not being wanted at all?
thiago> there's a flag in the message header indicating whether the caller expects a reply
thiago> it's also possible that dbus-send has already exited by the time the called service receives and processes the call
thiago> depending on how the called service is written, it may get that information that the caller exited
sphalerite> ah right
thiago> for example, it may be asking D-Bus what the PID, UID, etc. are of the caller, which will fail
```

PR #1010, opened 2018-04-04
Fix switchToGreeter not available without logind

a17r:
Please note that I am only forwarding this patch from Alexander Miller:
```
Seats can't be created until SeatManager's signals are
connected to the DisplayManager, or the latter won't see
them and switchToGreeter doesn't work. So split SeatManager
initialization from its constructor and call initialize it
only after all connections have been set up in DaemonApp's
constructor.

With logind there may have been enough delay before seats
got actually added so things would work, but it's still
cleaner to fix the order.
```

Fixes: https://bugs.gentoo.org/644718
Fixes: https://github.com/sddm/sddm/issues/824

### DIFF
--- a/src/daemon/SeatManager.h
+++ b/src/daemon/SeatManager.h
@@ -31,8 +31,9 @@ namespace SDDM {
     class SeatManager : public QObject {
         Q_OBJECT
     public:
-        explicit SeatManager(QObject *parent = 0);
+        explicit SeatManager(QObject *parent = 0) : QObject(parent) {}
 
+        void initialize();
         void createSeat(const QString &name);
         void removeSeat(const QString &name);
         void switchToGreeter(const QString &seat);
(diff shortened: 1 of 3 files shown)


src/daemon/SeatManager.h
```diff
@@ -31,8 +31,9 @@ namespace SDDM {
     class SeatManager : public QObject {
         Q_OBJECT
     public:
-        explicit SeatManager(QObject *parent = 0);
+        explicit SeatManager(QObject *parent = 0) {}
```
Vogtinator: Missing call to parent constructor.

piotr-dobrogost:
> So split SeatManager initialization from its constructor

From [Arguments against “initialize()” method instead of constructors](https://stackoverflow.com/a/9767835/95735) :
> This is usually known as Two phase or Multiphase Initialization and it is particularly bad because once a constructor call has finished successfully, you should have a ready to use object, In this case you won't have a ready-to-use object.

PR #1017, opened 2018-04-21
Don't fill UserModel if theme does not require it

### LINKED ISSUE #479: Add an option to prevent greeter from building a full user model
When working on a machine on which user authentication is through a network server (e.g. NIS), the greeter screen takes a lot of time to start up (~10-15 min in my case), leaving the screen blank.

It seems to be caused by the greeter building a UserModel by repeated calls to `getpwent()`. On a system with many users and each call to `getpwent()` querying the network server, it makes the login process unresponsive.

Along with the option not to display the full user list but a login/pw text box (see #406 or #450) there should be a way to avoid building the full user list and check user name only after user input.

Vogtinator:
Certain themes switch to a username input field if there are too many users
to show. In those cases we don't need to provide a complete user model.
The last logged in user is still added (if available) to keep the index
and data valid.

To tell sddm that a theme does not need a full user model, theme.conf has to
set "General/needsFullUserModel=false".

Fixes #479


Vogtinator:
It's more elegant, but also more complex for not much gain IMO.

You have to load at least the avatar threshold of users to tell whether to enable or disable them anyway.
And once avatars are disabled, most themes don't display the user list anyway.

If you have many users, there shouldn't be a user model at all, there's no point as they don't fit on the screen.

PR #1047, opened 2018-07-09
Fixes #879 - Adds sourcing of /etc/profile to fish

### LINKED ISSUE #879: SDDM 0.15.0 doesnt source /etc/profile.d/* anymore
After the upgrade on Arch it seems that my files under `/etc/profile.d/..` don't get sourced anymore... various environment variabled don't work anymore. A downgrade to 0.14.0 helped...

jamtur01:
This ensures the right paths are set by upstream apps that use the
/etc/profile(.d) method for setting paths, for example xdg or snap.

Signed-off-by: James Turnbull <james@lovedthanlost.net>


plfiorini:
We already source `/etc/profile` for bash, I don't see why we shouldn't do the same for fish.
While this looks good to me, I'd like to have @davidedmundson opinion since he commented on #879

PR #1051, opened 2018-07-17
Remove correct view when a screen is removed

davidedmundson:
This connect was done for all views so previously when any screen was
removed all views would be.


plfiorini:
Can't test because I'm back with one screen but makes sense. view->screen() might not give useful results when the screen is removed but the captured screen does.

PR #1053, opened 2018-07-18
Handle session re-use in the same slot as everything else

davidedmundson:
m_auth is connected to ::authentication in the constructor so use of a
second connect was really weird and runs in an undetermined order. 
It also doesn't properly cleanup if auth is called again.


Vogtinator:
Great, I used the "web editor" to resolve the trivial conflict and now GitHub doesn't show any diff anymore.

+1, though now it's more obvious that it does more than it should in the session reuse case, like change the last used session...

Vogtinator:
Now I see why: This was directly commited a while ago: b02b00559fa9a7d3281b91af7faab74ed0a4ff73

So already merged and passed retroactive review -> closing.

PR #1054, opened 2018-07-19
Use modern connects

davidedmundson:


Vogtinator:
Can we make use of `QOverload`/`qOverload` instead?

davidedmundson:
>Can we make use of QOverload/qOverload instead?

Yes, but it would mean bumping our required Qt.

plfiorini:
qOverload is in Qt 5.7 AFAIK, we require 5.8 (https://github.com/sddm/sddm/commit/fddbf5278ead6d323c0068bba56fc7975e2b0954)

PR #1055, opened 2018-07-19
Keep string termination characters in utmp entries

davidedmundson:
strncpy copied N characters and does not resolve termination characters.
If entry.ut_line is 5 chars long, and our source is > 5 bytes we will
copy 5 leaving no null termination.

entry is all zeroed out on construction, so making sure we strncpy one
char less should be enough.

Not a security issue as it seems utmp does safe string copying out.


plfiorini:
This http://man7.org/linux/man-pages/man5/utmp.5.html says:

```
String fields are terminated by a null byte ('\0')
       if they are shorter than the size of the field.
```

PR #1057, opened 2018-07-22
Change location of sddm.conf.d to DATAROOTDIR

a17r:
/usr/lib is not a location that should be hardcoded - depending on
distribution and architecture this can be different. So far, SDDM
does not use this path unless the above conditions happen to make
it coincide with the install location of its Qml modules.

DATAROOTDIR is defined as: Read-only architecture-independent data.
This seems to be a better fit for a system SDDM default config dir.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -184,7 +184,7 @@ set(WAYLAND_SESSION_COMMAND     "${DATA_INSTALL_DIR}/scripts/wayland-session"   
 
 set(CONFIG_FILE                 "${CMAKE_INSTALL_FULL_SYSCONFDIR}/sddm.conf"        CACHE PATH      "Path of the sddm config file")
 set(CONFIG_DIR                  "${CMAKE_INSTALL_FULL_SYSCONFDIR}/sddm.conf.d"      CACHE PATH      "Path of the sddm config directory")
-set(SYSTEM_CONFIG_DIR           "${CMAKE_INSTALL_PREFIX}/lib/sddm/sddm.conf.d"      CACHE PATH      "Path of the system sddm config directory")
+set(SYSTEM_CONFIG_DIR           "${CMAKE_INSTALL_FULL_DATAROOTDIR}/sddm/sddm.conf.d" CACHE PATH      "Path of the system sddm config directory")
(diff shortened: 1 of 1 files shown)


Maryse47:
SDDM config dirs paths are similar to those used by systemd (`/lib` and/or `/usr/lib` and `/etc`). It's consistent and users should be already familiar it. Config dir shouldn't be defined as read-only.

a17r:
> Config dir shouldn't be defined as read-only.

`SYSTEM_CONFIG_DIR` is not where users are going to change stuff, that's what `/etc/sddm.conf` is for.

> It absolutely does.

Misunderstanding: 'does not install stuff into', obviously it uses `SYSTEM_CONFIG_DIR`

a17r:
Well obviously, the 'readonly' statement is not meant to distract distro maintainers and developers. And I don't see how systemd is in any way relevant for this decision. I was thinking of `/usr/share` because it is used for all the standard config templates of KDE software. But I don't see why it had to be moved away from `/etc` in the first place, where usually `conf.d` dirs reside (see X11, NM, fonts, polkit...).

Maryse47:
Theme files are indeed data which isn't supposed to be edited so it perfectly fits for `/usr/share`. I don't know what KDE config templates you are referring to. After brief look I didn't found anything looking like config file there.

Systemd may be relevant as it's known for storing config files with `<config>.d` hierarchy in `/{usr/lib`  like `sysusers.d, modprobe.d, modules-load.d, tmpfiles.d, sysctl.d, environment.d` plus everything under `systemd` dir.

Considering above there are a lot of configs stored in `/{usr/lib` on every user machine already. That's why I think it's natural choice for sddm for storing its own configs there.

> But I don't know why it had to be moved away from /etc in the first place

It didn't moved away form `/etc`. `/etc/sddm.conf` and `/etc/sddm.conf.d` are perfectly valid paths for storing sddm configs. In fact it's similar to config files hierarchy used by...systemd.

plfiorini:
Yes exactly @Maryse47, those path where chosen to be similar to systemd in order to make it easier to remember given that systemd is now in most distributions and administrators get used to such names. We didn't invent anything new on purpose.

PR #1073, opened 2018-09-10
Rewind password db before scanning it (again) for available users.

adriaandegroot:


davidedmundson:
Heh, you would have thought endpwent would have reset it.
Silly unix

PR #1093, opened 2018-10-24
Don't concatenate SessionLogFile to $HOME if absolute

zougloub:
This allows to set absolute paths as SessionLogFile ; they aren't appended to the user's home directory but just taken as is.


Vogtinator:
The log file would also have to be where all users have write access to, which opens it up to privilege boundaries - not fun.
So let's just close this.

PR #1098, opened 2018-11-07
Make theme.conf values have a data type comparison for INT and BOOL

### LINKED ISSUE #1097: Pull booleans from theme.conf into themes QML as such
Right now if I pull a variable from a theme.conf that is designed as a boolean it is passed on as string rather than boolean. (True for INT as well or just any value actually)
Let's say there's a variable:
```
[General]
MakeBooleansBooleans=false
```
When I want to depend on this boolean in a theme I usually write the following:
```
QtStringVariable: config.MakeBooleansBooleans == "false" ? falsyString : trueString
```
This is apparently superfluous when it could just be:
```
QtStringVariable: !config.MakeBooleansBooleans ? falsyString : trueString
```
This does not work. Directly probing any variable will always return as string or undefined/null. So the last example will always return true if the variable has anything set in the config, either BOOL, INT or STRING or if nothing was set it will return as undefined/null.

What does work is just echo the variable as a direct value:
```
someQtBooleanVariable: config.MakeBooleansBooleans
```

Is there a way to force this behavior in SDDM? Or are there plans on making this work? Would this be possible at all? Is there some `config.bool.var` option I am not aware of?

This would make a lot of theme coding way faster and a lot cleaner (after all this is a simplified example).

MarianArlt:
Fixes #1097

The INI format of `theme.conf` determines its contents as plain text which does not include data types. Key values are therefore **passed on by Qt as strings by default** if not told different.
A theme may include keys designed specifically as integer or boolean though and may want to compare these to their data type.
The added checks will convert the key values depending on whether or not they are of such design or else just pass them on as plain text resulting in a string.

The following example shows how values would be passed to QML theme files after merging this change.
```
# theme.conf
[General]
forceSomething=TRUE      // BOOL TRUE
forceSomething="false"   // BOOL FALSE
forceSomething=0         // INT & FALSE
raiseMargin=372          // INT & TRUE
raiseMargin="83"         // INT & TRUE
printText=Foo            // STRING & TRUE
printText="Foo"          // STRING & TRUE
printText=2Foo3          // STRING & TRUE
```
This is a **breaking change**. There are themes comparing keys to true as a string, as a workaround for the former implementation, which passed on any key value as plain text.
```
anchors.marginTop: config.forceSomething == "true" ? config.raiseMargin : 0
```
Themes impelementing such comparisons will behave different after merging this change.

### DIFF
--- a/src/common/ThemeConfig.cpp
+++ b/src/common/ThemeConfig.cpp
@@ -24,6 +24,8 @@
 #include <QDebug>
 #include <QSettings>
 #include <QStringList>
+#include <QRegularExpression>
+#include <QRegularExpressionMatch>
 
 namespace SDDM {
     ThemeConfig::ThemeConfig(const QString &path) {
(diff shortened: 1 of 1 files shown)


MarianArlt:
Travis CI fails with "Makefile not found.", just as with #1093 . Is this from my side or SDDMs?

Vogtinator:
Not sure whether this is a good idea. Only the theme itself knows what type it really wants. While treating everything as a string isn't that intuitive, it's at least very consistent.

I'll propose an alternative approach in the original issue.

PR #1126, opened 2019-01-18
Fbsd fix shutdown

adriaandegroot:
Bits and bobs related to shutdown on FreeBSD:
 - vt and halt-command should be different by default
 - vt switching is not available, simplify by splitting sources (yes, I realise that that means there's a file that the developers won't *normally* see; the alternative is big chunks of `#ifdef ` in the existing `VirtualTerminal.cpp` which I don't think is attractive either).
 - logging via `updwtmpx()` is not available.


CMakeLists.txt
```diff
@@ -167,8 +167,16 @@ add_feature_info("elogind" ELOGIND_FOUND "elogind support")
 
 # Default behaviour if neither systemd nor elogind is used
 if (NOT ELOGIND_FOUND AND NOT SYSTEMD_FOUND)
-    set(MINIMUM_VT 7)
-    set(HALT_COMMAND "/sbin/shutdown -h -P now")
+    # Set the VT on which sddm will normally appear, and the
+    # commands for shutdown and reboot. On FreeBSD, there are
+    # normally more getty's running than on Linux.
+    if(${CMAKE_SYSTEM} MATCHES "FreeBSD")
```
plfiorini: i think you should quote "${CMAKE_SYSTEM}"

src/daemon/CMakeLists.txt
```diff
@@ -29,9 +29,16 @@ set(DAEMON_SOURCES
     SeatManager.cpp
     SignalHandler.cpp
     SocketServer.cpp
-    VirtualTerminal.cpp
 )
 
+# Different implementations of the VT switching code
+# (where the FreeBSD version does nothing).
+if(${CMAKE_SYSTEM} MATCHES "FreeBSD")
```
plfiorini: ditto

CMakeLists.txt
plfiorini: Also use CMAKE_SYSTEM_NAME instead of CMAKE_SYSTEM, the former might include the version

plfiorini:
Manually merged

PR #1127, opened 2019-01-18
(FreeBSD) Link to libutil

adriaandegroot:
This is preparatory work: the sddm-helper is going to need
libutil for setusercontext(), which handles integration with
login.conf and per-user configurations (e.g. language settings,
CPU affinity and resource limits).


src/helper/CMakeLists.txt
```diff
@@ -32,6 +34,18 @@ endif()
 
 add_executable(sddm-helper ${HELPER_SOURCES})
 target_link_libraries(sddm-helper Qt5::Network Qt5::DBus Qt5::Qml)
+if(${CMAKE_SYSTEM} MATCHES "FreeBSD")
```
plfiorini: CMAKE_SYSTEM is ${CMAKE_SYSTEM_NAME}-${CMAKE_SYSTEM_VERSION}, or ${CMAKE_SYSTEM_NAME} if CMAKE_SYSTEM_VERSION is empty which means that this might not always work.
See https://cmake.org/cmake/help/v3.4/variable/CMAKE_SYSTEM.html

Use CMAKE_SYSTEM_NAME instead.

src/helper/CMakeLists.txt
```diff
@@ -32,6 +34,18 @@ endif()
...
+    # On FreeBSD (possibly other BSDs as well), we want to use
```
plfiorini: If this is true then you have to check also if CMAKE_SYSTEM_NAME equals to OpenBSD or NetBSD

PR #1146, opened 2019-03-11
Patch queue for 0.18.1 hot fix

plfiorini:


plfiorini:
Actually version will be 0.18.1 I guess

PR #1155, opened 2019-03-28
Enable QSurfaceFormat::ResetNotification on new Qt

davidedmundson:
This fixes graphical glitches on nvidia after VT switching.

It's opt-in as it requires code paths to handle glGetError differently.

The version comparison is because my early implementations missed a code
path. This was fixed in Qt5.12.2, but we may as well play safe.

### DIFF
--- a/src/greeter/GreeterApp.cpp
+++ b/src/greeter/GreeterApp.cpp
@@ -40,6 +40,9 @@
 #include <QDebug>
 #include <QTimer>
 #include <QTranslator>
+#include <QLibraryInfo>
+#include <QVersionNumber>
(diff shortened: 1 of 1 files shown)


davidedmundson:
For context, the Qt flag handles https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glGetGraphicsResetStatus.xhtml

and does internal scene graph magic

plfiorini:
Maybe add context and https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glGetGraphicsResetStatus.xhtml to the commit message?

Issue #906, opened 2017-10-11
SDDM v0.16 config file order messed


Vincent43:
According to release notes files should be read in following order:

- First from /etc/sddm.conf for compatibility
- All files in /usr/lib/sddm/sddm.conf.d in alphabetical order
- All files in /etc/sddm.conf.d in alphabetical order

`man 5 sddm.conf` states:

> Configuration  loads all files in the configuration directories followed by the configuration file in the order listed below with the latter having the most precedent.
> 
>        **
>               /usr/lib/sddm/sddm.conf.d*
> 
>        System Message: WARNING/2 (/build/sddm/src/build/data/man/sddm.conf.rst:, line 20)
>               Inline strong start-string without end-string.
> 
>             /etc/sddm.conf.d
>             /etc/sddm.conf
> 
And indeed `/etc/sddm.conf` takes precedence which makes other dirs unusable.
Tested on Archlinux

BTW: there is typo in man source. I sent PR with fix. https://github.com/sddm/sddm/pull/907

davidedmundson:
The man pages matches the code. The release notes are wrong.

Vincent43:
But this means `/etc/sddm.conf` which is installed by default will overwrite all changes set in `sddm.conf.d` dirs. It makes them useless. I'm sure the code should match release notes behavior as it's consistent with other apps using this concept like `systemd`.

davidedmundson:
But we don't install /etc/sddm.conf by default...

Vincent43:
Well, It's installed using this this template https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/sddm

Do you argue with my arguments about correct order?

davidedmundson:
Yes.

/etc/sddm.conf is legacy but a user might have their mods there.

A system could install some /usr/lib/sddm

We want to load the user's custom settings.

I shall fix the release notes.

If anyone is still installing a /etc/sddm.conf they should stop .

Vincent43:
Every software I know uses order from release notes, when app.conf.d takes precedence over app.conf.

All `sddm.conf.d` are empty by default so they don't change anything user set in his `/etc/sddm.conf`

[...]

PR #1160, opened 2019-04-21
Replace 'precedent' with 'precedence' in manpage

jairlopez:
I think this change improves documentation readability; it's based on the usage of the word _precedence_ in other programming languages like [Java](https://introcs.cs.princeton.edu/java/11precedence/)

Please also note how contributors to [#906](https://github.com/sddm/sddm/issues/906) overlooked the word _precedent_, and instead, used the word precedence instinctively.

_Precedence_ is an uncountable noun so it doesn't need the definite article.

### DIFF
--- a/data/man/sddm.conf.rst.in
+++ b/data/man/sddm.conf.rst.in
@@ -14,7 +14,7 @@ sddm display manager configuration
 SYNOPSIS
 ========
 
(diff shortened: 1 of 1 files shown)


plfiorini:
Thanks for the PR, looks good.

jairlopez:
My pleasure! It's a great project.

PR #1161, opened 2019-04-25
Added the ability to work under embedded plaform (eglfs and linuxfb)

ElenaZaretskaya:
I need the display manager for embedded system, so I have decided to add this ability in the sddm project.
Use BUILD_EMBEDDED option to build the embedded version of sddm.
Embedded sddm uses QtVirtualKeyboard for input (Qt 5.11 is required for external language switch). Also to provide input in the national language from the hardware keyboard I have added the ability to switch between languages to Qt (QTBUG-74542). This ability will be available in Qt-5.13.


plfiorini:
Could you please explain why and how this is supposed to work.
Why don't you just run the greeter with `eglfs` or `linuxfb`?
And QTBUG-74542 is about mingw builds.

plfiorini:
What are supposed to be the desktop files in embedded-sessions and what is supposed to be XDG_SESSION_TYPE=embedded.
And why do you need all that keyboard stuff, that should be handled by linuxfb I guess.

ElenaZaretskaya:
I want that sddm could works without X Window System under eglfs or linuxfb.
Greeter with eglfs or linuxfb? But sddm-daemon will create XorgDisplayServer,
but no XWindow on the target platform. Qt handles keyboard events really, but it contains
english keymap only by default. For national languages you need to load the appropriate keymap, therefore I have added keymap files. But even if you load the keymap (QEglFSFunctions::loadKeymap(const QString&)), there's no way to switch between english and another language than to press AltGr as a modifier. But in SDDM layouts are switched by combobox. Then I have added the function that allows to change the language (sorry, it's QTBUG-72452 https://bugreports.qt.io/browse/QTBUG-72452)

ElenaZaretskaya:
Make libxcb optional dependency in CMakeLists? But how build sddm? In XorgDisplayServer.cpp #include <xcb/xcb.h>, Display create object of XorgDisplayServer: m_displayServer(new XorgDisplayServer(this)). Explain, please. How I can run sddm on the platform where there is no libxcb.

plfiorini:
You need a new DisplayServer implementation in the daemon (inherit from DisplayServer) in order to run the greeter with `-platform eglfs`.

An embedded variant of the internal theme is overkill as it is shipping keymaps.

Of course both daemon and greeter should build without X libraries, to do that you can add a CMake feature (ENABLE_X11) and define HAVE_X11 and #ifdef the code

plfiorini:
I feel like you can head to the right path if you simplify this quiet a lot

plfiorini:
Closing this PR because it's inactive

Issue #1159, opened 2019-04-20
Incorrect list of groups membership


mbucas:
### Symptoms

On my 2 Gentoo machines, users have a problem of group membership. After loging in with SDDM in a KDE session, when using `id` or `groups` without arguments, the list of groups includes those from the `root` user, and groups common to the user and `root` are duplicated, like `wheel`, `video`, `allowssh`. I observed that the list of groups is incorrect only in a KDE session, either in Konsole or xterm, but the list of groups is correct in other kinds of sessions like on a TTY or through SSH.

```
mick@xxx ~ $ groups
root bin daemon sys adm disk wheel wheel floppy uucp cron audio cdrom
dialout tape video video games cdrw apache usb vboxusers portage
allowssh allowssh svn users mick
mick@xxx ~ $ id
uid=1001(mick) gid=1001(mick)
groupes=1001(mick),0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),14(uucp),16(cron),18(audio),19(cdrom),20(dialout),26(tape),27(video),35(games),80(cdrw),81(apache),85(usb),102(vboxusers),250(portage),800(allowssh),909(svn),1000(users)
```

When run with a login, the list is correct

```
mick@xxx ~ $ groups mick
wheel cron audio cdrom video games cdrw apache usb vboxusers portage
allowssh svn users mick
mick@xxx ~ $ groups root
root bin daemon sys adm disk wheel floppy uucp dialout tape video allowssh
mick@xxx ~ $ id mick
uid=1001(mick) gid=1001(mick)
groupes=1001(mick),10(wheel),16(cron),18(audio),19(cdrom),27(video),35(games),80(cdrw),81(apache),85(usb),102(vboxusers),250(portage),800(allowssh),909(svn),1000(users)
mick@xxx ~ $ id root
uid=0(root) gid=0(root)
groupes=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),14(uucp),20(dialout),26(tape),27(video),800(allowssh)
```

The difference in output between `id` and `id mick` happens because `id` calls the syscall `getgroups` in the first case, but not in the other, as I could see with `strace`.

The list is also incorrect for `sddm-greeter` : it's running under user `sddm`, but the groups are those of `root` (before a KDE session is opened).

```
mick@xxx ~ $ grep Groups /proc/$(pidof sddm-greeter)/status
[...]

Issue #416, opened 2015-05-08
groups and id command don't work properly when login with SDDM


Pulfer:
"groups" and "id" command don't work properly when login with SDDM. They display only main user group (=username). But if I run "groups username" or "su username" (then "groups"), all groups are properly listed.

In case you need to check how SDDM is built, see:
https://abf.rosalinux.ru/import/sddm/tree/rosa2014.1

tkleindienst:
The root of this problem is the same as with my issue #414 
It seems sddm can not deal with some pam modules, esp.  pam_group
After looking for hours through /etc/pam.d/ I finally gave up. I have not found a solution yet.

tkleindienst:
Further remark: If you do "su username" pam does not utilize /etc/pam.d/sddm
Most likely /etc/pam.d/login will be used. That's why the group assignment works.

jleclanche:
@plfiorini @davidedmundson thoughts?

Pulfer:
Sorry, this bug is invalid. The issue was in kinit from KDE Frameworks 5, not in SDDM.

hslabbert:
@Pulfer please elaborate how an issue in kinit, which relates to kerberos, results in logins through sddm failing to honor pam_group.so / /etc/security/group.conf on a system configured through sssd for ldap authentication with no kerberos being used, and what config if any would allow sddm to honour pam_group.so / /etc/security/group.conf in such a setup.

Pulfer:
@hslabbert It was kinit from KDE Frameworks 5 ( https://quickgit.kde.org/?p=kinit.git ), not from kerberos. Setting start_kdeinit permissions to 4755 as defined in CMakeLists.txt ( https://quickgit.kde.org/?p=kinit.git&a=blob&h=46d6cb35454727e40f0d1fa066209045511b970e&hb=ae0e855c6b6b87d000988cc71e054f734109ec06&f=src%2Fstart_kdeinit%2FCMakeLists.txt#l21 ) leads to wrong group and id results.

hslabbert:
@Pulfer appreciate the clarification; thanks. I was stuck in kinit = MIT "kerberos init" with this being related to auth.  Yay for namespace collisions...
Apologies for the noise.

PR #1165, opened 2019-05-09
fix supplemental group initialization (pam_group.so)

jktr:
This MR reverts and attempts to re-implement #834.
It also serves as a fix for #1159, #414, and #416.

The original implementation in `1bc813d0` turned out to be broken and leads to privilege escalation in some setups, as it pulls in all supplemental groups at initialization (PAM+user+root) instead of just the user's groups (user+PAM).

This version moves the supplemental group initialization step from `UserSession` to `Backend`, and makes it accessible to `UserSession` via `HelperApp` so that `PamBackend` is able to inject its additional supplemental groups for modules like `pam_group.so` via `pam_setcred`.

As presented in #1159, the correct PAM initialization sequence seems to be as follows, and at least LightDM seems to do it this way [3]:

* `pam_start` (as root w/ root's sup groups)
* `pam_authenticate`
* `initgroups` (now root w/ user's sup groups)
* `pam_setcred` (now root w/ user's sup groups + other credentials)
* `pam_open_session`
* `setuid`/`setgid` (now user w/ user's sup groups + other user credentials)
* exec the user's session

SDDM's existing initialization model doesn't really seem compatible with this:
`PamBackend` calls `pam_setcred(PAM_ESTABLISH_CRED)` shortly before `pam_open_session`at the beginning of `PamBackend::openSession`, but the `initgroups`/`setgid`/`setuid` step happens after that in `UserSession::setupChildProcess` via `Backend::openSession` via `PamBackend::openSession`.

I couldn't find a good way to reorder the supplemental group initialization before `pam_setcred(PAM_ESTABLISH_CRED)` that doesn't require a bunch of `#ifdef`s or major refactoring. As a result, this MR instead calls `pam_setcred(PAM_REINITIALIZE_CRED)` after the supplemental group initialization in the new `Backend::setupSupplementalGroups` to force PAM to redo the initialization step and add its groups to the user's.

The re-initialization approach seems to work, at least during my testing on an archlinux box with standard PAM config and an opensuse system using NFS homes and `pam_sssd.so`. I also took a look at `pam_sm_setcred` in  all upstream PAM modules [0] and pam_sssd [2], and couldn't find any obvious case in which calling `PAM_REINITIALIZE_CRED` would break things if `PAM_ESTABLISH_CRED` worked. However, this could possibly cause side-effects or breakage for more esoteric pam modules.

I'm not too familiar with PAM or C++, so I'd appreciate your comments. Some suggestions on how to rework the control flow to avoid `PAM_REINITIALIZE_CRED` would be nice, too.

Please note that the original patch proposed by @mbucas in #1159 has been implemented here [1].

[0] https://github.com/linux-pam/linux-pam/search?l=C&q=pam_sm_setcred
[1] https://github.com/jktr/sddm/tree/fix/supplemental-groups-v1
[2] https://github.com/SSSD/sssd/blob/45efba71befd96c8e9fe0a51fc300cafa93bd703/src/sss_client/pam_sss.c#L2736
[3] https://github.com/CanonicalLtd/lightdm/blob/ff1c38c0616b3df66a5002ed7f531d53f218dfab/src/session-child.c#L512-L529

### DIFF
--- a/src/helper/Backend.cpp
+++ b/src/helper/Backend.cpp
@@ -29,6 +29,7 @@
 #include <QtCore/QProcessEnvironment>
 
 #include <pwd.h>
+#include <grp.h>
 
 namespace SDDM {
     Backend::Backend(HelperApp* parent)
@@ -78,5 +79,9 @@ namespace SDDM {
 
     bool Backend::closeSession() {
         return true;
+    }
+
+    bool Backend::setupSupplementalGroups(struct passwd *pw) {
+        return !initgroups(pw->pw_name, pw->pw_gid);
     }
 }
--- a/src/helper/Backend.h
+++ b/src/helper/Backend.h
@@ -22,6 +22,7 @@
 #define BACKEND_H
 
 #include <QtCore/QObject>
+#include <pwd.h>
 
 namespace SDDM {
     class HelperApp;
@@ -37,6 +38,8 @@ namespace SDDM {
 
         void setAutologin(bool on = true);
         void setGreeter(bool on = true);
+
+        virtual bool setupSupplementalGroups(struct passwd *pw);
 
     public slots:
         virtual bool start(const QString &user = QString()) = 0;
--- a/src/helper/HelperApp.cpp
+++ b/src/helper/HelperApp.cpp
@@ -253,6 +253,10 @@ namespace SDDM {
         return m_session;
     }
 
+    Backend *HelperApp::backend() {
+        return m_backend;
+    }
+
     const QString& HelperApp::user() const {
         return m_user;
     }
--- a/src/helper/HelperApp.h
+++ b/src/helper/HelperApp.h
@@ -39,6 +39,7 @@ namespace SDDM {
         virtual ~HelperApp();
 
         UserSession *session();
+        Backend *backend();
         const QString &user() const;
         const QString &cookie() const;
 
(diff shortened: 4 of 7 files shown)


Vogtinator:
What's the reason you split the call to `initgroups` into the Backend instead of calling it directly like the PAM functions were in the previous patch?

Vogtinator:
Can you elaborate on

> The original implementation in 1bc813d0 turned out to be broken and leads to privilege escalation in some setups, as it pulls in all supplemental groups at initialization (PAM+user+root) instead of just the user's groups (user+PAM).

Which setups are effected? I just checked here in a VM and my test user does not have any of root's supplemental groups active.

Ri0n:
Is this PR dead?
Gentoo uses this patchset. But it also has conflicts against master like this PR. So Gentoo is stuck with sddm-0.18.1 which doesn't work well with recent X (X forgets magic cookies).
Frankly I don't know how stable the master branch, but at least w/o these patches I don't have problems with magic cookies.

Ri0n:
>  but at least w/o these patches I don't have problems with magic cookies.

I'm wrong about this. gotta report it..

Vogtinator:
No response from the author, so let's close this

a17r:
@jktr still there? it seems this work is still necessary, unfortunately.

PR #1182, opened 2019-07-23
Only allocate VTs for seat0 [new branch]

lbssousa:
Currently only systemd-logind special seat named "seat0" can
handle VTs. Allocating a VT and setting XDG_VTNR for non-seat0
seats is currently pointless. Moreover, Xorg server 1.16 and
newer doesn't touch VTs if -seat option is passed with a
non-seat0 value, so there's no need to pass vt?? argument to
a non-seat0 Xorg server.

This fixes some leftovers from multi-seat support implemented by @davidedmundson.

Issue: #78


plfiorini:
LGTM but I'd like @davidedmundson to take a look

PR #1184, opened 2019-07-29
sddm.service: Limit restarts

Vogtinator:
Currently it's possible to have sddm in a crash loop. The systemd default
is to block restarts if there are more than 5 during 10s, but sddm is too
slow to start to trigger that.
Limit it to a max of two restarts during 30s to work reliably even on slow
systems with complex themes.

Reported on https://bugzilla.opensuse.org/show_bug.cgi?id=1142492


Vogtinator:
I'm not quite done testing, so please don't merge just yet...

plfiorini:
In that case you could use the WIP prefix

Vogtinator:
Test successful, replaced `sddm` with `false` in the service file and it failed quickly.

PR #1191, opened 2019-08-18
Add font option

flipwise:
Users may wish to further their SDDM theming efforts by using a custom font.

In order to facilitate this, this patch adds a config option which allows users to specify a font.

DEs can also use the option by adding code which would sync the user's font with SDDM.


flipwise:
> The font is not guaranteed to be used: the theme can override it
> Please at least say so in the man page

I added a line in the documentation, thanks!

PR #1197, opened 2019-09-09
Update danish translation

scootergrisen:
Based on ru.ts
Hope it is correct.

### DIFF
--- a/data/translations/da.ts
+++ b/data/translations/da.ts
@@ -18,6 +18,10 @@
 <context>
     <name>TextConstants</name>
     <message>
+        <source>Welcome to %1</source>
+        <translation>Velkommen til %1</translation>
+    </message>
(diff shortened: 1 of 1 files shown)


data/translations/da.ts
```diff
@@ -45,10 +49,46 @@
...
+    <message>
+        <source>Suspend</source>
+        <translation>Hvile</translation>
+    </message>
+    <message>
+        <source>Hibernate</source>
+        <translation>Dvale</translation>
+    </message>
+</context>
+</TS>
```
plfiorini: why does it end <TS> here?

scootergrisen: I based it on ru.ts.
Did not generate the file my self since this is difficult.
How else should it end?

PR #1202, opened 2019-09-30
Prevent duplicate session name.

Symbitic:
Several desktop sessions (e.g. KDE Plasma) already include the string " (Wayland)" in the session name. When this happens, the session name displayed to the user is "Plasma (Wayland) (Wayland)". This change makes it so that only "Plasma (Wayland)" will be displayed.

### DIFF
--- a/src/common/Session.cpp
+++ b/src/common/Session.cpp
@@ -169,7 +169,10 @@ namespace SDDM {
 
             if (line.startsWith(QLatin1String("Name="))) {
                 if (type == WaylandSession)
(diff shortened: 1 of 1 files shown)


Vogtinator:
> Several desktop sessions (e.g. KDE Plasma) already include the string " (Wayland)" in the session name.

They actually don't, at least not upstream and not now.

plfiorini:
Yeah it's badly phrased but I figured it doesn't hurt anyway

PR #1208, opened 2019-10-09
Update danish translation

scootergrisen:


scootergrisen:
Is it fine to make pull request only for develop or does it need to be done for master also?

plfiorini:
Yes it's fine you only need to send PRs for `develop` because it is merged into `master` at release time

PR #1216, opened 2019-12-12
Set re-use session option by default

davidedmundson:
I added Reuse session option a few years ago as often logging into a
graphical session twice is not what the user actually wants.

This is particularly important for Plasma which doesn't really support
multiple logins. This has got worse recently as systemd's defaults share
a DBus session across user sessions which makes two graphical logins
even more likely to explode.

Some distros have this set by default.

The option still remains for user, but it'll reduce my bug reports if
all distros have it on by default.


Pointedstick:
+1, this is a saner default IMO.

apachelogger:
+1

PR #1223, opened 2020-01-02
Cleanup auth

davidedmundson:


plfiorini:
lgtm, i was busy with the holidays and forgot to reply to your email
just one comment

src/daemon/Display.cpp
```diff
@@ -264,6 +264,11 @@ namespace SDDM {
     }
 
     void Display::startAuth(const QString &user, const QString &password, const Session &session) {
+
+        if (m_auth->isActive()) {
+            return;
```
plfiorini: maybe we should put a warning before return to tell auth was previously done?

plfiorini:
ops i just realized that this got merged in `master` rather than `develop`

davidedmundson:
Yes, I wanted it in stable?

plfiorini:
`develop` is branched for new releases, master is not. Unless it's a hotfix for the previous release and in that case we should cherry pick from `develop`

PR #1246, opened 2020-04-05
Use avatars in FacesDir first and if not found search other locations

gcala:
If Avatars are enabled and home is encrypted it takes 60s to popup the greeter at boot. The same is if copying $USER.face.icon in FacesDir since avatar is searched first in home.

The patch changes this logic. Now it search first in FacesDir and if found the greeter is shown, otherwise queries others default locations.


mathieujobin (commented):
this makes sense... I hit that bug where I was changing my profile picture but it was not taken into account. I have to modify the `.face.icon` for it to appear.
now I get it, it make sense, the system faces folder is taken last.

but it usally makes more sense to have the `userFace` here take precedence...

src/greeter/UserModel.cpp
```diff
@@ -134,16 +134,20 @@ namespace SDDM {
...
+                if (QFile::exists(systemFace))  {
                     user->icon = QStringLiteral("file://%1").arg(systemFace);
+                } else {
+                    const QString userFace = QStringLiteral("%1/.face.icon").arg(user->homeDir);
+                    QString accountsServiceFace = QStringLiteral("/var/lib/AccountsService/icons/%1").arg(user->name);
+
+                    if (QFile::exists(userFace)) {
+                        user->icon = QStringLiteral("file://%1").arg(userFace);
+                    } else if (QFile::exists(accountsServiceFace)) {
+                        user->icon = accountsServiceFace;
```
mathieujobin: the diff could be more minimal, only bringing up two lines, instead of changing the structure to have two different imbricated `if`

gcala: I pushed new code following your advice

mathieujobin (approved):

plfiorini:
Manually merged adding a comment and remove unnecessary curly braces

PR #1249, opened 2020-04-21
Cleanup sessions on exit

davidedmundson:
On exit we currently don't do any cleanup. This leaves stray processes
running. Some aren't even the responsibility of the session, such as
anything spawned in the Xsessions.d directory.

In the event of a hard reset (systemctl --restart sddm) we currently
don't even send a SIGTERM to the session leader.

By registering as a subreaper we can have the kernel automatically clean
up anything when sddm-helper is quit gracefully or unexpectedly.

### DIFF
--- a/src/helper/HelperApp.cpp
+++ b/src/helper/HelperApp.cpp
@@ -37,6 +37,7 @@
 
 #if defined(Q_OS_LINUX)
 #include <utmp.h>
+#include <sys/prctl.h>
 #endif
 #include <utmpx.h>
 #include <QByteArray>
@@ -50,6 +51,9 @@ namespace SDDM {
         qInstallMessageHandler(HelperMessageHandler);
 
         QTimer::singleShot(0, this, SLOT(setUp()));
+#if defined(Q_OS_LINUX)
+        prctl(PR_SET_CHILD_SUBREAPER);
+#endif
     }
 
     void HelperApp::setUp() {


plfiorini:
I assume that nobody is using Linux < 3.4 at this point.

src/helper/HelperApp.cpp
```diff
@@ -50,6 +51,9 @@ namespace SDDM {
         qInstallMessageHandler(HelperMessageHandler);
 
         QTimer::singleShot(0, this, SLOT(setUp()));
+#if defined(Q_OS_LINUX)
+        prctl(PR_SET_CHILD_SUBREAPER);
```
Vogtinator, line 55: According to the manpage:

```
       PR_SET_CHILD_SUBREAPER (since Linux 3.4)
              If arg2 is nonzero, set the "child subreaper" attribute of the calling process; if arg2 is zero, unset the attribute.
```

So this is an invalid call as the argument is missing.

davidedmundson:
>Does sddm currently handle SIGCHLD? If not, this would cause zombie processes to appear.

It does not explicitly. My understanding from the docs is that was only if we wanted to block until the children finished exiting. It might have temporary zombies, but they do all get a TERM signal, and in my testing disappear.

>So this is an invalid call as the argument is missing.

You are right :/ 
Which probably means it reads garbage which is almost always non-zero, hence appearing to "work". Will fix.

Vogtinator:
> It does not explicitly. My understanding from the docs is that was only if we wanted to block until the children finished exiting. It might have temporary zombies, but they do all get a TERM signal, and in my testing disappear.

What I'm mostly wondering about is how this behaves together with the weird `QProcess` way of dealing with children and also the glib event loop, which also does some weird stuff with signals.

Unfortunately this area is very badly documented...

davidedmundson:
>What I'm mostly wondering about is how this behaves together with the weird QProcess way of dealing with children and also the glib event loop, which also does some weird stuff with signals.

They do.
It means I can't tell you from an academic POV. From a practical POV I've tried and tested:

logging in (as the greeter is a session too)
logging out on X and wayland
having the session process crash
hard restarts of SDDM

and everything behaved as you would expect

Vogtinator:
Currently, anything orphaned gets parented straight to init, so init can decide what to do, either keep it or explicitly kill it. For systemd, this is explicitly configured as `KillUserProcesses`. Some distros set that to false after there was a huge backlash when systemd enabled that by default and stuff like `screen` just stopped working.

So currently I don't understand this PR for two reasons:
- Does this actually make a difference? When the session process dies (causing `sddm-helper` to exit), every left over process gets reparented to init anyway.
- If this would actually kill all (adopted) children of `sddm-helper` on session exit, we'd have the same effect as `KillUserProcesses=true`, but without any configuration switch.

Am I missing something?

PR #1267, opened 2020-06-22
Implement substitution of passwd fields in session log file path

lameventanas:
This allows the path in sessionLog to have substitution variables from the user's passwd entry: user, uid, gid, gecos and home.

This also allows setting the SessionLogFile to a path outside the user's home, but the defaults will remain unchanged:
`%{HOME}%/.local/share/sddm/xorg-session.log`
`%{HOME}%/.local/share/sddm/wayland-session.log`

### DIFF
--- a/data/man/sddm.conf.rst.in
+++ b/data/man/sddm.conf.rst.in
@@ -126,7 +126,7 @@ OPTIONS
 
 `SessionLogFile=`
         Path to the user session log file, relative to the home directory.
-        Default value is ".local/share/sddm/xorg-session.log".
+        Default value is "%{HOME}%/.local/share/sddm/xorg-session.log".
 
 `UserAuthFile=`
         Path to the Xauthority file, relative to the home directory.
--- a/src/common/Configuration.h
+++ b/src/common/Configuration.h
@@ -66,8 +66,8 @@ namespace SDDM {
             Entry(XauthPath,           QString,     _S("/usr/bin/xauth"),                       _S("Path to xauth binary"));
             Entry(SessionDir,          QString,     _S("/usr/share/xsessions"),                 _S("Directory containing available X sessions"));
             Entry(SessionCommand,      QString,     _S(SESSION_COMMAND),                        _S("Path to a script to execute when starting the desktop session"));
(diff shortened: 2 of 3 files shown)


plfiorini:
I'm not sure we want this. Any thoughts?
It's probably enough to allow tilde in the path and resolve it to the home directory.
Besides, I'm not sure we are using the right default path for logs.

lameventanas:
This would allow the logs to be anywhere.
Currently it can only be in the home directory.

plfiorini:
@lameventanas The point is that logs are written by your unprivileged user, that usually can only access his home directory. Writing in another path, would likely result in an error that prevents you from logging in.

lameventanas:
The default location remains the same.
This just allows logs to be outside the home directory if needed, which right now can't be done except by using symbolic links.
Whomever sets this up would have to arrange the permissions issue.

davidedmundson:
>I'm not sure we want this. Any thoughts?

Do I want to use it, no for the reasons you said. At a desktop level I'm trying to put more things into journald anyway.

Do I want this patch in? I certainly have no objections. The patch is right that you can't put in /var or /tmp without doing some per-invocation substitutions, so I'd rather have that than have people abuse the current system and lose logs.

Vogtinator (changes requested):
In its current state, this PR breaks existing configuration of the `SessionLogFile` value, as it now has a different meaning.

AFAICT, having this outside of `$HOME` is broken. The target file needs to be writable by the user, which means that either:

a) It already has to exist (not possible to guarantee)
b) It's in a world-writable directory (insecure, allows symlink attacks)

If this is actually desired, what I suggest instead is to just create a symlink to the (existing!) target file inside `~/.local/share/sddm/`, which makes sure that a) is the case as it only applies to users where this was explicitly changed.

plfiorini (changes requested):
I'm against this patch because having log files outside of $HOME is dangerous and this PR breaks existing configurations.

lameventanas:
> I'm against this patch because having log files outside of $HOME is dangerous and this PR breaks existing configurations.

It's not dangerous unless someone sets it up to write in world-writable directories, but it is possible to have user-writable only directories outside $HOME.

PR #1268, opened 2020-06-26
Move creation of cookie file before running X server

atupone:
I have no clue about what is written in the cookie.
I know that I was getting something like
(EE) failed to open authorization file "/var/run/sddm/{xxxxx}": no such file or directory
in /var/log/Xorg.0.log
It sounds to me that now the xserver run so fast that sddm has not created the cookie file in time
This pull request move the create of cookie before running the server


plfiorini:
Closing this PR, as @Vogtinator said this is not the right approach but nevertheless thanks for submitting this PR.

PR #1273, opened 2020-07-05
Move VT setup to sddm-helper

### LINKED ISSUE #1078: SDDM leaks FDs of ttys of wayland sessions
I found that I could not start a new session today as no VT could be opened.
The reason is that sddm had open file descriptors to `/dev/tty8` up to `/dev/tty63`.

I assume those were the FDs used for previous wayland sessions.

### LINKED ISSUE #1083: Race condition in VT setup
The user session is started after jumping to the selected VT.
It is possible that in between another process (like getty) takes ownership of the VT and the user session's startup fails.

Due to the FD leak this did not happen before, so this was uncovered by https://github.com/sddm/sddm/pull/1079.

This should be prevented explicitly by taking control of the tty until the session ends.

Vogtinator:
Previously, sddm set up the virtual terminal and jumped to it and after that,
sddm-helper opens it again and takes control of it. This only worked because
sddm leaked the terminal fd into sddm-helper.

The right way is to set up the virtual terminal and take control of it before
jumping to it. This is achieved by moving this responsibility to sddm-helper.

As the race is now fixed, also avoid leaking FDs in jumpToVt now.

Fixes #1078
Fixes #1083


src/common/VirtualTerminal.cpp
```diff
@@ -187,6 +187,8 @@ namespace SDDM {
                 qWarning("Couldn't finalize jump to VT %d: %s", vt, strerror(errno));
 
             close(activeVtFd);
+            if (fd != activeVtFd)
```
codewiz, line 190: This might also be:
  if (vtFd != -1) close(vtFd);

It seems slightly more obvious this way, but no strong opinions.

Would be good to hold these into a unique_fd class like this one:
  https://cs.android.com/android/platform/superproject/+/master:system/libbase/include/android-base/unique_fd.h

I don't know whether Qt or KDE Frameworks has anything similar.

codewiz:
I tested this patch manually.

It fixes the issue with the leaked ttys, but "Switch Session" does not seem to work (clicking on "Start New Session" brings you to the previous session). Was this a pre-existing problem with Wayland?

src/common/VirtualTerminal.cpp
Vogtinator: `QFile` would be ideal, but it doesn't support passing `O_NOCTTY`...

Vogtinator:
> I tested this patch manually.
> 
> It fixes the issue with the leaked ttys, but "Switch Session" does not seem to work (clicking on "Start New Session" brings you to the previous session). Was this a pre-existing problem with Wayland?

Does `loginctl activate` work? I'll have a look later/tomorrow.

Vogtinator:
> > I tested this patch manually.
> > It fixes the issue with the leaked ttys, but "Switch Session" does not seem to work (clicking on "Start New Session" brings you to the previous session). Was this a pre-existing problem with Wayland?
> 
> Does `loginctl activate` work? I'll have a look later/tomorrow.

Was a bug in Plasma, which I fixed: https://invent.kde.org/plasma/plasma-workspace/-/merge_requests/175
"Switch User" works fine here now in a Wayland session, both for switching to an X11 or Wayland session as well as creating a new one.

What appears broken though is `ReuseSession=true`, which needs a closer look.

Vogtinator:
> What appears broken though is ReuseSession=true, which needs a closer look.

That seems to be a multitude of issues, probably race conditions, but unrelated to this PR.
I have one of those fixed, but it still fails seemingly randomly, so I'm not making a PR just yet.

src/common/VirtualTerminal.cpp
Vogtinator: Changed

davidedmundson (approved):

plfiorini (approved):

PR #1280, opened 2020-07-29
Don't abort on failure to start the display server

Vogtinator:
Currently sddm logs a fatal error if X fails to start, which triggers an
abort and thus a very unclean shutdown and even a core dump.
In the case of a broken X setup, this leads to frequent service restarts
triggered by systemd which don't help and just result in core dumps and
if this fails in the handling of new seats or switchToGreeter, it'll even
kill all other sessions, potentially leading to data loss.

Logging it as critical error is enough, at least it doesn't make it worse.

### DIFF
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -102,15 +102,8 @@ namespace SDDM {
         return m_seat;
     }
 
-    void Display::start() {
-        // check flag
-        if (m_started)
-            return;
-
-        // start display server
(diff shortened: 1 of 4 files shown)


davidedmundson:
LGTM

Vogtinator:
While doing some tests for 1273 I broke X startup (unintentionally...) which allowed me to test this. It works as expected, sddm keeps running, the error is seen in the journal and retrying is possible.

plfiorini (approved):

PR #1283, opened 2020-08-05
Fix container anti pattern uses

luis-pereira:
Temporary containers are being created needlessly. Just iterate the
containers directly instead.
Also made the containers const to avoid possible C++11 range-for container
detachments.

### DIFF
--- a/src/common/ConfigReader.cpp
+++ b/src/common/ConfigReader.cpp
@@ -241,16 +241,18 @@ namespace SDDM {
         if (section) {
             if (entry && !entry->matchesDefault())
                 remainingEntries.insert(section, entry);
-            else
-                for (const ConfigEntryBase *b : section->entries().values())
+            else {
+                for (const ConfigEntryBase *b : qAsConst(section->entries()))
                     if (!b->matchesDefault())
                         remainingEntries.insert(section, b);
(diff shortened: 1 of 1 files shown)


src/common/ConfigReader.cpp
```diff
@@ -241,16 +241,20 @@ namespace SDDM {
         if (section) {
             if (entry && !entry->matchesDefault())
                 remainingEntries.insert(section, entry);
-            else
-                for (const ConfigEntryBase *b : section->entries().values())
+            else {
+                const auto entries = section->entries();
```
Vogtinator, line 245: Why a separate variable here?

luis-pereira: To be systematic.
Since ```section->entries()``` returns `const QMap` there's no need. But by looking at the code, that's not obvious. 
Just to make it more easier to understand, since there's no performance penalty.

Vogtinator: It's not highlighting that it's returning a `QMap`, so to indicate the `const`, you could just explicitly use `qAsConst` like below?

luis-pereira: I'm Ok with that.

PR #1286, opened 2020-08-12
Perform initialization on the initialization list

luis-pereira:
When an object of a class is created, the constructors of all member
variables are called consecutively in the order the variables are declared,
even if you don't explicitly write them to the initialization list.

Avoids double "initialization".

### DIFF
--- a/src/daemon/DisplayManager.cpp
+++ b/src/daemon/DisplayManager.cpp
@@ -129,11 +129,8 @@ namespace SDDM {
         }
     }
 
-    DisplayManagerSeat::DisplayManagerSeat(const QString &name, QObject *parent) : QObject(parent) {
-        // set name and path
-        m_name = name;
-        m_path = DISPLAYMANAGER_SEAT_PATH + name.mid(4);
-
+    DisplayManagerSeat::DisplayManagerSeat(const QString &name, QObject *parent)
+        : QObject(parent), m_name(name), m_path(DISPLAYMANAGER_SEAT_PATH + name.mid(4)) {
         // create adaptor
         new SeatAdaptor(this);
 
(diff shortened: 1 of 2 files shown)


src/daemon/DisplayManager.cpp
```diff
@@ -129,11 +129,7 @@ namespace SDDM {
         }
     }
 
-    DisplayManagerSeat::DisplayManagerSeat(const QString &name, QObject *parent) : QObject(parent) {
-        // set name and path
-        m_name = name;
-        m_path = DISPLAYMANAGER_SEAT_PATH + name.mid(4);
-
+    DisplayManagerSeat::DisplayManagerSeat(const QString &name, QObject *parent) : QObject(parent), m_name(name), m_path(DISPLAYMANAGER_SEAT_PATH + name.mid(4)) {
```
Vogtinator, line 132: Newline before the `:` would make the line shorter

src/daemon/DisplayManager.cpp
```diff
@@ -171,10 +167,7 @@ namespace SDDM {
        return daemonApp->displayManager()->Sessions(this);
     }
 
-    DisplayManagerSession::DisplayManagerSession(const QString &name, const QString &seat, const QString &user, QObject *parent) : QObject(parent), m_name(name), m_seat(seat), m_user(user) {
-        // set path
-        m_path = DISPLAYMANAGER_SESSION_PATH + name.mid(7);
-
+    DisplayManagerSession::DisplayManagerSession(const QString &name, const QString &seat, const QString &user, QObject *parent) : QObject(parent), m_name(name), m_path(DISPLAYMANAGER_SESSION_PATH + name.mid(7)), m_seat(seat), m_user(user) {
```
Vogtinator, line 170: Same here, please newline before the `:` as it's too long now

src/daemon/DisplayManager.cpp
luis-pereira: @Vogtinator I didn't know what coding style to follow.
Will fixit.

Is this good to go ?

``` diff
diff --git a/src/daemon/DisplayManager.cpp b/src/daemon/DisplayManager.cpp
index 531d715..08570dc 100644
--- a/src/daemon/DisplayManager.cpp
+++ b/src/daemon/DisplayManager.cpp
@@ -129,7 +129,8 @@ namespace SDDM {
         }
     }
 
-    DisplayManagerSeat::DisplayManagerSeat(const QString &name, QObject *parent) : QObject(parent), m_name(name), m_path(DISPLAYMANAGER_SEAT_PATH + name.mid(4)) {
+    DisplayManagerSeat::DisplayManagerSeat(const QString &name, QObject *parent) :
+        QObject(parent), m_name(name), m_path(DISPLAYMANAGER_SEAT_PATH + name.mid(4)) {
         // create adaptor
         new SeatAdaptor(this);
```

Vogtinator: Usually the line break is before the `:`, but I don't mind either way. That line is perfectly readable now.

luis-pereira: I misread what you have written.

davidedmundson (approved):

PR #1305, opened 2020-10-18
Prefer Wayland sessions over X11 ones

Conan-Kudo:
As a general goal and preference, we want to make the change
to Wayland as the default and X11 as the fallback.

This change codifies that preference.

Reference: https://fedoraproject.org/wiki/Changes/WaylandByDefaultForPlasma

### DIFF
--- a/src/common/Session.cpp
+++ b/src/common/Session.cpp
@@ -131,13 +131,13 @@ namespace SDDM {
         m_desktopNames.clear();
 
         switch (type) {
-        case X11Session:
-            m_dir = QDir(mainConfig.X11.SessionDir.get());
-            m_xdgSessionType = QStringLiteral("x11");
-            break;
         case WaylandSession:
             m_dir = QDir(mainConfig.Wayland.SessionDir.get());
             m_xdgSessionType = QStringLiteral("wayland");
+            break;
+        case X11Session:
+            m_dir = QDir(mainConfig.X11.SessionDir.get());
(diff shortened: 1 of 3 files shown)


Conan-Kudo:
I've submitted this as a draft PR because I want to show you folks what I'm trying to do in an attempt to implement the correct automatic session selection for the Fedora 34 Change to switch to Wayland by default for Plasma. I am currently attempting to test this and see if it does what I expect. If it does, I'll mark it as a proper PR for review. 😉

Conan-Kudo:
This seems to work in my local tests, so I'm unmarking this as draft and declaring it ready for review!

davidedmundson (approved):

davidedmundson:
Code wise seems sane.
Though potentially won't this change the default for all first-run users of all distros?

Conan-Kudo:
Most distributions ship a configuration file to select which session to use by default or don't install Wayland sessions by default. In either scenario, this won't change anything. If _both_ Wayland and X11 sessions are installed, then the list will sort the Wayland session higher, but if the user has previously logged in with X11 session or selected it, it will not change.

I verified this with Fedora Rawhide and openSUSE Tumbleweed and it behaves the way people expect it to. I expect the same for Arch/Manjaro and Kubuntu, as both also do not install the Wayland session by default.

plfiorini (approved):

Conan-Kudo:
Thanks for merging my PR, @plfiorini!

PR #1311, opened 2020-11-03
GitHub workflow

plfiorini:
Switch from Travis CI to GitHub workflows.


jleclanche (approved):

.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+            #export CXX=clazy
+          fi
+          cmake .. \
+            -DBUILD_MAN_PAGES:BOOL=ON \
+            -DENABLE_PAM:BOOL=${{ matrix.pam }} \
+            -DLOGIN_DEFS_PATH:path="/usr/etc/login.defs"
+          make -j $(getconf _NPROCESSORS_ONLN)
+          sudo make install
+      - name: Notification for push events
+        uses: rectalogic/notify-irc@v1
```
Vogtinator, line 56: References to third-party actions should use the full sha1.

.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+          - OFF
+    runs-on: ubuntu-latest
+    container:
+      image: liridev/${{ matrix.container }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Dependencies
+        run: |
+          set -x
+          zypper install python-docutils
```
Vogtinator, line 38:
```suggestion
          zypper install python3-docutils
```

.github/workflows/build.yml
plfiorini: that's cumbersome, why not tags?

plfiorini: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses

it's not mandatory to use a full sha1

```
We strongly recommend that you include the version of the action you are using by specifying a Git ref, SHA, or Docker tag number. If you don't specify a version, it could break your workflows or cause unexpected behavior when the action owner publishes an update.
```

tags are fine:

```
    Using the commit SHA of a released action version is the safest for stability and security.
    Using the specific major action version allows you to receive critical fixes and security patches while still maintaining compatibility. It also assures that your workflow should still work.
    Using the default branch of an action may be convenient, but if someone releases a new major version with a breaking change, your workflow could break.
```

Vogtinator: Security reasons: https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions

plfiorini: ok thanks

i had to remove the irc notification though because it doesn't work, i wll search one that works later

PR #1314, opened 2020-11-04
Improve font config deserialization

Vogtinator:
Going from QString -> QVariant -> QFont isn't ideal and means that it tries
to create a QFont from a null/empty QString as well, resulting in a warning:
QFont::fromString: Invalid description '(empty)'
Just skip the QVariant step and call QFont::fromString directly.

### DIFF
--- a/src/greeter/GreeterApp.cpp
+++ b/src/greeter/GreeterApp.cpp
@@ -258,10 +258,13 @@ namespace SDDM {
         }
 
         // Set font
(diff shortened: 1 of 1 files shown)


davidedmundson (approved):

Vogtinator:
Apparently force-pushes don't cancel mandatory approvals - that's a bit worrying. I just fixed a duplicate call to `get()` though so in this case it's fine...

PR #1315, opened 2020-11-05
Use PAM's username

### LINKED ISSUE #1303: sddm ignores @domain suffix for adding local groups
Hello,

I have a sssd setup with two domains. In order to avoid conflicts, sssd adds a suffix @domain to each user/group.
I can login using both user1@domain1 or user1 and sssd chooses the default domain from option "domain_resolution_order".
In any case, the login will turn into user1@domain1 for everything else after the login. However, I'm having problems with sddm.

When I type user1 (without the domain suffix), SSSD does load domain groups nicely (the ones with @domain suffix) but not local groups (added to /etc/group) that references domain users (with @domain suffix).

* user: user1@domain1
* user domain groups: group1@domain1, group2@domain1...

local groups

```
/etc/groups:
video:x:33:user1@domain1
```

* When I log using sddm as 'user1', user1@domain1 gets groups group1@domain1, group2@domain1 ✗(video is missing)
* When I log using ssh as 'user1', user1@domain1 gets groups group1@domain1, group2@domain1, video ✓
* When I log using sddm as 'user1@domain1', user1@domain1 gets groups @domain1, group2@domain1, video ✓
* When I log using ssh as 'user1@domain1', user1@domain1 gets groups group1@domain1, group2@domain1, video ✓

I can workaround the issue adding entries, with and without domain suffix:

```etc/groups:
video:x:33:user1@domain1,user1
```

But it might also grant a permission to user1@domain2 (if domain_resolution_order is inverted) or a future local user1.

I guess somewhere during pam authentication, it updates the login informed by the user (from 'user1' to 'user1@domain1') but sddm code does not consider that.

I didn't test but my guess is that after this point:

https://github.com/sddm/sddm/blob/047ef56e5cfa757ebfcb03a248edad579564b5f3/src/helper/UserSession.cpp#L141

The username should come from pw.pw_name or rpw->pw_name (&pw == rpw when it works?) and not from username.
Here, for example:

https://github.com/sddm/sddm/blob/047ef56e5cfa757ebfcb03a248edad579564b5f3/src/helper/UserSession.cpp#L174-L178

It might need to do like it is already happening here:

https://github.com/sddm/sddm/blob/047ef56e5cfa757ebfcb03a248edad579564b5f3/src/helper/UserSession.cpp#L207

I'm using sddm-0.18.1-3.2.x86_64 from opensuse Tumbleweed. sddm uses xorg but I load a kde wayland session.

luizluca:
The provided username might differ from full username. This happens
for sssd with multiple domains where the username is suffixed with
@domain. However, the user can type only username without a suffix
and let sssd find the first username match according to
domain_resolution_order.

Fixes #1303

Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>

### DIFF
--- a/src/helper/UserSession.cpp
+++ b/src/helper/UserSession.cpp
@@ -175,13 +175,13 @@ namespace SDDM {
         // fetch session's user's groups
         int n_user_groups = 0;
         gid_t *user_groups = NULL;
-        if (-1 == getgrouplist(username.constData(), pw.pw_gid,
+        if (-1 == getgrouplist(pw.pw_name, pw.pw_gid,
                                NULL, &n_user_groups)) {
             user_groups = new gid_t[n_user_groups];
-            if ((n_user_groups = getgrouplist(username.constData(),
+            if ((n_user_groups = getgrouplist(pw.pw_name,
                                               pw.pw_gid, user_groups,
                                               &n_user_groups)) == -1 ) {
-                qCritical() << "getgrouplist(" << username << ", " << pw.pw_gid
+                qCritical() << "getgrouplist(" << pw.pw_name << ", " << pw.pw_gid
                             << ") failed";
                 exit(Auth::HELPER_OTHER_ERROR);
             }


Vogtinator (approved):
LGTM, but does this also address the case that when logging in with the fully qualified username you somehow ended up with `@domain` as group (which makes no sense to me, as that doesn't exist)?

luizluca:
> LGTM, but does this also address the case that when logging in with the fully qualified username you somehow ended up with `@domain` as group (which makes no sense to me, as that doesn't exist)?

group@domain is expected (for domain groups), even with spaces in name, as well as groups without suffix (local groups). They coexist nicely. This is a subset of what I get when I log using pristine SSH or sddm after this PR is merged:

```
uid=33123(luizluca@domain1) gid=30000(domain users@domain1) grupos=30000(domain users@domain1),30001(group2@domain1),16(dialout),33(video)
```

With:

```
/etc/group:
dialout:x:16:luizluca@domain1
video:x:33:luizluca@domain1
```

Before this PR, I was adding a non-existing local luizluca user to dialout/video as a workaround:

```
/etc/group:
dialout:x:16:luizluca@domain1,luizluca
video:x:33:luizluca@domain1,luizluca
```

Vogtinator:
What I'm referring to is this line from the issue:

> When I log using sddm as 'user1@domain1', user1@domain1 gets groups \@domain1, group2@domain1, video ✓

There you're mentioning a `@domain1`, so one without prefix, which I don't really understand.

luizluca:
> There you're mentioning a `@domain1`, so one without prefix, which I don't really understand.

My bad. I meant "I get groups with suffix @domain1 like group2@domain1".

Vogtinator (approved):

Vogtinator:
The change makes sense to me, though I found that the `getgrouplist(3)` man page also does it wrong:

```
           /* Fetch passwd structure (contains first group ID for user) */

           pw = getpwnam(argv[1]);
           if (pw == NULL) {
               perror("getpwnam");
               exit(EXIT_SUCCESS);
           }

           /* Retrieve group list */

           if (getgrouplist(argv[1], pw->pw_gid, groups, &ngroups) == -1) {
               fprintf(stderr, "getgrouplist() returned -1; ngroups = %d\n",
                       ngroups);
               exit(EXIT_FAILURE);
           }
```

Vogtinator:
At this point I actually suspect a bug in the SSSD NSS module, as this handles both `getpwnam` and `getgrouplist`, apparently inconsistently. Could you open a bug report there?

As a workaround this looks ok.

PR #1317, opened 2020-11-06
Changed "User name" to "Username"

KeetKhat:
Changed "User name" to "Username" in the french translation to match TextConstants.qml


plfiorini (approved):

david-geiger:
Shouldn't this fix not done also for all translations?

KeetKhat:
You're right, I done that #1357 
My apologies for all of theses commit messages but my only choice was to use the web interface because i don't have my PC with me at this time

PR #1322, opened 2020-11-09
Only use the base name for $DESKTOP_SESSION

### LINKED ISSUE #852: DESKTOP_SESSION is inconsistent with other display managers
Investigating why my `XDG_DATA_DIRS` environment variable appears so broken:

    $ echo $XDG_DATA_DIRS
    /usr/share//usr/share/xsessions/plasma:/usr/local/share/:/usr/share/:/var/lib/snapd/desktop

I discovered that, on Ubuntu, it is computed in part from DESKTOP_SESSION, in the `/etc/X11/Xsession.d/60x11-common_xdg_path` script. It boils down to:

    XDG_CONFIG_DIRS=/etc/xdg/xdg-"$DESKTOP_SESSION":/etc/xdg
    XDG_DATA_DIRS=/usr/share/"$DESKTOP_SESSION":/usr/local/share/:/usr/share/

Apparently other display managers and/or distros put a basename in DESKTOP_SESSION, such as `gnome`, `kde` or `kde-plasma` (though it is not clear who is doing what). sddm explicitly uses an absolute path (in Session::setTo() and then Session::desktopSession()), and seems to be the only one doing so. Is there a reason why? Should it align with others?

Some links:
* Ubuntu bug report from 2011 that led to the creation of the script: https://bugs.launchpad.net/ubuntu/+source/xorg/+bug/804734
* Ubuntu bug report from 2017 that complains about the content of the variable: https://bugs.launchpad.net/ubuntu/+source/xorg/+bug/1668722
* Similar bug report for KDE neon: https://bugs.kde.org/show_bug.cgi?id=366180

Vogtinator:
Other DMs don't use the path.

Fixes #852

WIP because untested.

### DIFF
--- a/src/common/Session.cpp
+++ b/src/common/Session.cpp
@@ -89,7 +89,7 @@ namespace SDDM {
 
     QString Session::desktopSession() const
     {
-        return fileName().replace(s_entryExtention, QString());
+        return QFileInfo(m_fileName).completeBaseName();
     }
 
     QString Session::desktopNames() const


davidedmundson (approved):

Vogtinator:
I'm wondering: Did anyone test this?

norbusan:
> I'm wondering: Did anyone test this?

Well, I tested it on our Debian packages, replaced the buggy patch with the one here, and it seems to work. At least it recognizes the session, uses the same full path as before, and doesn't crash. Anything else to check?

Vogtinator:
Nope, all is fine then. Thanks!

PR #1324, opened 2020-11-12
Retry starting the display server

### LINKED ISSUE #1316: [Regression] 0.19.0 fails to start while starting my X server
This is what systemctl status stays

```
● sddm.service - Simple Desktop Display Manager
     Loaded: loaded (/usr/lib/systemd/system/sddm.service; enabled; vendor preset: disabled)
     Active: active (running) since Thu 2020-11-05 18:09:49 CET; 40s ago
       Docs: man:sddm(1)
             man:sddm.conf(5)
   Main PID: 471 (sddm)
      Tasks: 2 (limit: 38081)
     Memory: 23.2M
     CGroup: /system.slice/sddm.service
             └─471 /usr/bin/sddm

de nov. 05 18:09:49 xps15 sddm[471]: Initializing...
de nov. 05 18:09:49 xps15 sddm[471]: Starting...
de nov. 05 18:09:49 xps15 sddm[471]: Logind interface found
de nov. 05 18:09:50 xps15 sddm[471]: Adding new display on vt 1 ...
de nov. 05 18:09:50 xps15 sddm[471]: Loading theme configuration from ""
de nov. 05 18:09:50 xps15 sddm[471]: Display server starting...
de nov. 05 18:09:50 xps15 sddm[471]: Adding cookie to "/var/run/sddm/{d02df92c-de3d-4dfc-84b9-1f563812434a}"
de nov. 05 18:09:50 xps15 sddm[471]: Running: /usr/bin/X -nolisten tcp -auth /var/run/sddm/{d02df92c-de3d-4dfc-84b9-1f563812434a} -background none -noreset -displayfd 17 -seat seat0 vt1
de nov. 05 18:09:50 xps15 sddm[471]: Failed to read display number from pipe
de nov. 05 18:09:50 xps15 sddm[471]: Could not start Display server on vt 1
```

I've traced the regression to commit 74394fd1cff26b0516af87539906a31bf2f49825

This is most probably a race since i can 100% reproduce the issue on my fast i9 but on the older i7 and i5 it works fine.

Another thing that points into a race is that if i going into vt 2, log in and issue a `systemctl restart sddm`, it brings sddm up just fine

Vogtinator:
Even if the CanGraphical property of a Seat is true, it's possible that it's
still too early for X to start, as it might need some driver or device which
isn't present yet.

Fixes #1316

### DIFF
--- a/src/daemon/Seat.cpp
+++ b/src/daemon/Seat.cpp
@@ -28,6 +28,7 @@
 
 #include <QDebug>
 #include <QFile>
+#include <QTimer>
 
 #include <functional>
 
@@ -52,7 +53,7 @@ namespace SDDM {
         return m_name;
     }
 
-    bool Seat::createDisplay(int terminalId) {
+    void Seat::createDisplay(int terminalId) {
         //reload config if needed
         mainConfig.load();
 
@@ -84,12 +85,24 @@ namespace SDDM {
         m_displays << display;
 
         // start the display
-        if (!display->start()) {
-            qCritical() << "Could not start Display server on vt" << terminalId;
-            return false;
+        startDisplay(display);
+    }
+
+    void Seat::startDisplay(Display *display, int tryNr) {
+        if (display->start())
+            return;
+
+        // It's possible that the system isn't ready yet (driver not loaded,
+        // device not enumerated, ...). It's not possible to tell when that changes,
+        // so try a few times with a delay in between.
+        qWarning() << "Attempt" << tryNr << "starting the Display server on vt" << display->terminalId() << "failed";
+
+        if(tryNr >= 3) {
+            qCritical() << "Could not start Display server on vt" << display->terminalId();
+            return;
         }
 
-        return true;
+        QTimer::singleShot(2000, display, [=] { startDisplay(display, tryNr + 1); });
     }
 
     void Seat::removeDisplay(Display* display) {
--- a/src/daemon/Seat.h
+++ b/src/daemon/Seat.h
@@ -35,13 +35,15 @@ namespace SDDM {
         const QString &name() const;
 
     public slots:
-        bool createDisplay(int terminalId = -1);
+        void createDisplay(int terminalId = -1);
         void removeDisplay(SDDM::Display* display);
 
     private slots:
         void displayStopped();
 
     private:
+        void startDisplay(SDDM::Display *display, int tryNr = 1);
+
         QString m_name;
 
         QVector<Display *> m_displays;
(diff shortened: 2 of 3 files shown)


plfiorini:
Can't we do something async? DisplayServer emtis a started() signal.

tsdgeos:
Unfortunately this doesn't fix things for me, it even breaks them more.

For some reason i can't even switch to VT2 anymore

albert-astals-cid-kdab:
The journalctl log is super weird though, i can't understnad why it has two attempt 0 
[with_patch.txt](https://github.com/sddm/sddm/files/5533723/with_patch.txt)

albert-astals-cid-kdab:
This seems to work better https://github.com/tsdgeos/sddm/commit/3eb5c0ece63b6942b7d7fafb4b479018cda3f9d7 for me.

No idea how dangerous is to spin an event loop there

Vogtinator:
I see what happens. The `QProcess` from the failed X server sent a signal that it exited, but no event loop is running so it's only delivered after the next attempt succeeded and it returns to the event loop. Unfortunately this signal causes X to be stopped again...

So what needs to be done is to wait until the old process finished and got cleaned up, and only then the next attempt can be made.

Vogtinator:
`XorgDisplayServer` doesn't really handle the case where `start()` fails and is later called again. It doesn't clean up properly and also doesn't emit a signal when the old process is gone and can be started again. So currently this waits 2s and hopes for the best.

Any change there would conflict with https://github.com/sddm/sddm/pull/1312, so I'd either wait until that is merged or add that to the commits here.

It would be ideal if when `start()` fails, it blocks until the `QProcess` is deleted, then this would "just work".

tsdgeos:
I can confirm this seems to fix it, it aint pretty but it isn't mega terribly "slow" either

tsdgeos:
Now, would it make sense to try to talk to the systemd people about this "wait for all cards to be ready before trying to start X"?

Vogtinator:
> Now, would it make sense to try to talk to the systemd people about this "wait for all cards to be ready before trying to start X"?

That's not possible, as it's not known in advance which cards are expected but not initialized yet. It's only known after the fact. Someone might plug in one over USB 10 minutes after booting, for instance.

davidedmundson (approved):
Thanks for following up on this

tsdgeos:
Can we get this landed and 0.19.1 released? 

Not being able to use the software at all seems like a huge regression to me.

tsdgeos:
Ping?

Vogtinator:
> XorgDisplayServer doesn't really handle the case where start() fails and is later called again. It doesn't clean up properly

I changed it so that when `XorgDisplayServer::start` fails, it makes sure that the process is stopped/stopping when it returns and added a check to ensure that an early call to `start` before the previous `process` finished does not happen.

tsdgeos:
ok, i can confirm that this branch still fixes the issue for me.

davidedmundson (approved):

PR #1329, opened 2020-11-18
Resolve full username before PAM

### LINKED ISSUE #1328: When informed login != real login, session does not appear in loginctl
Since #1315 got merged, I can login with user or user@domain and have groups initialized correctly.
However, I noticed that pulseaudio was not working. I saw that XDG_RUNTIME_DIR was not set and
/var/run/user/<userid> was missing. Also, listing session with loginctl, I didn't see my current x11 session started
by sddm. Using user@domain I have no issues. The same with other session managers or even ssh.

I failed to discover where the "user provided login" was being used and even where a seat should be created.
If the "user provided login" is being used outside UserSession, the real login returned by pam might need to be saved globally for other to use.

luizluca:
Sssd NSS can map "username" to "username@domain". Although PAM authenticates
"username" (using SSSD module), it fails to properly setup the user
environment. PAM needs to receive the real username, in this case,
"username@domain".

Fixes #1328

Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>

### DIFF
--- a/src/helper/HelperApp.cpp
+++ b/src/helper/HelperApp.cpp
@@ -42,6 +42,8 @@
 #include <utmpx.h>
 #include <QByteArray>
 
+#include <pwd.h>
+
 namespace SDDM {
     HelperApp::HelperApp(int& argc, char** argv)
             : QCoreApplication(argc, argv)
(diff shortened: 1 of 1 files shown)


luizluca:
I shamelessly copied the getpwnam_r code from UserSesssion into HelperApp. Maybe it might be a better approach.

luizluca:
As a side-effect, the saved "last used username" will have the @domain appended.

davidedmundson:
We need to remove the relevant bit of code in `UserSession::setupChildProcess` that this is copied from.

That takes the qobject_cast<HelperApp*>(parent())->user() and will end up looking up in getpwnam_r twice, only now with the already looked up username.

luizluca:
> We need to remove the relevant bit of code in `UserSession::setupChildProcess` that this is copied from.

I copied it from `UserSession::setup`, but `UserSession::setupChildProcess`  might be an old name.

> That takes the qobject_cast<HelperApp*>(parent())->user() and will end up looking up in getpwnam_r twice, only now with the already looked up username.

@davidedmundson, This seems to be something not too complex but I couldn't do it with my shallow sddm knowledge. I didn't get the relationship between helpers.

I hope "we need to..." does not depend on me :-)

davidedmundson:
>I hope "we need to..." does not depend on me :-)

It blocks this MR.

PR #1260, opened 2020-05-29
Revert "Fixes #879 - Adds sourcing of /etc/profile to fish"

Vogtinator:
Reverts sddm/sddm#1047

Fixes https://github.com/sddm/sddm/issues/1059

Issue #1059, opened 2018-07-27
fish dies when sourcing /etc/profile


maxyz:
As mentioned by the fish developers, fish is not a POSIX 1003.1 compatible shell, thus sourcing /etc/profile with it only leads to syntax errors [Fish bug #3665](https://github.com/fish-shell/fish-shell/issues/3665)

As such, please consider reverting https://github.com/sddm/sddm/commit/7a27e1a4ad291725c3905f1e2e18d0fcaf7943c5

This will reopen #879, wich might need a different solution, maybe something like:
```shell
$SHELL --login -c "/bin/sh -c '[ -f /etc/profile ] && . /etc/profile; export -p' > $xsess_tmp"
```

Happy hacking,

PR #1331, opened 2020-11-23
Re-add fish /etc/profile sourcing

soredake:
Fixes https://github.com/sddm/sddm/issues/879
Also added sourcing of `~/.profile`, ubuntu adds `~/.local/bin` to PATH in this file, so this will be useful.

### DIFF
--- a/data/scripts/Xsession
+++ b/data/scripts/Xsession
@@ -39,6 +39,8 @@ case $SHELL in
     rm -f $xsess_tmp
     ;;
   */fish)
+    [ -f /etc/profile ] && . /etc/profile
+    [ -f $HOME/.profile ] && . $HOME/.profile
     xsess_tmp=`mktemp /tmp/xsess-env-XXXXXX`
     $SHELL --login -c "/bin/sh -c 'export -p' > $xsess_tmp"
     . $xsess_tmp
--- a/data/scripts/wayland-session
+++ b/data/scripts/wayland-session
@@ -43,6 +43,8 @@ case $SHELL in
     rm -f $wlsess_tmp
     ;;
   */fish)
+    [ -f /etc/profile ] && . /etc/profile
+    [ -f $HOME/.profile ] && . $HOME/.profile
     xsess_tmp=`mktemp /tmp/xsess-env-XXXXXX`
     $SHELL --login -c "/bin/sh -c 'export -p' > $xsess_tmp"
     . $xsess_tmp


soredake:
@Vogtinator @bastimeyer

Vogtinator:
Question is why this would not reintroduce #1059. I wouldn't merge this before #1059 didn't get sufficiently debugged.

soredake:
I don't know how it can die if this file is run with /bin/sh....

soredake:
Related: https://salsa.debian.org/qt-kde-team/3rdparty/sddm/-/merge_requests/2

soredake:
@maxyz please test this

bastimeyer:
Since this would revert the changes of #1260, it would resolve the issue I've described there:
https://github.com/sddm/sddm/pull/1260#issuecomment-723678353

Regarding the PR's changes and the addition of `$HOME/.profile`, you probably want to put `$HOME` into quotation marks to make sure that potential spaces in `$HOME` will work correctly.

Not sure what #1059 is about, as `/etc/profile` is not being sourced by fish, at least to my understanding.

Conan-Kudo:
The issue is that you're _forcing_ fish to source it, and that causes it to crash.

soredake:
@Conan-Kudo forcing where?

soredake:
Right now two users can confirm that this script is run with /bin/sh and env is correctly passed to fish, and only one (@maxyz) saying (in 2018...) that it does not work without explaining anything, i don't think #1059 can be debugged without original reporter.

soredake:
I see two possible routes:
1. Close this pr as user who reported breakage is inactive and will not explain how script run by /bin/sh can break fish.
2. Merge this as there confirmation that all works as intended. I personally tried to launch X11 and Wayland sessions of KDE Plasma with pr many times, all successful with inherited env.

soredake:
@Vogtinator so I've installed kde neon in virtualbox, installed fish, set user shell to fish, edited files to source /etc/profile|$HOME/.profile, rebooted vm, logged into wayland|x11 session and woah, everything is working, $PATH is filled with `.local/bin` as $HOME/.profile is sourced correctly, $XDG_DATA_DIRS is filled with flatpak dirs, is this enough proof that shebang `#!/bin/sh` in session files is working correctly and fish isn't breaking anything?

echo of env vars and user shell
Screenshot_20210217_110557
flatpak apps are showing fine as $XDG_DATA_DIRS is filled with flatpak dirs
Screenshot_20210217_110553

Vogtinator:
That means #1059 is not reproducible, which implies that @maxyz either had an invalid setup (`fish` as `/bin/sh`) or reported the issue without actually having it. I don't see any issue with that, so once this PR got rebased, I'll accept it.

soredake:
@Vogtinator pr rebased

soredake:
> That means #1059 is not reproducible, which implies that @maxyz either had an invalid setup (`fish` as `/bin/sh`) or reported the issue without actually having it. I don't see any issue with that, so once this PR got rebased, I'll accept it.

Fish as /bin/sh would break entire system (rn i found 317 scripts with /bin/sh shebang), so it's not realistic, but reporting without actually having this problem is possible.

Vogtinator (approved):

Slickytail:
This change breaks sddm for me and prevents me from logging in. I use fish and sway, and my `/bin/sh` is not fish. The `wayland-session.log` indicates that the `xsess-env-XXXXXX` tempfile is trying to set `PROFILEREAD` -- which is exported readonly in `/etc/profile`.

Vogtinator:
> I use fish and sway, and my /bin/sh is not fish.

That is an invalid setup, because fish is not a POSIX compatible shell and `/bin/sh` is defined to be one. So you have to change that.

Slickytail:
> > I use fish and sway, and my /bin/sh is not fish.
> 
> That is an invalid setup, because fish is not a POSIX compatible shell and `/bin/sh` is defined to be one. So you have to change that.

Are you saying fish users are supposed to symlink `/bin/sh` to fish?? What else would I use as `/bin/sh` besides normal shell?

Vogtinator:
> > > I use fish and sway, and my /bin/sh is not fish.
> > 
> > 
> > That is an invalid setup, because fish is not a POSIX compatible shell and `/bin/sh` is defined to be one. So you have to change that.
> 
> Are you saying fish users are supposed to symlink `/bin/sh` to fish?? What else would I use as `/bin/sh` besides normal shell?

Apparently I read your comment too quickly - I missed the `not`. In the case that your `/bin/sh` is not fish, what is it instead?

> The wayland-session.log indicates that the xsess-env-XXXXXX tempfile is trying to set PROFILEREAD -- which is exported readonly in /etc/profile.

While that is an error, it should actually be harmless without `set -e`. So the reason it breaks is something else.

Slickytail:
As far as I can tell, my `sh` is bash running in sh mode (the prompt says `sh-5.1$`). 

That's the only thing there is in `wayland-session.log`. Are there any other logfiles I should post?

Vogtinator:
> That's the only thing there is in wayland-session.log. Are there any other logfiles I should post?

The systemd journal might be useful if you have it. You could add `set -x` to `wayland-session` for more output, my guess is that some environment variable breaks it. Maybe something fish specific causes startup to fail.

gitfib:
I'm having the same issue on openSUSE Tumbleweed. `PROFILEREAD` is a readonly export by openSUSE's /etc/profile, and the PR attempts to source it twice.
It's possible to use `diff` in order to only source variables that have changed, but that would add a new dependency, right? Should it be fixed in downstream?

Slickytail:
> I'm having the same issue on openSUSE Tumbleweed.

I'm also using openSUSE Tumbleweed. Seems like it might be a problem with Tumbleweed's `profile`?

FYI: The default `~/.profile` for Tumbleweed includes the line `test -z "$PROFILEREAD" && . /etc/profile || true` which sources `/etc/profile`. Maybe this is the source of the issue?

Vogtinator:
I assume this is only with Wayland sessions?

gitfib:
I don't know why but yes, only encountered it in Wayland. I was also trying the Wayland (not x11) in a tty, based on what sddm said it was executing in journalctl
Actually it can be reproduced in a simple:
```bash
fish$ sh # which is a symlink to bash
sh$ . /etc/profile
sh$ . <(export -p)
```

Slickytail:
Yes, only with wayland. I can start an icewm session just fine but sway
won't start.

ZaWertun:
> I'm having the same issue on openSUSE Tumbleweed. `PROFILEREAD` is a readonly export by openSUSE's /etc/profile, and the PR attempts to source it twice. It's possible to use `diff` in order to only source variables that have changed, but that would add a new dependency, right? Should it be fixed in downstream?

It could be solved by just exporting variable instead of exporting and reassigning it.

I.e. use `export A`, and not `export A=1` in temporary script.
All variables are already defined there is no need to reassign them, it could be inspected with `set -vx` option & eximining the output of the script.

So my solution is:
```patch
--- /usr/share/sddm/scripts/wayland-session.orig        2022-08-24 23:27:18.402899836 +0300
+++ /usr/share/sddm/scripts/wayland-session     2022-08-24 23:14:19.502945427 +0300
@@ -34,7 +34,7 @@
     [ -f /etc/profile ] && . /etc/profile
     [ -f $HOME/.profile ] && . $HOME/.profile
     xsess_tmp=`mktemp /tmp/xsess-env-XXXXXX`
-    $SHELL --login -c "/bin/sh -c 'export -p' > $xsess_tmp"
+    $SHELL --login -c "/bin/sh -c 'export -p | cut -d= -f1' > $xsess_tmp"
     . $xsess_tmp
     rm -f $xsess_tmp
     ;;
```

Works fine for me on test VM with openSUSE Tumbleweed.

ZaWertun:
Simple example to demonstrate what has been failing in this script:
```shell
$ /bin/sh -c 'readonly A=1; export A=1'
/bin/sh: line 1: A: readonly variable
```

PR #1367, opened 2021-02-15
Iterate PR using Wayland for the greeter

aleixpol:
This is mostly based on the work @plfiorini did in https://github.com/sddm/sddm/pull/616.

Biggest changes are:
* Decoupling the compositor implementation from SDDM by making it configurable.
* Executing both the compositor and the client on from same pam-authenticated process, otherwise our compositor gets killed when the greeter is launched.

### DIFF
--- a/src/auth/Auth.h
+++ b/src/auth/Auth.h
@@ -52,7 +52,6 @@ namespace SDDM {
         Q_OBJECT
         // not setting NOTIFY for the properties - they should be set only once before calling start
         Q_PROPERTY(bool autologin READ autologin WRITE setAutologin NOTIFY autologinChanged)
-        Q_PROPERTY(bool displayServer READ isDisplayServer WRITE setDisplayServer NOTIFY displayServerChanged)
         Q_PROPERTY(bool greeter READ isGreeter WRITE setGreeter NOTIFY greeterChanged)
         Q_PROPERTY(bool verbose READ verbose WRITE setVerbose NOTIFY verboseChanged)
         Q_PROPERTY(QString cookie READ cookie WRITE setCookie NOTIFY cookieChanged)
@@ -92,7 +91,6 @@ namespace SDDM {
         static void registerTypes();
 
         bool autologin() const;
-        bool isDisplayServer() const;
         bool isGreeter() const;
         bool verbose() const;
         const QString &cookie() const;
@@ -127,14 +125,6 @@ namespace SDDM {
         void setAutologin(bool on = true);
 
         /**
-         * Set mode to display server
-         * This will bypass authentication checks and will wait
-         * until the display server is actually started, for example
-         * when the Wayland compositor socket is created
-         */
-        void setDisplayServer(bool on = true);
-
-        /**
          * Set mode to greeter
          * This will bypass authentication checks
          */
@@ -154,9 +144,14 @@ namespace SDDM {
 
         /**
         * Set the session to be started after authenticating.
-        * @param path Path of the session executable to be started
+        * @param command Path of the session executable to be started
         */
-        void setSession(const QString &path);
+        void setSession(const QString &command);
+
+        /**
+         * Set the @p command to be executed before launching the greeter
+         */
+        void setCompositor(const QString &command);
 
         /**
          * Set the display server cookie, to be inserted into the user's $XAUTHORITY
@@ -170,15 +165,17 @@ namespace SDDM {
         */
         void start();
 
+        void stop();
+
     Q_SIGNALS:
         void autologinChanged();
         void greeterChanged();
-        void displayServerChanged();
         void verboseChanged();
         void cookieChanged();
         void userChanged();
         void sessionChanged();
         void requestChanged();
+        void compositorChanged();
 
         /**
         * Emitted when authentication phase finishes
--- a/src/auth/Auth.h
+++ b/src/auth/Auth.h
@@ -52,6 +52,7 @@ namespace SDDM {
         Q_OBJECT
         // not setting NOTIFY for the properties - they should be set only once before calling start
         Q_PROPERTY(bool autologin READ autologin WRITE setAutologin NOTIFY autologinChanged)
+        Q_PROPERTY(bool displayServer READ isDisplayServer WRITE setDisplayServer NOTIFY displayServerChanged)
         Q_PROPERTY(bool greeter READ isGreeter WRITE setGreeter NOTIFY greeterChanged)
         Q_PROPERTY(bool verbose READ verbose WRITE setVerbose NOTIFY verboseChanged)
         Q_PROPERTY(QString cookie READ cookie WRITE setCookie NOTIFY cookieChanged)
@@ -91,6 +92,7 @@ namespace SDDM {
         static void registerTypes();
 
         bool autologin() const;
+        bool isDisplayServer() const;
         bool isGreeter() const;
         bool verbose() const;
         const QString &cookie() const;
@@ -123,6 +125,14 @@ namespace SDDM {
         * @param on true if should autologin
         */
         void setAutologin(bool on = true);
+
+        /**
+         * Set mode to display server
+         * This will bypass authentication checks and will wait
+         * until the display server is actually started, for example
+         * when the Wayland compositor socket is created
+         */
+        void setDisplayServer(bool on = true);
 
         /**
          * Set mode to greeter
@@ -163,6 +173,7 @@ namespace SDDM {
     Q_SIGNALS:
         void autologinChanged();
         void greeterChanged();
+        void displayServerChanged();
         void verboseChanged();
         void cookieChanged();
         void userChanged();
--- a/src/common/Configuration.h
+++ b/src/common/Configuration.h
@@ -44,6 +44,9 @@ namespace SDDM {
                                                                                                    "NOTE: Currently ignored if autologin is enabled."));
         Entry(InputMethod,         QString,     QStringLiteral("qtvirtualkeyboard"),                   _S("Input method module"));
         Entry(Namespaces,          QStringList, QStringList(),                                  _S("Comma-separated list of Linux namespaces for user session to enter"));
+        Entry(DisplayServer,       QString,     _S("x11"),                                      _S("Which display server should be used.\n"
+                                                                                                   "NOTE: Wayland support is currently considered experimental.\n"
+                                                                                                   "Valid values are: x11, wayland."));
         //  Name   Entries (but it's a regular class again)
         Section(Theme,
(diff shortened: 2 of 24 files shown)


plfiorini:
Thanks for resurrecting this, I will give it a spin

src/auth/Auth.h
```diff
@@ -124,6 +124,14 @@ namespace SDDM {
         */
         void setAutologin(bool on = true);
 
+        /**
+         * Set mode to display server
+         * This will bypass authentication checks and will wait
+         * until the display server is actually started, for example
+         * when the Wayland compositor socket is created
+         */
+        void setDisplayServer(bool on = true);
```
davidedmundson: Is this still used?

davidedmundson (changes requested):
Bugs we need to fix before landing:
 - From what I can tell, starting an X11 session isn't going to work? 
 
 - We might want some fallback to X if the wayland greeter fails?
 
Bugs that I am happy to say we address in the future

 - Default keyboard layout won't be loaded from localectl on Plasma
 (arguably this is due to a very weird design decision in locale1 not being in kwin but in startplasma)
 
 - Multi-seat will be broken. That will have two users on wayland-0

plfiorini:
@davidedmundson Multi-seat: the compositor should create a socket whose name references the seat, like `sddm-wayland-$SEAT`

zzag:
@plfiorini I believe the issue is that there is no generic way to say "compositor $foo, please bind the following socket $xyz"

src/helper/HelperStartWayland.cpp
```diff
@@ -0,0 +1,85 @@
...
+{
+    QCoreApplication app(argc, argv);
+    if (argc != 3) {
+        QTextStream(stderr) << "Wrong number of arguments\n";
+        return 33;
+    }
+    const QString compositor = app.arguments()[1];
+    const QString client = app.arguments()[2];
+
+    auto watcher = new SDDM::WaylandSocketWatcher(&app, QProcessEnvironment::systemEnvironment());
```
plfiorini, line 62: Set `WAYLAND_SOCKET` to `sddm-wayland-$SEAT` here or where the helper is called.
Here is probably better as you can set it only when the compositor process is spawned.

plfiorini (changes requested):

plfiorini:
> @plfiorini I believe the issue is that there is no generic way to say "compositor $foo, please bind the following socket $xyz"

Set `WAYLAND_DISPLAY` on the user session.

zzag:
This is going to work only if the compositor calls `wl_display_add_socket()` with a `nullptr` socket name. However, many compositors nowadays use `wl_display_add_socket_auto()`, which doesn't check WAYLAND_DISPLAY.

plfiorini:
> This is going to work only if the compositor calls `wl_display_add_socket()` with a `nullptr` socket name. However, many compositors nowadays use `wl_display_add_socket_auto()`, which doesn't check WAYLAND_DISPLAY.

But `wl_display_add_socket_auto()` scans through `$XDG_RUNTIME_DIR`  for an unused wayland-$d (where $d is < 32) and binds the first available.

EDIT: A compositor should also handle multi-seat themeselves reading `$XDG_SEAT`

plfiorini:
Having the compositor configurable will give us more flexibility, but the default runs `weston` with its shell.
We should be using fullscreen-shell since I wrote a shell integration for Qt upstream a long time ago.

src/common/Configuration.h
```diff
@@ -79,6 +82,8 @@ namespace SDDM {
             Entry(SessionCommand,      QString,     _S(WAYLAND_SESSION_COMMAND),                _S("Path to a script to execute when starting the desktop session"));
 	    Entry(SessionLogFile,      QString,     _S(".local/share/sddm/wayland-session.log"),_S("Path to the user session log file"));
             Entry(EnableHiDPI,         bool,        false,                                      _S("Enable Qt's automatic high-DPI scaling"));
+            Entry(CompositorCommand,   QString,     _S("weston"),                               _S("Choose which Wayland compositor we will be starting"));
```
plfiorini: `weston --shell=fullscreen-shell.so`

plfiorini: and then run the greeter setting `QT_WAYLAND_SHELL_INTEGRATION=fullscreen-shell-v1`

plfiorini (changes requested):

plfiorini:
There is no keyboard layout switching.
As far as I know there is no API to list the keymaps, we should be doing it ourselves.
We should also make a protocol between the greeter and the compositor for switching it.

EDIT: Not necessarily right now, this can be something to do in the future.
The protocol could also be useful for installers like Calamares.

plfiorini:
@aleixpol Why not doing a small compositor for SDDM, it's so easy with QtWayland?

That would give us maximum control over it.
We could keep the setting so that users can use their own compositor, after all we do the same thing for the X display server.

aleixpol:
> @aleixpol Why not doing a small compositor for SDDM, it's so easy with QtWayland?

It's always a possibility. My impression is that whatever we do will meet escalating requirements. I know that for example by using kwin we could easily integrate the KScreen configuration between both easily. Also things like keyboard switching we can implement just once. With a KDE mindset we want to avoid the risk of maintaining two stacks.

aleixpol:
With my last commit I made it so Xorg starts. Only one problem: it doesn't work.

This is the tail of my Xorg.0.log, maybe someone here knows something about starting X sessions ^^'.

```
[   103.794] (II) modeset(0): Output DP-1 using initial mode 3840x2160 +0+0
[   103.794] (II) modeset(0): Output HDMI-1 using initial mode 1920x1200 +3840+0
[   103.794] (==) modeset(0): Using gamma correction (1.0, 1.0, 1.0)
[   103.794] (==) modeset(0): DPI set to (96, 96)
[   103.794] (II) Loading sub module "fb"
[   103.794] (II) LoadModule: "fb"
[   103.794] (II) Loading /usr/lib/xorg/modules/libfb.so
[   103.794] (II) Module fb: vendor="X.Org Foundation"
[   103.794]    compiled for 1.20.10, module version = 1.0.0
[   103.794]    ABI class: X.Org ANSI C Emulation, version 0.4
[   103.794] (EE) modeset(0): drmSetMaster failed: Device or resource busy
[   103.794] (EE) 
Fatal server error:
[   103.794] (EE) AddScreen/ScreenInit failed for driver 0
[   103.794] (EE) 
```

plfiorini:
We should be using rootless Xorg (Xorg with user privileges): https://wiki.archlinux.org/index.php/xorg#Rootless_Xorg

aleixpol:
> We should be using rootless Xorg (Xorg with user privileges)

I gave it a go, not really working (yet?). I hope we could do both approaches separately so this doesn't become a big fat PR supposed to solve every problem.

https://github.com/aleixpol/sddm/pull/1

plfiorini:
> > We should be using rootless Xorg (Xorg with user privileges)
> 
> I gave it a go, not really working (yet?). I hope we could do both approaches separately so this doesn't become a big fat PR supposed to solve every problem.
> 
> [aleixpol#1](https://github.com/aleixpol/sddm/pull/1)

I don't remember what was missing from my patch.

What is missing?

aleixpol:
> > > We should be using rootless Xorg (Xorg with user privileges)
> > 
> > 
> > I gave it a go, not really working (yet?). I hope we could do both approaches separately so this doesn't become a big fat PR supposed to solve every problem.
> > [aleixpol#1](https://github.com/aleixpol/sddm/pull/1)
> 
> I don't remember what was missing from my patch.
> 
> What is missing?

I cannot get it to work. Also it seems to be launching separate Auth for the session and the X server which this PR already addresses for Wayland because it doesn't work.

plfiorini:
> > > > We should be using rootless Xorg (Xorg with user privileges)
> > > 
> > > 
> > > I gave it a go, not really working (yet?). I hope we could do both approaches separately so this doesn't become a big fat PR supposed to solve every problem.
> > > [aleixpol#1](https://github.com/aleixpol/sddm/pull/1)
> > 
> > 
> > I don't remember what was missing from my patch.
> > What is missing?
> 
> I cannot get it to work. Also it seems to be launching separate Auth for the session and the X server which this PR already addresses for Wayland because it doesn't work.

I'm updating the original Xorg rootless PR so we can reuse some bits here.
EDIT: #1371

src/helper/HelperStartWayland.cpp
```diff
@@ -0,0 +1,96 @@
...
+    QObject::connect(app, &QCoreApplication::aboutToQuit, p, [p] {
+        p->terminate();
+        if (!p->waitForFinished()) {
+            p->kill();
+        }
+    });
+    p->start(program, args);
+    if (!p->waitForStarted()) {
+        QTextStream(stderr) << "Failed to start: " << cmd << ". " << p->errorString() << Qt::endl;
+        return true;
```
plfiorini: I guess it was meant to be `return false`

src/helper/HelperStartWayland.cpp
```diff
@@ -0,0 +1,96 @@
...
+    QProcess* p = new QProcess(app);
+    p->setProcessChannelMode(QProcess::ForwardedChannels);
+    QObject::connect(app, &QCoreApplication::aboutToQuit, p, [p] {
+        p->terminate();
+        if (!p->waitForFinished()) {
+            p->kill();
+        }
+    });
+    p->start(program, args);
+    if (!p->waitForStarted()) {
```
plfiorini: 10s timeout?

src/helper/HelperStartWayland.cpp
```diff
@@ -0,0 +1,96 @@
...
+{
+    QStringList args = QProcess::splitCommand(cmd);
+    const QString program = args.takeFirst();
+
+    auto app = QCoreApplication::instance();
+    QProcess* p = new QProcess(app);
+    p->setProcessChannelMode(QProcess::ForwardedChannels);
+    QObject::connect(app, &QCoreApplication::aboutToQuit, p, [p] {
+        p->terminate();
+        if (!p->waitForFinished()) {
```
plfiorini: 5s timeout?

plfiorini (changes requested):

tpgxyz:
Hi, I am happy to report that sddm works with Wayland greeter. I'm running it for a whole day.

For an experiment I build sddm rpm for OpenMandriva with this PR and adjusted sddm.conf to use Wayland as display.

One thing i noticed, is that after leaving greeter to plasma Wayland i see for 1 sec my TTY line prompt.

Here is the video https://youtu.be/QKUp-rr8h-w

plfiorini:
Now that there's #1379 this is obsolete.
Only thing missing from #1379 compared to this one, is the compositor environment setting that can be added later.

PR #1368, opened 2021-02-19
Allow one-click passwordless login (fix #751)

### LINKED ISSUE #751: Passwordless Login
Please allow a passwordless login. This is a common usecase for family members. It has been already reported a few times (e.g #2, #577, #505), but the issues have always been closed.

I am not talking about autologin here, but a passwordless login for a second or third user

thanks!

R0bur:
Login acts by click on the user icon.
For passwordless login (https://github.com/sddm/sddm/issues/751):
1) the user must be a member of the "nopasswdlogin" group;
2) passwordless login must be allowed in /etc/pam.d/sddm.

### DIFF
--- a/src/greeter/theme/Main.qml
+++ b/src/greeter/theme/Main.qml
@@ -87,11 +87,15 @@ Rectangle {
                 onLogin: sddm.login(model.name, password, sessionIndex);
 
                 MouseArea {
-                    anchors.fill: parent
-                    onClicked: {
+                    anchors.left: parent.left; anchors.top: parent.top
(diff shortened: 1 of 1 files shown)


plfiorini (changes requested):
This is not the right approach, you are doing the login everytime the picture box is clicked and you will do it for every user.
Besides the whole concept of the nopasswdlogin group is wrong because it will make SDDM skip the password even for those accounts that have one.
What we really want is to detect when an account has no password and display a button instead of the password field.

PR #1369, opened 2021-02-19
Upstream freebsd patches

adriaandegroot:
These are downstream (packaging) patches that we carry against the 0.19 release. I can't tell whether they should go to *devel* or *master*, so I picked one. These patches were created by @arrowd , I'm just pushing them here (come to think of it, I should have set `--author`).

The **reason** I'm throwing them out here is that @aleixpol was shaming me for having downstream patches, and @davidedmundson asked me to take a look at @Vogtinator 's rework-xauth #1230 for FreeBSD -- except that doesn't apply on top of our patches, so I need to untangle bits **anyway**.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,9 @@ add_feature_info("PAM" PAM_FOUND "PAM support")
 include(CheckFunctionExists)
 check_function_exists(getspnam HAVE_GETSPNAM)
 
+# XAU
+pkg_check_modules(LIBXAU REQUIRED "xau")
+
 # XCB
 find_package(XCB REQUIRED)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,6 +90,17 @@ check_function_exists(getspnam HAVE_GETSPNAM)
 
 # XAU
 pkg_check_modules(LIBXAU REQUIRED "xau")
+# If possible, convert xau to a full path; if not, keep the existing
+# setting (which is probably -lXau .. which fails on FreeBSD because
+# the **directory** it is in is not searched.
+find_library(_xau_lib
+        NAMES ${LIBXAU_LIBRARIES} xau Xau
+        HINTS ${LIBXAU_LIBRARY_DIRS} ${LIBXAU_INCLUDE_DIR}/../lib ${LIBXAU_LIBRARIES}
+)
+if(_xau_lib)
+    set(LIBXAU_LIBRARIES ${_xau_lib})
+endif()
+
 
 # XCB
 find_package(XCB REQUIRED)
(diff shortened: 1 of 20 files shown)


adriaandegroot:
The latter three commits are @Vogtinator 's Xauth, followed by build fixes; those could be incorporated in the upstream commit, although I had to massage it a little to fit with the other changes from FreeBSD (e2e2684 clashes with it).

src/helper/UserSession.cpp
```diff
@@ -50,7 +50,12 @@ namespace SDDM {
...
         } else if (env.value(QStringLiteral("XDG_SESSION_TYPE")) == QLatin1String("x11")) {
-            const QString cmd = QStringLiteral("%1 \"%2\"").arg(mainConfig.X11.SessionCommand.get()).arg(m_path);
+#ifdef Q_OS_FREEBSD
+            const QString cmdFormat = QStringLiteral("%1 %2");
+#else
+            const QString cmdFormat = QStringLiteral("%1 \"%2\"");
+#endif
+            const QString cmd = cmdFormat.arg(mainConfig.X11.SessionCommand.get()).arg(m_path);
             qInfo() << "Starting:" << cmd;
             QProcess::start(cmd);
```
aleixpol, line 90: Should be ported to:
`QProcess::start(mainConfig.X11.SessionCommand.get(), {m_path});`

The current QProcess::start(cmd) that we're using everywhere is deprecated anyway. I'm not sure if SessionCommand can have arguments, if so we'll need to use QProcess::splitCommand().

aleixpol (commented):
LGTM otherwise. I guess it would make sense to merge Vogt's patch first though.

CMakeLists.txt
```diff
@@ -90,6 +90,17 @@ check_function_exists(getspnam HAVE_GETSPNAM)
 
 # XAU
 pkg_check_modules(LIBXAU REQUIRED "xau")
+# If possible, convert xau to a full path; if not, keep the existing
+# setting (which is probably -lXau .. which fails on FreeBSD because
+# the **directory** it is in is not searched.
+find_library(_xau_lib
```
arrowd, line 96: According to [docs](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html), `LIBXAU_LINK_LIBRARIES` should contain libraries with absolute paths. Doesn't that work?

arrowd (approved):
> These patches were created by @arrowd , I'm just pushing them here 

Not really, exactly these patches were already in place when I started fiddling with SDDM. My own patches were submitted in https://github.com/sddm/sddm/pull/1234 , so no need in changing commits author.

The PR looks OK, except "Redesign xauth handling", for which I can't provide a review.

Thanks for upstreaming this!

plfiorini:
We use git flow, so PRs have to target `develop`.
Also, why you did include @Vogtinator's patch for Xauth here, can you make a couple of separate PR's?

plfiorini (changes requested):
See comments

Pointedstick:
Ping @adriaandegroot

adriaandegroot:
There's no point in leaving this open in parallel with #1230

Issue #1281, opened 2020-07-30
ReuseSession=true is broken for multiple sessions


Vogtinator:
If sddm decides to reuse a session after successful authentication by unlocking and activating it, there are three processes competing for the VT switch:

1. `systemd-logind` is told to activate the target session
2. The `Xorg` display server for the greeter wants to restore the old VT on shutdown
3. When the greeter display is stopped, `sddm` tries to switch to the latest display

In the simplest (and the only one which currently works), all of those switch to the same VT. In other cases, especially when there is more than a single active session, it's more likely broken than working.

1 is the intended behaviour, nothing to fix there.
2 can be avoided by passing `-novtswitch`, but it appears like Xorg ignores this for some reason... It would be possible to do the reusing only after `Xorg` shut down, but that will cause flicker.
3 can be fixed by checking whether the stopped display was on the current VT and not doing anything otherwise.

grahamperrin:
Also #917

Pointedstick:
Is this fixed by https://github.com/sddm/sddm/pull/1690?

Vogtinator:
In combination with #1460, #1644 and #1692, hopefully.

PR #1371, opened 2021-02-22
Run X11 display server without root privileges

### LINKED ISSUE #246: Fix rootless xorg support
```
[10:48:14] <AnAkkk> after this change in the Xorg 1.16 arch package, SDDM no longer works correctly because at the time that X starts, it can't find the nvidia module https://bugs.archlinux.org/task/41259
[10:48:19] <AnAkkk> is that something we need to fix in sddm?
[10:48:44] <AnAkkk> I have to restart sddm with systemctl every time to make it work
[11:07:08] <AnAkkk> ok, actually it's related to the dkms systemd service + nvidia-dkms
[11:07:18] <AnAkkk> that I use to rebuild the nvidia module on kernel updates
[11:07:38] <AnAkkk> it seems to unload the nvidia module, X can't find it
```

plfiorini:
Introduce an alternative display server option to run the greeter the
X11 server as an unprivileged user.

Closes: #246


src/daemon/XorgUserDisplayServer.h
```diff
@@ -0,0 +1,61 @@
...
+#define SDDM_XORGUSERDISPLAYSERVER_H
+
+#include "DisplayServer.h"
+
+class QProcess;
+
+namespace SDDM {
+    class XorgUserDisplayServer : public DisplayServer {
+        Q_OBJECT
+        Q_DISABLE_COPY(XorgDisplayServer)
```
aleixpol: It doesn't build, this should be `Q_DISABLE_COPY(XorgUserDisplayServer)`

plfiorini: I know, that's why it's a draft :)
I pushed it so I can easily grab the source code and continue on another PC.

src/common/XAuth.h
```diff
@@ -0,0 +1,47 @@
...
+    XAuth();
+
+    QString authPath() const;
+    QString cookie() const;
+
+    void setup(bool testing);
+    bool addCookie(const QString &display, const QString &fileName);
+
+private:
+    bool m_setup = false;
```
plfiorini, line 47: ops, remove this

davidedmundson:
Let me know if this gets out of WIP, or if you hit something that you need a second set of eyes on.

plfiorini:
Things to do:

- [x] Save xauthority in the runtime directory not cache
- [x] Run DisplayCommand script, set cursor and DisplayStop command as user
- [x] Document that those scripts will be executed as sddm in sddm.conf man
- [x] Create a socket to pass the display from sddm-helper-x11 to sddm-helper
- [x] Set said display in XorgUserDisplayServer
- [x] Why do Xorg fail saying that only console users are allowed? From the code I see that this error is given when either stdin, stdout or stderr is not a character device. I just don't understand why S_ISCHR fails
- [x] Why Xorg user sessions do not work? i3 fails to connect to the display
- [x] After a while input devices are paused by logind and we cannot do anything in the greeter. This happens after I have tried to launch a user session.
- [x] Go through all the themes and clear the password field when the auth fails

aleixpol:
>  Why do Xorg fail saying that only console users are allowed? From the code I see that this error is given when either stdin, stdout or stderr is not a character device. I just don't understand why S_ISCHR fails

I did see this when redirecting the output, maybe it's the same reason?

marcthe12:
Xorg needs have stdin (from a simple test with startx). I suggest we need set stdin to be set to /dev/ttyN for the Xserver as workaround.

plfiorini:
Rebased on top of #1377

davidedmundson:
> Why Xorg user sessions do not work? i3 fails to connect to the display

Fixed in the PR that I just sent to your personal branch that this is based on.
We were creating the xauthority cookie before we worked out what display to use.

plfiorini:
Thanks guys for the help. This PR is now ready for your reviews.
At this point, it doesn't even matter to review #1377 since it's contained here.

src/common/Configuration.h
```diff
@@ -37,6 +37,9 @@ namespace SDDM {
         enum NumState { NUM_NONE, NUM_SET_ON, NUM_SET_OFF };
 
         //  Name                   Type         Default value                                   Description
+        // TODO: In SDDM 2.0 we want x11-user to be the default
```
Conan-Kudo: We don't even have SDDM 1.0 yet, what milestone is this TODO aimed for?

Conan-Kudo (approved):
Pretty much LGTM

src/common/Configuration.h
plfiorini: The plan was to merge this PR and #1379  and then release SDDM 1.0.
Do a few minor release and release SDDM 2.0.
For 2 reasons: it's time for SDDM 1.0 and I didn't feel like breaking the workflow now.

Conan-Kudo: I guess. I think you should switch it for SDDM 1.0, but otherwise it makes sense.

Conan-Kudo (approved):

aleixpol:
I quickly tested it and I don't seem to be able to log out from X11 sessions using this. Tried it with plasma, gnome and icewm.

Starting them seems to work fine. :+1:

src/helper-x11/xorguserhelper.cpp
```diff
@@ -0,0 +1,274 @@
...
+    if (!startServer())
+        return false;
+
+    // Setup display
+    startDisplayCommand();
+
+    // Start client process
+    if (!startClient())
+        return false;
+
```
davidedmundson: We need some sort of 

    connect(m_clientProcess, QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished),
            this, [](int, QProcess::ExitStatus) {
            QCoreApplication::instance()->quit();
    });
    
here
    
Even though there's one in XOrgUserHelper::startProcess, that's purely error handling and checks the return code.

When the main client exits, even if it's gracefully we still want to quit.

davidedmundson:
There's still a few things left to fix, there's the issue above which is breaking our logout, but even with that fixed my original greeter we go back to is in a broken state.

I'll try to follow up on that.

components/2.0/PictureBox.qml
```diff
@@ -106,6 +106,7 @@ FocusScope {
             Keys.onPressed: {
                 if (event.key === Qt.Key_Return || event.key === Qt.Key_Enter) {
                     container.login();
+                    password.text = "";
```
davidedmundson: Can we split this commit out

src/daemon/Greeter.cpp
```diff
@@ -261,6 +285,14 @@ namespace SDDM {
             qDebug() << "Greeter session failed to start";
     }
 
+    void Greeter::onDisplayServerReady(const QString &displayName)
```
davidedmundson, line 273: Can you talk me through this.

The greeter starts, the helper creates an X server, it gets a name, which we're sending back to the daemon.
Why does the daemon need to know this? For xauth cookie handling?

src/helper-x11/main.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+#include <QCoreApplication>
+
+#include "Constants.h"
+#include "SignalHandler.h"
+#include "xorguserhelper.h"
+
+using namespace SDDM;
+
+int main(int argc, char *argv[])
```
davidedmundson: I don't like this nested tree of helpers upon helpers.
It works, but it seems unnecessarily convoluted.

I don't see why the existing helper app can't start fork and start X wait, then fork and start the user session.

Super worst case we move utmp logging from the helper app to the daemon, we can make the helper drop privileges earlier in the main code rather than after the fork which might make things even cleaner.

src/daemon/Display.cpp
```diff
@@ -391,13 +433,14 @@ namespace SDDM {
...
-        // error happens (in this case we want to show the message from the
-        // greeter
-        if (status != Auth::HELPER_AUTH_ERROR)
+        // Restart the greeter and display server if sddm-helper exited
+        // with an internal error or a session error, otherwise we just
+        // want to reuse the existing so we jump to its VT
+        if (status == Auth::HELPER_SESSION_ERROR ||
+                status == Auth::HELPER_OTHER_ERROR)
             stop();
+        else
```
davidedmundson: If I've gone through the greeter, quit that, spawned a user session and then logged out gracefully, we need to restart the greeter.

This else statement prevents that, we just VT switch to a somewhat dead session.

src/helper-x11/xorguserhelper.cpp
aleixpol: https://github.com/plfiorini/sddm/pull/5

src/helper-x11/main.cpp
aleixpol: The fact that it works probably is enough? I'd say if we want to do a refactoring, we should first get the wayland one in as well, which has a similar problem and do it all together rather than holding the features on implementation details.

src/daemon/Display.cpp
aleixpol: With https://github.com/plfiorini/sddm/pull/5 I don't see the problem anymore. I'm not sure how to reproduce the problem you're seeing here.

src/daemon/Greeter.cpp
plfiorini: i just wanted to know the display name from the daemon just like before, in case it's useful in the future.

plfiorini: i guess that in the we can think about it in the future, if we really need it and aim for simplicity here

plfiorini: the only place where it seems it's used is in Display::displayServerStarted() when the socket server is started but we might use something else, maybe a random name.
SocketServer creates the socket like this:
```
QString socketName = QStringLiteral("sddm-%1-%2").arg(displayName).arg(generateName(6));
```
Since the socket name is passed around and not guessed we can really drop the displayName from the socket name and we are good to go since we already use a generated name

src/helper-x11/main.cpp
plfiorini: Do we need to land your utmp patch before? @davidedmundson

davidedmundson: I think they'll certainly make things easier and cleaner.  Means we only need to drop privileges once.

The utmp patches also require the display name to be sent to the daemon, so your addition above "for the future" is useful already! :)

src/daemon/Display.cpp
davidedmundson: I'm confused how it works then.

I log out, my user session helper stops, I switch to this VT, who restarts the greeter?

davidedmundson: AHA! I think I know why it works for you.

When we log in, the greeter stays running.
On the default theme when we go back to the greeter and try to log in again, this is fine.

On the breeze theme because we fade away and disable things. We do switch back to the greeter but it's in a broken state.

Is it the design that the greeter stay running? 

Also is m_terminalId correct here in the cases that the greeter helper spawns X?

src/helper-x11/main.cpp
davidedmundson: I made a rebase and adjusted it so the main sddm-helper ran XOrgUserHelper in process. 
https://github.com/davidedmundson/sddm/blob/user_xorg_after_utmp 

I have a few qDebugs and whatnot to remove, but the end result seems easier to follow.

src/daemon/Display.cpp
plfiorini: > Is it the design that the greeter stay running?

yes, one of the patches makes the greeter stay alive hence we just need to switch vt

davidedmundson: >yes, one of the patches makes the greeter stay alive hence we just need to switch vt

It's quite a bit of memory to waste keeping it and Xorg alive, just to make it slightly faster if someone does log out. I don't think it's a good trade off.

Also as observed it introduces a compatibility issue.

src/helper-x11/main.cpp
davidedmundson: >Do we need to land your utmp patch before? @davidedmundson

If it's going to prove pragmatically easier, we could merge this first, and then I can do those rounds of simplification on top before we have a release.

plfiorini: > I made a rebase and adjusted it so the main sddm-helper ran XOrgUserHelper in process.
> https://github.com/davidedmundson/sddm/blob/user_xorg_after_utmp
> 
> I have a few qDebugs and whatnot to remove, but the end result seems easier to follow.

404 error :(

src/daemon/Display.cpp
plfiorini: Now the user sessions is closed and we see a prompt and then the greeter is shown.
This is horrible!

Perhaps we can do like this:

1. helper tells daemon that the session is about to quit
2. daemon spawns the greeter and switch to its vt, then tells the helper about it
3. helper quits

But we shouldn't be worrying about this stuff in this PR.
This has nothing to do with running xorg with the sddm user.

src/helper-x11/main.cpp
plfiorini: I will merge #1383 and rebase on top of it, then move the code to run the display server from sddm-helper-x11 to sddm-helper

davidedmundson: Maybe https://github.com/davidedmundson/sddm/tree/user_xorg_after_utmp works? Though it will be slightly out of date now maybe?

We can do it before or after, whichever is easier.

aleixpol:
~~I have been testing it and it seems to work fine except that in some instances it seems to fail at switching to the user's xorg session tty. I'll see if I can provide a patch.~~

aleixpol:
> I have been testing it and it seems to work fine except that in some instances it seems to fail at switching to the user's xorg session tty. I'll see if I can provide a patch.

I'm pretty sure this was an issue with my installation and cannot reproduce anymore.

If I had the powers I'd approve the patch.

davidedmundson (approved):
I still want to do the helper refactor before release, but as stated in the in-line comment we can resolve that after merging this so we don't stall this forever.

davidedmundson:
It's worth noting that one of these commits is of super-importance even with the current Xorg daemon: "Allocate VT for the display"

We have an important downstream bug that if you start a wayland session then call switch user, SDDM will start on what SDDM thinks is the next available VT, which naturally is the one the wayland session is on. This fails and we can't switch users. 

This patch here fixes that, which is nice \o/

Vogtinator:
> It's worth noting that one of these commits is of super-importance even with the current Xorg daemon: "Allocate VT for the display"

Does this PR also help with #1281?

plfiorini:
Ready for review!

Conan-Kudo (approved):

aleixpol:
+1 just tested, worked as expected.

davidedmundson (approved):

AdamWill:
What happened to the commit "Make sure the helper process is stopped when shutting down." that was previously part of this PR?

edit: oh, I guess it was its own PR as https://github.com/sddm/sddm/pull/1377 , and was in this one accidentally?

plfiorini:
> What happened to the commit "Make sure the helper process is stopped when shutting down." that was previously part of this PR?
> 
> edit: oh, I guess it was its own PR as #1377 , and was in this one accidentally?

It was needed due to an error that is now solved

ThisNekoGuy:
How would a user run the Xorg server as rootless with SDDM?
This isn't done by default (I don't understand why, considering it's a potential security risk) and editing `Xwrapper.config` to set `needs_root_rights = no` doesn't seem to make a difference.

aleixpol:
You need to set `DisplayServer=x11-user` as the documentation says.

ThisNekoGuy:
> You need to set `DisplayServer=x11-user` as the documentation says.

I did put this in the Xwrapper.config file but this caused my system to hang on startup.
(I'm on Arch)

plfiorini:
> > You need to set `DisplayServer=x11-user` as the documentation says.
> 
> I did put this in the Xwrapper.config file but this caused my system to hang on startup.
> (I'm on Arch)

No, you have to put it in the SDDM configuration file.

ThisNekoGuy:
> > > You need to set `DisplayServer=x11-user` as the documentation says.
> > 
> > 
> > I did put this in the Xwrapper.config file but this caused my system to hang on startup.
> > (I'm on Arch)
> 
> No, you have to put it in the SDDM configuration file.

Ah, okay; sorry, I misunderstood where it was supposed to go.
Is this going to be enabled by default going forward?

plfiorini:
> > > > You need to set `DisplayServer=x11-user` as the documentation says.
> > > 
> > > 
> > > I did put this in the Xwrapper.config file but this caused my system to hang on startup.
> > > (I'm on Arch)
> > 
> > 
> > No, you have to put it in the SDDM configuration file.
> 
> Ah, okay; sorry, I misunderstood where it was supposed to go.
> Is this going to be enabled by default going forward?

Yes sure, but right now we can't make it the default because some people rely on the fact that it's running as root

Xwang1976:
The man files says "For x11-user you might need to configure Xorg.wrap(1)".
Is there any specific option which must be present in Xwrapper.config to have a rootless Xorg with sddm?
I suppose I have to put:
allowed_users = anybody
needs_root_rights = no

Is it correct?

plfiorini:
> The man files says "For x11-user you might need to configure Xorg.wrap(1)".

Tweak that file only if it's not working.

> Is there any specific option which must be present in Xwrapper.config to have a rootless Xorg with sddm?
> I suppose I have to put:
> allowed_users = anybody
> needs_root_rights = no
> 
> Is it correct?

Yes that is correct and should force X11 to run without root privileges, but you only have to configure this stuff if really needed.
I don't expect much people (or even no one) to need to tweak it.

Xwang1976:
I've set x11-user in the sddm conf file and now I have Xorg owned by my user. The same happens for sddm-helper. Sddm is running as root.
Is it correct?
Meantime I can confirm you that I did not have to create the Xwrapper.config file at all

Xwang1976:
I've noted that command launched with firejail are not able to connect to the X server. With GDM in rootless mode the same command works correctly.
It seems that GDM and SDDM call Xorg using different options.

Xwang1976:
When using sddm, echo $XAUTHORITY gives /run/user/1000/c04e4bb9-0531-4ab3-86a8-caaa6f30a661 with the file name changingat every reboot
Using GDM echo $XAUTHORITY gives /run/user/1000/gdm/Xauthority

Moreover when I look at the Xorg process it is runned by sddm as:
/usr/lib/Xorg -nolisten tcp -background none -seat seat0 -noreset -keeptty -novtswitch -verbose 3 -auth /run/user/1000/c04e4bb9-0531-4ab3-86a8-caaa6f30a661 -displayfd 8 vt1 -logfile /dev/null

GDM instead runs Xorg in this way:
/usr/lib/Xorg vt2 -displayfd 3 -auth run run/user/1000/gdm/Xauthority -nolisten tcp -background none -noreset -keeptty -novtswitch -verbose  3

Finally running xauth when started by sddm I get:
access control enabled, only authorized clients can connect

Whereas running when uning GDM I get:
access control enabled, only authorized clients can connect
SI:localuser:andreak

Xwang1976:
When using sddm, echo $XAUTHORITY gives /run/user/1000/c04e4bb9-0531-4ab3-86a8-caaa6f30a661 with the file name changingat every reboot
Using GDM echo $XAUTHORITY gives /run/user/1000/gdm/Xauthority

Moreover when I look at the Xorg process it is runned by sddm as:
/usr/lib/Xorg -nolisten tcp -background none -seat seat0 -noreset -keeptty -novtswitch -verbose 3 -auth /run/user/1000/c04e4bb9-0531-4ab3-86a8-caaa6f30a661 -displayfd 8 vt1 -logfile /dev/null

GDM instead runs Xorg in this way:
/usr/lib/Xorg vt2 -displayfd 3 -auth run run/user/1000/gdm/Xauthority -nolisten tcp -background none -noreset -keeptty -novtswitch -verbose  3

Finally running xauth when started by sddm I get:
access control enabled, only authorized clients can connect

Whereas running when using GDM I get:
access control enabled, only authorized clients can connect
SI:localuser:andreak

silverqx:
Does this mean we will have a rootless X in the v0.20 release? That would be 🔥

What will be needed? I have read the conversations above and it looks like a user will have to add the following to the `/etc/sddm.conf.d/rootless.conf` as it will not be a default:
```
[General]
DisplayServer=x11-user
```

Am I right or did I misunderstand something 😁? thx

Conan-Kudo:
That is how you do it, yes.

silverqx:
When do you plan to release 0.20? I'm really waiting for this.

silverqx:
Ok, I tried this and I have been using it for the last month or so, it really works as we described, just adding the `DisplayServer=x11-user` to the `sddm.conf` is enough and the `X` server will start `rootless`. 👌

Everything works fine for me but I have one big issue, when I log into the system, write the password down and press enter, then the `sddm` switches me to the `tty1` (ctrl+alt+F1) to the first console and __I have to manually__ switch to the `vt7` (alt+F7). If I don't do this it stays on the `tty1` for around 1min or so and then it switches to the `vt7`, it takes soo long and I'm not even sure if it's true because I didn't try this a long time. 🐛

I'm on Gentoo with OpenRC and the `X` server is started like this:
```bash
exec /usr/bin/X -nolisten tcp -keeptty "$@" "vt$XDG_VTNR"
```

And the `$XDG_VTNR` is set to the `7` number, I don't think changing it to the `1` would be a good idea.

This problem is not happening without the `DisplayServer=x11-user` so with the previous so-called `root` mode. It is only happening with this new `rootless` mode.

Is there any workaround for this or a patch? How can I force the sddm to switch to the `vt7` after login? @plfiorini

PR #1372, opened 2021-02-22
Minor optimization of shell scipts

ghost:
Would I bother you if I replaced every test -z with case statement and awk with tr for optimization?

### DIFF
--- a/releng/prepare-relnotes
+++ b/releng/prepare-relnotes
@@ -21,7 +21,7 @@
 export prev=$1
 export next=$2
 
-export contributors=`git shortlog v${prev}..develop -sn --no-merges | awk -F'\t' '{print $2}' | tr -s "\\n" "*" | sed -e 's@*@, @g' -e 's@, $@\n@g'`
+export contributors=`git shortlog v${prev}..develop -sn --no-merges | tr -d'\t' -f 2- | tr -s "\\n" "*" | sed -e 's@*@, @g' -e 's@, $@\n@g'`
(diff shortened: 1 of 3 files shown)


plfiorini:
Can you explain why is this necessary and what is the benefit?

releng/prepare-relnotes
```diff
@@ -21,7 +21,7 @@
 export prev=$1
 export next=$2
 
-export contributors=`git shortlog v${prev}..develop -sn --no-merges | awk -F'\t' '{print $2}' | tr -s "\\n" "*" | sed -e 's@*@, @g' -e 's@, $@\n@g'`
+export contributors=`git shortlog v${prev}..develop -sn --no-merges | cut -d'\t' -f 2 | tr -s "\\n" "*" | sed -e 's@*@, @g' -e 's@, $@\n@g'`
```
tpgxyz, line 24: That's ok for me, as cut call is lighter than awk :)

tpgxyz:
I may add 2 cents from my side:

- `case` compares one shell word to many shell patterns, i case of `-z` test, this make no sense to me
- `cut` is lighter than `awk` so here may be like 0,01% of speedup

releng/prepare-relnotes
plfiorini: This might be the only change that is worth having compared to the others, although the impact will not be noticeable because it's so small and this script is used a few times by the developers not by SDDM itself.

plfiorini (changes requested):
Those `case` changes don't make much sense to me

PR #1377, opened 2021-03-08
Stop the helper process

plfiorini:
Make sure the helper process is stopped when shutting down.


Vogtinator:
Does this make any observable difference? At least when sddm itself is shut down, the subreaper (most likely `init`) would inherit the helper processes and do the same to them already.

I'm also wondering whether this introduces behaviour similar to https://github.com/sddm/sddm/pull/1249 - that now children of the sessions are forcibly killed. In theory it shouldn't make a difference...

plfiorini:
When developing Xorg without root, I had restarted sddm and the display server was still around after that, making it impossible to start it again.   With this patch we make sure that we close nicely the helper process.

PR #1379, opened 2021-03-13
Greeter on Wayland

plfiorini:
Depends on #1371 and at the moment that PR is contained here too.
Once #1371 is merged, I will rebased this on `develop`.


plfiorini:
@aleixpol Check this out

src/helper-x11/xorguserhelper.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    return true;
+}
+
+bool XOrgUserHelper::startServer()
+{
+    // Create pipe for communicating with X server
+    // 0 == read from X, 1 == write to X
+    int pipeFds[2];
+    if (::pipe(pipeFds) != 0) {
```
zzag, line 136: The file descriptor for the read endpoint will be leaked to the X server process. Maybe set the O_CLOEXEC flag after creating the pipe?

```
fcntl(pipeFds[0], F_SETFD, FD_CLOEXEC);
```

src/helper-wayland/waylanduserhelper.cpp
```diff
@@ -0,0 +1,166 @@
...
+        if (!start())
+            QCoreApplication::exit(EXIT_FAILURE);
+    }
+}
+
+bool WaylandUserHelper::startProcess(const QString &cmd,
+                                     const QProcessEnvironment &env,
+                                     QProcess **p)
+{
+    auto args = QProcess::splitCommand(cmd);
```
zzag, line 90: Note that QProcess::splitCommand() was introduced in Qt 5.15, but it seems like SDDM requires Qt >= 5.8.

tpgxyz:
I've compiled sddm with this PR, updated it on my OpenMandriva Lx box, updated configs, and it works with wayland greeter :)
Using it for couple of hours, from time to time do relogin, logout, swtich user etc.

src/helper-wayland/waylanduserhelper.cpp
plfiorini: it might be the time to require a newer version

Mazino-Urek:
Will wayland support land before Plasma 5.22?

Conan-Kudo:
@plfiorini Can you rebase this PR now that #1371 has been merged?

PR #1382, opened 2021-03-17
Source files using `--login` for bash and zsh

oxalica:
Simplify codes and make `zsh` also source `/etc/profile`.
`/etc/profile` contains basic environment variables like `QT_IM_MODULE` on NixOS. Ignoring it will make input method unusable.


marcthe12:
there is Xsession script also which performs the same logic. I suggest to do the necessary changes to that script also

oxalica:
@marcthe12 Fixed.

davidedmundson (approved):
It does indeed seem simpler and appears to work the same

data/scripts/Xsession
```diff
@@ -8,26 +8,11 @@
 # Note that the respective logout scripts are not sourced.
```
tavianator, line 8: I think this won't be true any more

oxalica: We `exec` at the end of script, I confirmed logout scripts have no possibility to run when we  `exec`.

tavianator: Oh true, never mind!

PR #1383, opened 2021-03-18
Move utmp logic to Display

davidedmundson:
By moving this in display we don't need sddm-helper to perform anything
as root after the initial setup.

This allows us to drop the priveleges in UserSession before forking.

This should unblock #1230 as well as give us a path to simplify the
paths where the helper is running the display manager.


davidedmundson:
Tested with a combination of "who" and 

utmpdump /var/run/utmp

plfiorini (approved):

PR #1393, opened 2021-04-20
Greeter on Wayland - rebased

aleixpol:
Rebased version of https://github.com/sddm/sddm/pull/1379

Had to be creative in some places since it changed quite a bit.

### DIFF
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -48,6 +48,7 @@
 #include "Login1Manager.h"
 #include "Login1Session.h"
 #include "VirtualTerminal.h"
+#include "WaylandDisplayServer.h"
 
 #if defined(Q_OS_LINUX)
 #include <utmp.h>
@@ -70,6 +71,8 @@ namespace SDDM {
             m_displayServerType = X11DisplayServerType;
         else if (displayServerType == QStringLiteral("x11-user"))
             m_displayServerType = X11UserDisplayServerType;
+        else if (displayServerType == QStringLiteral("wayland"))
+            m_displayServerType = WaylandDisplayServerType;
         else {
             qWarning("\"%s\" is an invalid value for General.DisplayServer: fall back to \"x11\"",
                      qPrintable(displayServerType));
@@ -84,6 +87,10 @@ namespace SDDM {
         case X11UserDisplayServerType:
             m_displayServer = new XorgUserDisplayServer(this);
             m_greeter->setDisplayServerCommand(XorgUserDisplayServer::command(this));
+            break;
+        case WaylandDisplayServerType:
+            m_displayServer = new WaylandDisplayServer(this);
+            m_greeter->setDisplayServerCommand(mainConfig.Wayland.CompositorCommand.get());
             break;
         }
 
--- a/src/helper/UserSession.cpp
+++ b/src/helper/UserSession.cpp
@@ -27,6 +27,7 @@
 #include "VirtualTerminal.h"
 #include "XAuth.h"
 #include "xorguserhelper.h"
+#include "waylandhelper.h"
 
 #include <sys/types.h>
 #include <sys/ioctl.h>
@@ -45,7 +46,8 @@ namespace SDDM {
         , m_process(new QProcess(this))
         , m_xorgUser(new XOrgUserHelper(this))
     {
-        connect(m_process, QOverload<int>::of(&QProcess::finished), this, &UserSession::finished);
+        m_process->setProcessChannelMode(QProcess::ForwardedChannels);
+        connect(m_process, QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished), this, &UserSession::finished);
         connect(m_xorgUser, &XOrgUserHelper::displayChanged, this, [this, parent](const QString &display) {
             auto env = processEnvironment();
             env.insert(QStringLiteral("DISPLAY"), m_xorgUser->display());
@@ -57,16 +59,25 @@ namespace SDDM {
     }
 
     bool UserSession::start() {
-        QProcessEnvironment env = qobject_cast<HelperApp*>(parent())->session()->processEnvironment();
+        auto helper = qobject_cast<HelperApp*>(parent());
+        QProcessEnvironment env = helper->session()->processEnvironment();
 
         setup();
 
         if (!m_displayServerCmd.isEmpty()) {
-            m_xorgUser->setEnvironment(env);
-            if (!m_xorgUser->start(m_displayServerCmd))
-                return false;
+            if (env.value(QStringLiteral("XDG_SESSION_TYPE")) == QLatin1String("wayland") && env.value(QStringLiteral("XDG_SESSION_CLASS")) == QLatin1String("greeter")) {
+                m_wayland = new WaylandHelper(this);
+                m_wayland->setEnvironment(env);
+                if (!m_wayland->startCompositor(m_displayServerCmd))
+                    return false;
+            } else {
+                m_xorgUser->setEnvironment(env);
+                if (!m_xorgUser->start(m_displayServerCmd))
+                    return false;
+            }
         }
 
+        bool isWaylandGreeter = false;
         if (env.value(QStringLiteral("XDG_SESSION_TYPE")) == QLatin1String("x11")) {
             if (env.value(QStringLiteral("XDG_SESSION_CLASS")) == QLatin1String("greeter")) {
                 qInfo() << "Starting X11 greeter session:" << m_path;
@@ -79,11 +90,19 @@ namespace SDDM {
                 m_process->start(mainConfig.X11.SessionCommand.get(), QStringList{m_path});
             }
         } else if (env.value(QStringLiteral("XDG_SESSION_TYPE")) == QLatin1String("wayland")) {
-            const QString cmd = QStringLiteral("%1 %2").arg(mainConfig.Wayland.SessionCommand.get()).arg(m_path);
-            qInfo() << "Starting Wayland user session:" << cmd;
-            m_process->start(mainConfig.Wayland.SessionCommand.get(), QStringList{m_path});
-            m_process->closeWriteChannel();
-            m_process->closeReadChannel(QProcess::StandardOutput);
+            if (env.value(QStringLiteral("XDG_SESSION_CLASS")) == QLatin1String("greeter")) {
+                isWaylandGreeter = true;
+                auto args = QProcess::splitCommand(m_path);
+                m_process->setProgram(args.takeFirst());
+                m_process->setArguments(args);
+                m_wayland->startGreeter(m_process);
+            } else {
+                const QString cmd = QStringLiteral("%1 %2").arg(mainConfig.Wayland.SessionCommand.get()).arg(m_path);
+                qInfo() << "Starting Wayland user session:" << cmd;
+                m_process->start(mainConfig.Wayland.SessionCommand.get(), QStringList{m_path});
+                m_process->closeWriteChannel();
+                m_process->closeReadChannel(QProcess::StandardOutput);
+            }
         } else {
             qCritical() << "Unable to run user session: unknown session type";
         }
@@ -91,6 +110,9 @@ namespace SDDM {
         if (m_process->waitForStarted()) {
             int vtNumber = processEnvironment().value(QStringLiteral("XDG_VTNR")).toInt();
             VirtualTerminal::jumpToVt(vtNumber, true);
+            return true;
+        } else if (isWaylandGreeter) {
+            // This is probably fine, we need the compositor to start first
             return true;
         }
 
@@ -103,8 +125,10 @@ namespace SDDM {
         if (!m_process->waitForFinished(5000))
             m_process->kill();
 
-        if (!m_displayServerCmd.isEmpty())
+        if (!m_displayServerCmd.isEmpty()) {
             m_xorgUser->stop();
+            m_wayland->stop();
+        }
     }
 
     QProcessEnvironment UserSession::processEnvironment() const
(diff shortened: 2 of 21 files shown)


src/CMakeLists.txt
```diff
@@ -3,3 +3,4 @@ configure_file("common/Constants.h.in" "common/Constants.h" IMMEDIATE @ONLY)
 add_subdirectory(daemon)
 add_subdirectory(greeter)
 add_subdirectory(helper)
+#add_subdirectory(helper-wayland)
```
zzag: there's no such a directory, remove it?

components/2.0/PictureBox.qml
```diff
@@ -106,6 +106,7 @@ FocusScope {
             Keys.onPressed: {
                 if (event.key === Qt.Key_Return || event.key === Qt.Key_Enter) {
                     container.login();
+                    password.text = "";
```
plfiorini: this is not needed

src/daemon/Display.cpp
```diff
@@ -455,6 +462,15 @@ namespace SDDM {
...
             m_greeter->start();
+
+        // Restart the greeter and display server if sddm-helper exited
+        // with an internal error or a session error, otherwise we just
+        // want to reuse the existing so we jump to its VT
+        if (status == Auth::HELPER_SESSION_ERROR ||
+                status == Auth::HELPER_OTHER_ERROR)
+            stop();
+        else
+            VirtualTerminal::jumpToVt(m_terminalId, true);
```
plfiorini: we decided not to stop the greeter

src/greeter/GreeterApp.cpp
```diff
@@ -167,9 +167,6 @@ namespace SDDM {
 
         view->engine()->addImportPath(QStringLiteral(IMPORTS_INSTALL_DIR));
 
-        // connect proxy signals
-        connect(m_proxy, &GreeterProxy::loginSucceeded, view, &QQuickView::close);
-
         // we used to have only one window as big as the virtual desktop,
```
plfiorini: this should not be removed

src/helper/UserSession.cpp
```diff
@@ -57,16 +59,25 @@ namespace SDDM {
...
         if (!m_displayServerCmd.isEmpty()) {
-            m_xorgUser->setEnvironment(env);
-            if (!m_xorgUser->start(m_displayServerCmd))
-                return false;
+            if (env.value(QStringLiteral("XDG_SESSION_TYPE")) == QLatin1String("wayland") && env.value(QStringLiteral("XDG_SESSION_CLASS")) == QLatin1String("greeter")) {
+                m_wayland = new WaylandHelper(this);
+                m_wayland->setEnvironment(env);
+                if (!m_wayland->startCompositor(m_displayServerCmd))
+                    return false;
+            } else {
```
plfiorini, line 73: else if XDG_SESSION_TYPE == x11

src/helper/UserSession.cpp
```diff
@@ -224,6 +248,7 @@ namespace SDDM {
             qCritical() << "setgid(" << pw.pw_gid << ") failed for user: " << username;
             exit(Auth::HELPER_OTHER_ERROR);
         }
+        qputenv("XDG_RUNTIME_DIR", QByteArrayLiteral("/run/user/") + QByteArray::number(pw.pw_uid));
```
plfiorini, line 253: why?

src/helper/waylandhelper.cpp
```diff
@@ -0,0 +1,120 @@
...
+    process->setInputChannelMode(QProcess::ForwardedInputChannel);
+    process->setProcessChannelMode(QProcess::ForwardedChannels);
+    process->setProcessEnvironment(m_environment);
+
+    qDebug() << "Starting Wayland process" << cmd << m_environment.value(QStringLiteral("USER"));
+    connect(process, QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished),
+            process, [](int exitCode, QProcess::ExitStatus exitStatus) {
+        qDebug() << "wayland compositor finished" << exitCode << exitStatus;
+        if (exitCode != 0 || exitStatus != QProcess::NormalExit)
+            QCoreApplication::instance()->quit();
```
plfiorini, line 82: this will quit the helper, we don't want this

src/helper/waylandhelper.cpp
```diff
@@ -0,0 +1,120 @@
...
+        *p = process;
+
+    qDebug() << "started succesfully" << cmd;
+    return true;
+}
+
+void WaylandHelper::switchVt()
+{
+    int vtNumber = m_environment.value(QStringLiteral("XDG_VTNR")).toInt();
+    VirtualTerminal::jumpToVt(vtNumber, true);
```
plfiorini, line 105: i think UserSession already do that

plfiorini (changes requested):
the branch contains commits that we don't want anymore.
i would have applied the patch on top of the new `develop` instead of rebasing it.

components/2.0/PictureBox.qml
aleixpol: yep, rebased 1 commit too much.

src/greeter/GreeterApp.cpp
aleixpol: yep, rebased 1 commit too much.

src/helper/waylandhelper.cpp
aleixpol: I had to bring back the WaylandSocketWatcher because we need to wait for the compositor to start before we can start the sddm-greeter.

An alternative implementation would be to have the sddm-greeter block at start until the compositor appears.

src/helper/waylandhelper.cpp
aleixpol: This code is copied from `XOrgUserHelper::startProcess`.

I guess we'll need to include a start failed signal.

Ultimately we'll need quitting anyway. If the compositor (wayland or xorg) can't be started.

src/helper/UserSession.cpp
aleixpol: For WaylandSocketWatcher to work.

aleixpol:
Ping?

aleixpol:
Ping?

Mazino-Urek:
It would have really great if this feature landed before Plasma 5.23.

plfiorini:
Added a couple of environment variables for fullscreen-shell and disable window decoration, then rebased on top of the latest develop.  I will give it one more last test later this evening and merge if it works fine.

plfiorini (approved):

aleixpol:
Thanks!

isopix:
Waylanded SDDM really sucks on mulit-monitor setups, because it’s not cloned across displays, and menu items appeared on internal(closed laptop lid) display, than I’ve been clicking (so it was impossible to change Session for me). It uses Weston by default, and from my research Weston doesn’t support clone-mode-by-default (ass application argument).
Maybe other Wayland compositor(than weston) could do that?

aleixpol:
> Maybe other Wayland compositor(than weston) could do that?

Yes, that's what the CompositorCommand setting is for.

isopix:
I know. But for now sddm with wayland is downgrade to sddm on X11 on
multimonitor setups.

I wanted to point that IMHO SDDM should choose other compositor for
wayland as default.

I've simply went back to X11 setup. But SDDM should definitely work
with multi-monitor setups. And while strict multimonitor setups are in
decline, due to widescreens larger displays becoming cheaper, every
laptop with external monitor acts as multi-monitor setup (because
while I've set internal display as disabled, I must go trough SDDM for
it).

Also SDDM should work out-of-the-box on as many configurations as
possible, and mine laptop+ext. screen is not so exotic. So I think the
current state of SDDM+weston mariage (in wayland mode) is not
acceptable (I know that X11 is default, but I guess that with current
pro-wyaland trend, you want to change that as quickly as possible)

cheers,
Krzysztof Krakowia

ppaalanen:
Was there a discussion somewhere on how you ended up choosing to use Weston's fullscreen-shell protocol extension?

I see that you are planning to run on more than Weston, which I think is a good idea, and I suppose it also means the greeter works on the xdg-shell family of protocol extensions, not just fullscreen-shell.

As for the clone mode, you could have the greeter set up similar windows on all outputs. That does not need clone mode support from Weston, and it will fit diverse monitors much *much* better than a clone mode because a) resolution aware text rendering instead of raster scaling, and b) you can fit the UI layout to the available screen estate on each monitor.

At Weston upstream, we had kind of abandoned the fullscreen-shell, but now that it actually has a use case, I wonder what to do. Maybe we should open a Weston issue about the future of the fullscreen-shell?

Weston has also kiosk shell plugin which seems more promising going forward and exposes the xdg-shell family of protocol extensions.

plfiorini:
> Was there a discussion somewhere on how you ended up choosing to use Weston's fullscreen-shell protocol extension?

No, I just wanted something that didn't involve a panel but it's temporary.

> I see that you are planning to run on more than Weston, which I think is a good idea, and I suppose it also means the greeter works on the xdg-shell family of protocol extensions, not just fullscreen-shell.
>

The greeter should work very well with xdg-shell.

> As for the clone mode, you could have the greeter set up similar windows on all outputs. That does not need clone mode support from Weston, and it will fit diverse monitors much _much_ better than a clone mode because a) resolution aware text rendering instead of raster scaling, and b) you can fit the UI layout to the available screen estate on each monitor.
> 

I need to investigate what's going on here: SDDM creates a window for each screen that is reported by Qt therefore there is no clone mode here as opposed to X11.  This should actually be very good for the use case of @isopix but my guess is that Weston still reports the laptop output even though the lid is closed and that's why the greeter goes into that output.

> At Weston upstream, we had kind of abandoned the fullscreen-shell, but now that it actually has a use case, I wonder what to do. Maybe we should open a Weston issue about the future of the fullscreen-shell?
> 
> Weston has also kiosk shell plugin which seems more promising going forward and exposes the xdg-shell family of protocol extensions.

At the time there was no kiosk shell, so the fullscreen-shell was the best way to get something without a panel.
I'm on Fedora 33 with Weston 8.0.0 and this shell is not available.
Using the kiosk shell plugin is the best going forward and I'm considering to switch to it now, before the SDDM release.

davidedmundson:
>The greeter should work very well with xdg-shell.

Released Qt presents a challenge QWindow::setFullscreen() on XDG calls xdg_toplevel.set_fullscreen(null). So on a multi-screen setup you don't know where the window will end up. 

Qt's fullscreen-shell did forward the screen.

XDGShell implementation is changed in Qt6 and in KDE's Qt backports.

plfiorini:
> > The greeter should work very well with xdg-shell.
> 
> Released Qt presents a challenge QWindow::setFullscreen() on XDG calls xdg_toplevel.set_fullscreen(null). So on a multi-screen setup you don't know where the window will end up.

Oh I forgot about that

> Qt's fullscreen-shell did forward the screen.

Yeah I remember I did get that right when I wrote the shell integration plugin :)

> 
> XDGShell implementation is changed in Qt6 and in KDE's Qt backports.

Do you know if distros are using the KDE's Qt backports?

ppaalanen:
Thanks!

I would assume that Weston does not handle the laptop lid switch at all, so probably the internal panel does not even turn off, unless it also appears disconnected from KMS perspective.

darkbasic:
> Do you know if distros are using the KDE's Qt backports?

I think Arch does: https://archlinux.org/packages/extra/x86_64/qt5-base/

isopix:
On /29/21, Pier Luigi Fiorini ***@***.***> wrote:
> I need to investigate what's going on here: SDDM creates a window for each
> screen that is reported by Qt therefore there is no clone mode here as
> opposed to X11.  This should actually be very good for the use case of
> @isopix but my guess is that Weston still reports the laptop output even
> though the lid is closed and that's why the greeter goes into that output.

I havem't investigated it rurther, nor I've opened laptop lid. Simply
decided this setup is broken, switched to X11 backend and posted there

SSDM started in wayland mode on external screen, but dialogs (like
session/language/keyboard selection) haven't appeared on this display
(not sure about internal one). I was able to login, but not to use
stuff that required 'menus' (that starts as external windows/dialogs)

src/helper/UserSession.cpp
Vogtinator: This looks not great. Either something like `pam_systemd` creates it and it can be used or sddm has to create its own and use that path. Just guessing that it will be `/run/user/<uid>` smells like a hack.

davidedmundson: pam_systemd does create it

Vogtinator: Exactly, but then it should only be used in the context that `$XDG_RUNTIME_DIR` is actually set in

Vogtinator: In FreeBSD for instance (or with ConsoleKit in general, not sure), `XDG_RUNTIME_DIR` is actually set up by the session scripts and somewhere below `/tmp` IIRC.

jbeich: On FreeBSD both `consolekit2` ([upstream](https://github.com/ConsoleKit2/ConsoleKit2/blob/1.2.4/configure.ac#L160-L167) + [downstream](https://github.com/freebsd/freebsd-ports/blob/594ba3e582e7/sysutils/consolekit2/Makefile#L53)) and `pam_xdg` ([downstream](https://github.com/freebsd/freebsd-ports/blob/594ba3e582e7/sysutils/pam_xdg/Makefile#L23)) use `/var/run/user/<uid>`, hardcoded at build-time.

jbeich: I mean, `XDG_RUNTIME_DIR` set to `/tmp` (as opposed to fallback when unset) shouldn't be used outside of debugging as [basedir-spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) says "The directory MUST be owned by the user, and he MUST be the only one having read and write access to it. Its Unix access mode MUST be 0700."

ppaalanen: Right. If something else (e.g. pam_systemd) created XDG_RUNTIME_DIR, then you must READ the environment variable XDG_RUNTIME_DIR to find out its value. Or maybe there is a d-bus interface to ask for it if this is not your current session, I don't know. It would be incorrect to guess what the value might be and set XDG_RUNTIME_DIR to that guessed value.

PR #1396, opened 2021-04-23
Fix user session logs file saving

davidedmundson:
Logs were set up in the former setupChildProcess doing manual management
of file descriptors. This was needed as we wanted to create the error
file as a regular user.

In the refactor this redirection broke, but now that sddm-helper was
changed to drop priveledges earlier, we can now use QProcess methods
directly.

### DIFF
--- a/src/helper/UserSession.cpp
+++ b/src/helper/UserSession.cpp
@@ -82,6 +82,8 @@ namespace SDDM {
             const QString cmd = QStringLiteral("%1 %2").arg(mainConfig.Wayland.SessionCommand.get()).arg(m_path);
             qInfo() << "Starting Wayland user session:" << cmd;
             m_process->start(mainConfig.Wayland.SessionCommand.get(), QStringList{m_path});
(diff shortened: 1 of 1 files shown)


davidedmundson:
Caused by my patch, so apologies for that

src/helper/UserSession.cpp
```diff
@@ -82,6 +82,8 @@ namespace SDDM {
             const QString cmd = QStringLiteral("%1 %2").arg(mainConfig.Wayland.SessionCommand.get()).arg(m_path);
             qInfo() << "Starting Wayland user session:" << cmd;
             m_process->start(mainConfig.Wayland.SessionCommand.get(), QStringList{m_path});
+            m_process->closeWriteChannel();
+            m_process->closeReadChannel(QProcess::StandardOutput);
```
aleixpol, line 86: Maybe it would make sense to only close the standard output if it's not a greeter? For greeters I find it's useful to see what's going on.

plfiorini (approved):

PR #1408, opened 2021-05-18
sddm service is a part of graphical target

tpgxyz:
This allow to stop or restart sddm.service when graphical.target revices one of these signals.

### DIFF
--- a/services/sddm.service.in
+++ b/services/sddm.service.in
@@ -3,6 +3,7 @@ Description=Simple Desktop Display Manager
 Documentation=man:sddm(1) man:sddm.conf(5)
 Conflicts=getty@tty1.service
 After=systemd-user-sessions.service getty@tty1.service plymouth-quit.service systemd-logind.service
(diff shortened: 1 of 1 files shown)


davidedmundson:
Intuitively I would agree it makes sense, my concern is that neither GDM nor LightDM do this and in general we want to be somewhat consistent.

I'll ping a GDM author.

davidedmundson:
>I'll ping a GDM author.

"It makes sense, but probably also simply doesn't make a difference in practice"

Which I would also agree with, it's an odd thing to stop, but shouldn't do any harm in the general case

davidedmundson (approved):

tpgxyz:
Guess other DM's are not aware of `PartOf`.
Anyways this introduces some consistency, so if I switch target, I'll not be left with any running services from my previous target.

tpgxyz:
@davidedmundson @plfiorini Can you guys please merge this ?

Issue #1414, opened 2021-06-09
SDDM fails to start X11 session in Ubuntu hirsute 21.04


guss77:
When using the latest (0.19, or develop) SDDM with Ubuntu hirsute, the X11 session fails to start - the X server does not start and the greeter immediately shows again.

Looking at the user's journal, I see these errors:

```
-- Journal begins at Sat 2021-05-29 13:03:25 IDT. --
Jun 09 22:24:08 vesho sddm-helper[4972]: QDir::mkpath: Empty or null file name
Jun 09 22:24:08 vesho sddm-helper[4972]: Adding cookie to ""
Jun 09 22:24:08 vesho sddm-helper[4972]: QFSFileEngine::open: No file name specified
Jun 09 22:24:08 vesho sddm[4979]: /usr/bin/xauth:  file -q does not exist
Jun 09 22:24:08 vesho sddm[4979]: /usr/bin/xauth: (stdin):1:  bad "remove" command line
Jun 09 22:24:08 vesho sddm[4979]: /usr/bin/xauth: (stdin):2:  bad "add" command line
Jun 09 22:24:08 vesho sddm-helper[4972]: Xauthority path: "/run/user/1000/7f748bc4-5dfb-4f97-8f72-1bc5b341506e"
Jun 09 22:24:08 vesho sddm-helper[4972]: Running server: /usr/bin/X -nolisten tcp -background none -seat seat0 -noreset -keeptty -novtswitch -verbose 3 -auth /run/user/1000/7f748bc4-5dfb-4f97-8f72-1bc5b341506e -displayfd 8 vt3 -logfile /dev/null
Jun 09 22:24:08 vesho sddm[4980]: (EE)
Jun 09 22:24:08 vesho sddm[4980]: Fatal server error:
Jun 09 22:24:08 vesho sddm[4980]: (EE)
Jun 09 22:24:08 vesho sddm[4980]: Invalid argument -logfile with elevated privileges
Jun 09 22:24:08 vesho sddm[4980]: (EE)
Jun 09 22:24:08 vesho sddm[4980]: (EE)
Jun 09 22:24:08 vesho sddm[4980]: Please consult the The X.Org Foundation support
Jun 09 22:24:08 vesho sddm[4980]:          at http://wiki.x.org
Jun 09 22:24:08 vesho sddm[4980]:  for help.
Jun 09 22:24:08 vesho sddm[4980]: (EE)
[...]

PR #1415, opened 2021-06-10
remove `-logfile` arg that causes server to fail

guss77:
See https://github.com/sddm/sddm/issues/1414 and https://linux.die.net/man/1/xorg. And the comment on that line was also wrong.

Fixes issue #1414


plfiorini (approved):

plfiorini:
BTW the comment was not wrong that line caused the xserver to log to stdout which was then captured by sddm-helper

guss77:
> BTW the comment was not wrong that line caused the xserver to log to stdout which was then captured by sddm-helper

Specifying the log file as `/dev/null` causes logs to show in stdout? I wouldn't have expected that - but this may be the motif for X? :shrug: 'nyway - thanks for the merge!

PR #1422, opened 2021-06-24
Bump to Qt 5.15, port away from deprecated APIs

aleixpol:


plfiorini (changes requested):
please update README.md as well

plfiorini (approved):

PR #1427, opened 2021-07-06
Improve error handling

aleixpol:
Now that we are not running sessions with almighty processes they're bound to fail.

This patch tries to increase the visibility of these problems and offers some mechanisms so that it doesn't leave the system stuck.

### DIFF
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -116,6 +116,11 @@ namespace SDDM {
         // connect login result signals
         connect(this, SIGNAL(loginFailed(QLocalSocket*)), m_socketServer, SLOT(loginFailed(QLocalSocket*)));
         connect(this, SIGNAL(loginSucceeded(QLocalSocket*)), m_socketServer, SLOT(loginSucceeded(QLocalSocket*)));
+
+        connect(m_greeter, &Greeter::failed,
+                QCoreApplication::instance(), [] {
+                    QCoreApplication::instance()->exit(23);
+                });
     }
 
     Display::~Display() {
--- a/src/helper/UserSession.cpp
+++ b/src/helper/UserSession.cpp
@@ -68,12 +68,19 @@ namespace SDDM {
             if (env.value(QStringLiteral("XDG_SESSION_TYPE")) == QLatin1String("wayland") && env.value(QStringLiteral("XDG_SESSION_CLASS")) == QLatin1String("greeter")) {
                 m_wayland = new WaylandHelper(this);
                 m_wayland->setEnvironment(env);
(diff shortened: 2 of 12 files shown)


src/daemon/Display.cpp
```diff
@@ -116,6 +116,9 @@ namespace SDDM {
         // connect login result signals
         connect(this, SIGNAL(loginFailed(QLocalSocket*)), m_socketServer, SLOT(loginFailed(QLocalSocket*)));
         connect(this, SIGNAL(loginSucceeded(QLocalSocket*)), m_socketServer, SLOT(loginSucceeded(QLocalSocket*)));
+
+        connect(m_greeter, &Greeter::failed,
+                QCoreApplication::instance(), &QCoreApplication::quit);
```
davidedmundson: Do we want to exit with a return code so that systemd will report us as failed?

src/helper/UserSession.cpp
```diff
@@ -68,12 +68,19 @@ namespace SDDM {
             if (env.value(QStringLiteral("XDG_SESSION_TYPE")) == QLatin1String("wayland") && env.value(QStringLiteral("XDG_SESSION_CLASS")) == QLatin1String("greeter")) {
                 m_wayland = new WaylandHelper(this);
                 m_wayland->setEnvironment(env);
-                if (!m_wayland->startCompositor(m_displayServerCmd))
+                connect(m_wayland, &WaylandHelper::failed, this, &UserSession::stop);
```
davidedmundson, line 71: Why do we use a different path for late failure compared to failing to start?

davidedmundson: But you're right that no-one seems to be calling stop()

Should this be called from: 
Backend::closeSession

?

aleixpol: > Why do we use a different path for late failure compared to failing to start?

Which path are you referring to? ::setup calls exit() upon error

> Should this be called from `Backend::closeSession`?

This will have closeSession be called because `connect(m_session, &UserSession::finished, this, &HelperApp::sessionFinished);`

davidedmundson: I was referring to:
```
           if (!m_wayland->startCompositor(m_displayServerCmd)) {
                    Q_EMIT finished(1);
```

and asking why this isn't

                connect(m_wayland, &WaylandHelper::failed, this,[this]() {
                  Q_EMIT finished(1);
                });

aleixpol: Addressed by having stop() emit finished.

I don't think it makes sense to stop when we failed to start.

davidedmundson (approved):

PR #1431, opened 2021-07-30
Drop unused code

davidedmundson:
This is done in the helper as it's shared with rootless X


aleixpol:
+1 

It seems like you need to rebase?

plfiorini (approved):

PR #1432, opened 2021-08-05
Several VT switch fixes

davidedmundson:
Avoid duplicated VT Switch in UserSession

----

    Avoid incorrect terminal switching
    
    When running with rootless Xorg or wayland we setup a new VT
    when we are about to launch the session.
    
    This avoids an incorrect switch and SDDM taking ownership of a VT we
    don't use.

---

    Stop any active session when closing the daemon, this just helps cleanup when running sudo systemctl restart sddm

### DIFF
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -253,6 +253,8 @@ namespace SDDM {
         // stop the greeter
         m_greeter->stop();
 
+	m_auth->stop();
+
         // stop socket server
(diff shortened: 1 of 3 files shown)


davidedmundson:
Edit, don't merge as-is, I need to investigate a potential regression

davidedmundson:
> Edit, don't merge as-is, I need to investigate a potential regression

Was a misnomer. All good

PR #1437, opened 2021-08-16
Avoid creating a new TTY when using rootful X starting an X session

davidedmundson:
This is problematic as even though we stay on the TTY we tell logind a
different session is in use. This leads to the session being considered
inactive which causes lots of issues.


aleixpol:
+1

aleixpol (approved):

PR #1445, opened 2021-09-13
Avoid starting a new session on exit

davidedmundson:
Stopping SDDM will tear down the display.
This calls m->auth() stop, which stops the helper (any active session)

A newly introduced waitForFinished() means we can emit signals that the
process has finished.

We don't want Display's handling of the session ending to be active as
we can end up starting a greeter during shutdown.


aleixpol (approved):
Makes sense, thanks!

plfiorini (approved):

PR #1449, opened 2021-09-29
Address privilege dropping

aleixpol:
Our sddm-helper needs to stay a root process as it's in charge of dealing with PAM.  Alternatively the pam calls in PamBackend will fail.

To do so, we get helper processes to run xorg+greeter for x11-user and wayland-compositor+greeter for wayland. Obviously for general x11 don't need a helper as it runs as root.

### DIFF
--- a/src/helper/HelperApp.cpp
+++ b/src/helper/HelperApp.cpp
@@ -144,6 +144,7 @@ namespace SDDM {
             return;
         }
 
+        Q_ASSERT(getuid() == 0);
         if (!m_backend->authenticate()) {
             authenticated(QString());
 
@@ -199,6 +200,8 @@ namespace SDDM {
     }
 
     void HelperApp::sessionFinished(int status) {
+        Q_ASSERT(getuid() == 0);
+
         m_backend->closeSession();
 
         // write logout to utmp/wtmp
--- a/src/helper/HelperApp.cpp
+++ b/src/helper/HelperApp.cpp
@@ -202,7 +202,6 @@ namespace SDDM {
             const QString vt = env.value(QStringLiteral("XDG_VTNR"));
             if (env.value(QStringLiteral("XDG_SESSION_CLASS")) != QLatin1String("greeter")) {
                 // cache pid for session end
-                m_session->setCachedProcessId(m_session->processId());
                 utmpLogin(vt, displayId, m_user, m_session->processId(), true);
             }
         }
--- a/src/helper/HelperApp.cpp
+++ b/src/helper/HelperApp.cpp
@@ -302,6 +302,7 @@ namespace SDDM {
     HelperApp::~HelperApp() {
         Q_ASSERT(getuid() == 0);
 
+        m_session->stop();
         m_backend->closeSession();
 
         // write logout to utmp/wtmp
--- a/src/helper/HelperApp.cpp
+++ b/src/helper/HelperApp.cpp
@@ -67,7 +67,7 @@ namespace SDDM {
     }
 
     void HelperApp::setUp() {
-        QStringList args = QCoreApplication::arguments();
+        const QStringList args = QCoreApplication::arguments();
         QString server;
         int pos;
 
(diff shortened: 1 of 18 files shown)


davidedmundson:
It's a definite major regression that needs fixing before release. 
My patch to drop priveleges early is flawed, I found a few modules that need pam_close_session to be as root, and it's in the documentation.

Direction wise, I don't think we have a choice, so ++

Review wise, I think we can squash a lot of these commits.

src/helper/HelperApp.cpp
```diff
@@ -156,21 +189,29 @@ namespace SDDM {
...
+            sessionOpened(true);
+
+            // write successful login to utmp/wtmp
+            QProcessEnvironment env = m_session->processEnvironment();
+            QString displayId = env.value(QStringLiteral("DISPLAY"));
+            QString vt = env.value(QStringLiteral("XDG_VTNR"));
+            if (env.value(QStringLiteral("XDG_SESSION_CLASS")) != QLatin1String("greeter")) {
+                // cache pid for session end
+                m_session->setCachedProcessId(m_session->processId());
+                utmpLogin(vt, displayId, m_user, m_session->processId(), true);
```
davidedmundson, line 205: why do we call m_session->setCachedProcessId() from here rather than in
`Session::start()`

Then it's all encapsulated.

src/helper/HelperApp.cpp
```diff
@@ -156,21 +189,29 @@ namespace SDDM {
...
+                sessionOpened(false);
                 exit(Auth::HELPER_SESSION_ERROR);
                 return;
             }
 
-            sessionOpened(true, m_session->processId());
+            sessionOpened(true);
+
+            // write successful login to utmp/wtmp
+            QProcessEnvironment env = m_session->processEnvironment();
```
davidedmundson: const 

on most these args

src/helper/xorguserhelper.cpp
```diff
@@ -95,10 +105,14 @@ bool XOrgUserHelper::startProcess(const QString &cmd,
     // Make sure to forward the input of this process into the Xorg
     // server, otherwise it will complain that only console users are allowed
     auto *process = new QProcess(this);
-    process->setInputChannelMode(QProcess::ForwardedInputChannel);
-    process->setProcessChannelMode(QProcess::ForwardedChannels);
     process->setProcessEnvironment(env);
-
+    process->setInputChannelMode(QProcess::ForwardedInputChannel);
```
davidedmundson, line 109: why do we do this?

src/helper/HelperStartWayland.cpp
```diff
@@ -0,0 +1,59 @@
...
+}
+
+int main(int argc, char** argv)
+{
+    qInstallMessageHandler(WaylandHelperMessageHandler);
+    Q_ASSERT(::getuid() != 0);
+    QCoreApplication app(argc, argv);
+    if (argc != 3) {
+        QTextStream(stderr) << "Wrong number of arguments\n";
+        return 33;
```
davidedmundson: I assume this was just to help with debugging.

I think we can go  to just return  -1

src/helper/HelperStartWayland.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    using namespace SDDM;
+    WaylandHelper helper;
+    QObject::connect(&app, &QCoreApplication::aboutToQuit, &helper, [&helper] {
+        qDebug("quitting helper-start-wayland");
+        helper.stop();
+    });
+    QObject::connect(&helper, &WaylandHelper::failed, &app, [&app] {
+        QTextStream(stderr) << "Failed to start wayland session" << Qt::endl;
+        app.exit(35);
```
davidedmundson: and here

src/helper/HelperApp.cpp
aleixpol: This is just part of the review because of the reverted commit. Will address it because it's an easy thing to do.

src/helper/HelperStartWayland.cpp
aleixpol: Changed it to return 1 and 2.

src/helper/HelperApp.cpp
aleixpol: I'll move it, but this also is coming from the reverted commit.

aleixpol:
@davidedmundson addressed your comments, I didn't fold the changes into where they were added for the cases where it was coming from the reverted code.

src/helper/xorguserhelper.cpp
aleixpol: As the comment above says:

    // Make sure to forward the input of this process into the Xorg
    // server, otherwise it will complain that only console users are allowed

davidedmundson (approved):

yan12125:
This PR seems to break login for rootless X11. I got the following error from `journalctl`:

> Failed to take control of /dev/tty7: Operation not permitted

aleixpol:
> This PR seems to break login for rootless X11

Would you be able to create an issue and provide how you have it configured so I can reproduce and address it?

yan12125:
Ah, someone beats me :) https://github.com/sddm/sddm/issues/1456

PR #1475, opened 2021-11-21
Added localization of session names to Session.cpp

Cistoge:


src/common/Session.cpp
```diff
@@ -151,45 +153,59 @@ namespace SDDM {
 
         qDebug() << "Reading from" << m_fileName;
 
-        QFile file(m_fileName);
-        if (!file.open(QIODevice::ReadOnly))
-            return;
+        QSettings Settings(m_fileName, QSettings::IniFormat);
```
davidedmundson: style throughout sddm is to use lowerCamelCase for variables

src/common/Session.cpp
```diff
@@ -151,45 +153,59 @@ namespace SDDM {
...
+        QStringList Locales;
+	Locales << QLocale().name() << QLocale().name().remove(QRegExp(QLatin1String("_.*")));
```
davidedmundson: what is the second locale for?

src/common/Session.cpp
```diff
@@ -151,45 +153,59 @@ namespace SDDM {
...
+                if (LocalizedName.compare(Name) == 0)
+                        continue;
+                else
+                        break;
         }
 
-        file.close();
+        if (type == WaylandSession)
+                if ((Name.endsWith(QLatin1String(" (Wayland)"))) ||
+                        (Name.compare(QLatin1String("Last used session")) == 0))
```
davidedmundson: Please use the &== overload, it's far more readable

src/common/Session.cpp
```diff
@@ -151,45 +153,59 @@ namespace SDDM {
...
+                LocalizedComment = Settings.value(QLatin1String("Comment") + QLatin1String("[") + locale + QLatin1String("]"), Comment).toString();
+ 
+                if (LocalizedComment.compare(Comment) == 0)
+                        continue;
+                 else
+                        break;
+        }
+ 
+        m_comment = QObject::tr("%1").arg(LocalizedComment);
+        m_exec = Settings.value(QLatin1String("Exec"), QLatin1String("")).toString();
```
davidedmundson: Don't use QLatin1String("") as default values.

QString() (or just doing nothing) is better. A null string is more semantically correct than a valid empty string.

src/common/Session.cpp
```diff
@@ -151,45 +153,59 @@ namespace SDDM {
...
+        for (QString locale : qAsConst(Locales)) {
+                LocalizedComment = Settings.value(QLatin1String("Comment") + QLatin1String("[") + locale + QLatin1String("]"), Comment).toString();
+ 
+                if (LocalizedComment.compare(Comment) == 0)
+                        continue;
+                 else
+                        break;
+        }
+ 
+        m_comment = QObject::tr("%1").arg(LocalizedComment);
```
davidedmundson: Why are we going via tr?

src/common/Session.cpp
```diff
@@ -151,45 +153,59 @@ namespace SDDM {
...
+        QString LocalizedName;
 
-            if (line.startsWith(QLatin1String("["))) {
-                // The section name ends before the last ] before the start of a comment
-                int end = line.lastIndexOf(QLatin1Char(']'), line.indexOf(QLatin1Char('#')));
-                if (end != -1)
-                    current_section = line.mid(1, end - 1);
-            }
+        for (QString locale : qAsConst(Locales)) {
+                LocalizedName = Settings.value(QLatin1String("Name") + QLatin1String("[") + locale + QLatin1String("]"), Name).toString();
```
davidedmundson: This can all be reduced to:

```
        m_displayName = Settings.value(QLatin1String("Name"));
        for (QString locale : qAsConst(Locales)) {
                localizedName = Settings.value(QLatin1String("Name") + QLatin1String("[") + locale + QLatin1String("]")).toString();
                if (!LocalizedName.isEmpty()) {
                        m_displayName = localizedName;
                        break;
                 }
        }
```

davidedmundson: QLatin1Char for chars.

davidedmundson (changes requested):

src/common/Session.cpp
Cistoge: That is because in desktop files it is often the case that some locales are written with a language code only, while others are written with a language code and a territory code, separated by _.

src/common/Session.cpp
Cistoge: Ok, I changed that (will upload later) and hope that I did not forget a variable.

src/common/Session.cpp
```diff
@@ -151,45 +153,43 @@ namespace SDDM {
 
         qDebug() << "Reading from" << m_fileName;
 
-        QFile file(m_fileName);
-        if (!file.open(QIODevice::ReadOnly))
-            return;
+        QSettings settings(m_fileName, QSettings::IniFormat);
+        QStringList locales;
+        locales << QLocale().name() << QLocale().name().remove(QRegExp(QLatin1String("_.*")));
```
Vogtinator, line 158: `QRegExp` is deprecated and removed in Qt6

src/common/Session.cpp
```diff
@@ -151,45 +153,43 @@ namespace SDDM {
...
+        settings.beginGroup(QLatin1String("Desktop Entry"));
 
-            if (line.startsWith(QLatin1String("["))) {
-                // The section name ends before the last ] before the start of a comment
-                int end = line.lastIndexOf(QLatin1Char(']'), line.indexOf(QLatin1Char('#')));
-                if (end != -1)
-                    current_section = line.mid(1, end - 1);
-            }
+        m_displayName = settings.value(QLatin1String("Name"));
+        for (QString locale : qAsConst(Locales)) {
```
Vogtinator, line 166: Probably won't build, it's called `locales`

src/common/Session.cpp
```diff
@@ -151,45 +153,43 @@ namespace SDDM {
...
+                QString localizedName = settings.value(QLatin1String("Name") + QLatin1Char('[') + locale + QLatin1Char(']'), QString()).toString();
```
Vogtinator, line 167: This should probably be in a separate method like `getLocalizedDesktopKey`

Vogtinator (changes requested):

aleixpol:
This https://github.com/sddm/sddm/pull/1645 is a rebase of this change with the comments here addressed.

Vogtinator:
> This #1645 is a rebase of this change with the comments here addressed.

Perfect, so we can close this

PR #1478, opened 2021-11-26
Fix login of Wayland sessions with root Xorg

aleixpol:
If we already started the Xorg server, we'll need to give the wayland
session a different tty.

### DIFF
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -358,6 +358,12 @@ namespace SDDM {
         // last session later, in slotAuthenticationFinished()
         m_sessionName = session.fileName();
 
+        if (session.type() == Session::WaylandSession && m_displayServerType == X11DisplayServerType) {
+            // If it's an X11 display server, we'll have an (idling) Xorg process where
+            // m_terminalId is right now, we need to find another VT
+            m_terminalId = VirtualTerminal::setUpNewVt();
+        }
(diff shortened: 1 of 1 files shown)


a-andreyev (approved):
tried with PinePhone, working as expected

Strit:
Applying this MR as a patch also works for me.

src/daemon/Display.cpp
```diff
@@ -164,6 +164,11 @@ namespace SDDM {
         }
         if (findSessionEntry(mainConfig.Wayland.SessionDir.get(), autologinSession)) {
             sessionType = Session::WaylandSession;
+            if (m_displayServerType == X11DisplayServerType) {
+                // If it's an X11 display server, we'll have an (idling) Xorg process where
+                // m_terminalId is right now, we need to find another one
+                m_terminalId = VirtualTerminal::setUpNewVt();
```
davidedmundson: How does the non-autologin case of wayland on rootfull Xorg work?

aleixpol: Yes, this did not make a lot of sense. I've moved it down to startAuth which is where we used to deal with m_terminalId anyway.

I reproduced that it didn't work to log into wayland from root-x11 and that it works after this patch.

davidedmundson (approved):

PR #1484, opened 2021-12-17
Also spawn a separate tty when greeting

aleixpol:
This way we don't need to manage the state of the greeter that closely,
e.g. when auth fails and such, as there's cases where we will need to
have both the greeter and the actual session loaded side by side.

### DIFF
--- a/src/daemon/Greeter.cpp
+++ b/src/daemon/Greeter.cpp
@@ -315,6 +315,11 @@ namespace SDDM {
         }
     }
 
+    bool Greeter::isRunning() const {
+        return (m_process && m_process->state() == QProcess::Running)
+            || (m_auth->isActive());
+    }
+
     void Greeter::onReadyReadStandardError()
     {
         if (m_process) {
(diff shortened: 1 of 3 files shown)


ReillyBrogan:
I've been using SDDM master with this patch added and can confirm that it resolves my issue, and that no other regressions have been noticed.

src/daemon/Greeter.cpp
```diff
@@ -315,6 +315,12 @@ namespace SDDM {
         }
     }
 
+    bool Greeter::isRunning() const {
+        return (m_process && m_process->state() == QProcess::Running)
+            || (m_auth->isActive());
+        ;
```
davidedmundson: stray ;

davidedmundson:
>e.g. when auth fails and such,

This being "m_terminalId" doesn't make sense, it's not a property of the display anymore. It's a property of the session being spawned. Both the greeter and session belong to this display object but on different terminals.

I think we can make this a local var.

aleixpol:
Addressed @davidedmundson's comment. Good catch!

davidedmundson (approved):

PR #1489, opened 2022-01-05
greeter: Do not populate Wayland sessions if /dev/dri does not exist 

Conan-Kudo:
Check for `/dev/dri` which should indicate that the Wayland will work
on this hardware.

### DIFF
--- a/src/greeter/SessionModel.cpp
+++ b/src/greeter/SessionModel.cpp
@@ -22,6 +22,7 @@
 
 #include "Configuration.h"
 
+#include <QFileInfo>
 #include <QVector>
 #include <QProcessEnvironment>
 #include <QFileSystemWatcher>
@@ -40,19 +41,26 @@ namespace SDDM {
     };
 
     SessionModel::SessionModel(QObject *parent) : QAbstractListModel(parent), d(new SessionModelPrivate()) {
+        // Check for flag to show Wayland sessions
+        bool dri_active = QFileInfo::exists(QStringLiteral("/dev/dri"));
+
         // initial population
         beginResetModel();
-        populate(Session::WaylandSession, mainConfig.Wayland.SessionDir.get());
+        if (dri_active)
+            populate(Session::WaylandSession, mainConfig.Wayland.SessionDir.get());
         populate(Session::X11Session, mainConfig.X11.SessionDir.get());
         endResetModel();
 
         // refresh everytime a file is changed, added or removed
         QFileSystemWatcher *watcher = new QFileSystemWatcher(this);
         connect(watcher, &QFileSystemWatcher::directoryChanged, [this](const QString &path) {
+            // Recheck for flag to show Wayland sessions
+            bool dri_active = QFileInfo::exists(QStringLiteral("/dev/dri"));
             beginResetModel();
             d->sessions.clear();
             d->displayNames.clear();
-            populate(Session::WaylandSession, mainConfig.Wayland.SessionDir.get());
+            if (dri_active)
+                populate(Session::WaylandSession, mainConfig.Wayland.SessionDir.get());
             populate(Session::X11Session, mainConfig.X11.SessionDir.get());
             endResetModel();
         });


Conan-Kudo:
This patch comes from @jlintonarm for Fedora, and I'm attempting to upstream it as part of preparing a rebase to the latest SDDM code.

davidedmundson:
@adriaandegroot  is this the same in the BSD world?

aleixpol:
technically, we could have wayland sessions without DRI. In fact, if X.Org can I don't see why wayland couldn't. Should we at least make it a setting?

adriaandegroot:
@davidedmundson FreeBSD has `/dev/dri/card0` if DRI is available, yes.

Regarding the code itself:
 - I'd use `QFile::exists()` rather than `QFileInfo::exists()` myself, but I don't really know **why** I'd do that. The documentation doesn't mention advantages to either way.
 - `bool dri_active` could be made const (in both scopes) or, since it's used only once, calculated in-place inside the `if()`, or, if naming-things-well is a priority, turned into a static inline one-liner `dri_active()`.

lnxsr:
postmarketos-ui-plasma-desktop uses sddm. They support several aged mobile devices with old android vendor kernels missing proper (none/broken/outdated) DRI/KMS support.

Conan-Kudo:
Most Wayland environments do not work at all without DRI/KMS. Plasma Wayland supposedly does, but I've never managed to get kwin to work correctly with fbdev, and upstream Linux is in the process of ripping out fbdev entirely now that SimpleDRM exists.

XakepSDK:
Is `/dev/dri` standardized? I think there was C API (header) to access `/dev/dri` in implementation independent way, but i may be wrong, I'm not C Linux dev.

Conan-Kudo:
Reference downstream Fedora bugs:

* https://bugzilla.redhat.com/1952431
* https://bugzilla.redhat.com/show_bug.cgi?id=2016788
* https://bugzilla.redhat.com/show_bug.cgi?id=2016310

Vogtinator:
> Most Wayland environments do not work at all without DRI/KMS. 
> Plasma Wayland supposedly does, but I've never managed to get kwin to work correctly with fbdev

I had it working in VMs, but it's not a great experience. weston has a fbdev backend which works fairly well, but is not tried by default for some reason, so logging in fails anyway.

IMO this at least needs some refactoring. The check for `/dev/dri/` contents should be in a separate `bool isSystemWaylandCapable()` function or similar.

However, I'd also say that this is mostly unnecessary because on systems with logind, sddm will only start up on seats with the `CanGraphical=true` property, which is set if DRM devices are found (`master-of-seat` property in udev). That check can be disabled by using `nomodeset` though.

The progress of offering DRM for pretty much all devices (through SimpleDRM) also makes this less necessary IMO.

jlinton:
So this patch is slightly more robust version of a previous patch fedora was carrying which had issues with hammering the filesystem on every udev rule execution.

> However, I'd also say that this is mostly unnecessary because on systems with logind, sddm will only start up on seats with the `CanGraphical=true` property, which is set if DRM devices are found (`master-of-seat` property in udev). That check can be disabled by using `nomodeset` though.

It is a bit more complex, it will start a seat if logind/etc fail to respond in time, or there aren't any drm/CanGraphical seats. Which can happen for example because there is/was a bug in systemd that results in logind being stopped and then restarted when sddm is starting up during `systemctl isolate` calls. And of course, it is not just nomodeset that results in master-of-seat not being set, since it basically only gets set on DRM capable devices, which before simpleDRM includes a bunch of !drm devices (efifb for starters, older less popular graphics cards, etc).

So, yes, there are cases where sddm fails to start following X startup (and just leaves ones linux machine looking hung with a blank screen, and no log messages), but in this case sddm shows up, and presents the users with desktop sessions that themselves won't start.

I would worry that the /dev/dri would need a ifdef linux/etc wrapper, but that sounds like it isn't strictly needed although, I'm not sure all the enviroments sddm is actually run.

aleixpol (approved):
LGTM

PR #1494, opened 2022-01-17
Delay for logind, and fallback to seat0

Conan-Kudo:
There is systemd/logind race with when restarting
sddm that causes logind1 not to be available. Previously
this meant the seat0 was immediately created regardless
of the state of CanGraphical.

Fixing this, though we still want seat0 to be started
if none of the seats appear to be graphical. Presumably
there are some graphics on the machine, otherwise
why run sddm? Wait a bit, and create seat0 anyway. If
this fails the output from Xorg should tell us why. This
is generally a better strategy than what happens a good
amount of time now, where sddm is started and silent about
why the screen is blank.

### DIFF
--- a/src/daemon/LogindDBusTypes.cpp
+++ b/src/daemon/LogindDBusTypes.cpp
@@ -8,6 +8,8 @@
 
 #include <QDebug>
 
+#include <unistd.h>
+
 class LogindPathInternal {
 public:
     LogindPathInternal();
@@ -46,17 +48,26 @@ LogindPathInternal::LogindPathInternal()
     qRegisterMetaType<UserInfoList>("UserInfoList");
     qDBusRegisterMetaType<UserInfoList>();
 
-    if (QDBusConnection::systemBus().interface()->isServiceRegistered(QStringLiteral("org.freedesktop.login1"))) {
-        qDebug() << "Logind interface found";
-        available = true;
-        serviceName = QStringLiteral("org.freedesktop.login1");
-        managerPath = QStringLiteral("/org/freedesktop/login1");
-        managerIfaceName = QStringLiteral("org.freedesktop.login1.Manager");
-        seatIfaceName = QStringLiteral("org.freedesktop.login1.Seat");
-        sessionIfaceName = QStringLiteral("org.freedesktop.login1.Session");
-        userIfaceName = QStringLiteral("org.freedesktop.login1.User");
-        return;
+#ifdef HAVE_SYSTEMD
+    // systemd-logind should be running, although because it takes a few moments to restart after
+    // systemctl isolate calls, it may not yet be running. Wait a few seconds for it, while blocking everything else.
+    int logind_wait_seconds = 50;
+    while (logind_wait_seconds--) {
+        if (QDBusConnection::systemBus().interface()->isServiceRegistered(QStringLiteral("org.freedesktop.login1"))) {
+            qDebug() << "Logind interface found";
+            available = true;
+            serviceName = QStringLiteral("org.freedesktop.login1");
+            managerPath = QStringLiteral("/org/freedesktop/login1");
+            managerIfaceName = QStringLiteral("org.freedesktop.login1.Manager");
+            seatIfaceName = QStringLiteral("org.freedesktop.login1.Seat");
+            sessionIfaceName = QStringLiteral("org.freedesktop.login1.Session");
+            userIfaceName = QStringLiteral("org.freedesktop.login1.User");
+            return;
+        }
+	qDebug() << "Sleeping for systemd-logind";
+	usleep(100000);
     }
+#endif
 
     if (QDBusConnection::systemBus().interface()->isServiceRegistered(QStringLiteral("org.freedesktop.ConsoleKit"))) {
         qDebug() << "Console kit interface found";
--- a/src/daemon/SeatManager.cpp
+++ b/src/daemon/SeatManager.cpp
@@ -26,6 +26,9 @@
 #include <QDBusMessage>
 #include <QDBusPendingReply>
 #include <QDBusContext>
+#include <QDebug>
+#include <QFileInfo>
+#include <QTimer>
 
 #include "LogindDBusTypes.h"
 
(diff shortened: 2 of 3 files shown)


Conan-Kudo:
This patch comes from @jlintonarm for Fedora, and I'm attempting to upstream it as part of preparing a rebase to the latest SDDM code.

Conan-Kudo:
Reference downstream bugs:

* https://bugzilla.redhat.com/show_bug.cgi?id=2011991
* https://bugzilla.redhat.com/show_bug.cgi?id=2016310

src/daemon/SeatManager.h
```diff
@@ -49,6 +49,7 @@ namespace SDDM {
     private:
         QHash<QString, Seat *> m_seats; //these will exist only for graphical seats
         QHash<QString, LogindSeat*> m_systemSeats; //these will exist for all seats
+        void checkSeat(void);
```
aleixpol, line 52: the void argument shouldn't be there.

src/daemon/SeatManager.cpp
```diff
@@ -152,6 +157,23 @@ namespace SDDM {
...
+    // won't initially be available so there is a race
+    // between determing if a efifb/etc graphical object
+    // is the only graphics on the machine, or a DRM driver
+    // will take over the display. So we will hang out for a few
+    // seconds and if none of the seats are declared cangraphical
+    // its possible the only graphics on the machine don't have
+    // a drm driver.
+    void SeatManager::checkSeat(void) {
+        if (m_seats.isEmpty()) {
+	     //if (QFileInfo::exists(QStringLiteral("/dev/fb0"))) {
```
aleixpol, line 170: We can probably remove the commented if?

src/daemon/SeatManager.h
Conan-Kudo: Fixed.

src/daemon/SeatManager.cpp
Conan-Kudo: Fixed.

src/daemon/LogindDBusTypes.cpp
```diff
@@ -46,17 +48,26 @@ LogindPathInternal::LogindPathInternal()
...
+            available = true;
+            serviceName = QStringLiteral("org.freedesktop.login1");
+            managerPath = QStringLiteral("/org/freedesktop/login1");
+            managerIfaceName = QStringLiteral("org.freedesktop.login1.Manager");
+            seatIfaceName = QStringLiteral("org.freedesktop.login1.Seat");
+            sessionIfaceName = QStringLiteral("org.freedesktop.login1.Session");
+            userIfaceName = QStringLiteral("org.freedesktop.login1.User");
+            return;
+        }
+	qDebug() << "Sleeping for systemd-logind";
```
aleixpol, line 67: Needs extra 4 spaces to indent with the if.

aleixpol (commented):
Is it possible that you forgot to upload your changes? Can't see them applied.

Conan-Kudo:
@aleixpol Yup, I didn't push apparently...

src/daemon/LogindDBusTypes.cpp
Conan-Kudo: Fixed.

src/daemon/LogindDBusTypes.cpp
```diff
@@ -46,17 +48,26 @@ LogindPathInternal::LogindPathInternal()
...
-        managerIfaceName = QStringLiteral("org.freedesktop.login1.Manager");
-        seatIfaceName = QStringLiteral("org.freedesktop.login1.Seat");
-        sessionIfaceName = QStringLiteral("org.freedesktop.login1.Session");
-        userIfaceName = QStringLiteral("org.freedesktop.login1.User");
-        return;
+#ifdef HAVE_SYSTEMD
+    // systemd-logind should be running, although because it takes a few moments to restart after
+    // systemctl isolate calls, it may not yet be running. Wait a few seconds for it, while blocking everything else.
+    int logind_wait_seconds = 50;
+    while (logind_wait_seconds--) {
```
aleixpol, line 55: Could we just change the isServiceRegistered to startService?
```
auto reply = QDBusConnection::systemBus().interface()->startService(QStringLiteral("org.freedesktop.login1"));
if (reply.isValid()) {
```
This way we would be able to remove that ifdef and loop, they are a bit odd...

src/daemon/SeatManager.cpp
aleixpol: I can still see them :eye:

src/daemon/LogindDBusTypes.cpp
aleixpol: I've put the proposal here for you to test: https://github.com/aleixpol/sddm/tree/delay-seat-init

jlinton: Hey, so the problem isn't that logind isn't starting, its that while sddm is marked as dependent in logind, it looked like systemd was getting ahead of itself and sddm was starting before logind was registered on dbus. Hence the delay, is enough to slow sddm down enough for logind to arrive.

I opened a github issue about it: but have not returned to the problem. In part because the simpledrm driver has largly fixed the core issue and i'm a bit confused about how it was happening too, despite having clear evidence of such, so maybe its was just some kind of dbus ordering issue.

https://github.com/systemd/systemd/issues/22022

aleixpol: Okay, then let's see how this resolves in systemd and we can follow up on this later on.

Vogtinator:
I don't really understand the issue itself or this workaround:

1. Why is `systemctl isolate graphical.target` called during boot?
2. Why does that trigger a **stop** of `logind`? It's should be needed by `multi-user.target` already
3. Why does systemd stop logind if it's also pulled in by `graphical.target`? (Maybe there's some issue that the service isn't actually pulled in, but only got started through dbus activation?)
4. Why does it work to sleep until the service is available on the bus instead of just sending a message and waiting for a reply with the usual dbus timeout?

Issue #1353, opened 2021-01-02
Cursor size does not respect dpi


joshuakraemer:
My X server is set to 144 dpi. In X, the mouse cursor size is chosen accordingly (32 px). In SDDM however, the smallest cursor is used (24 px). The size can be overridden with the environment variable `XCURSOR_SIZE`, but it should be set automatically.

dylif:
This could be a duplicate of #569

Issue #965, opened 2018-01-18
Buggy when the cursor size get changed


jacob-emery:
I am using KDE Neon Standard Edition with the latest SDDM.
Since my laptop features a 14' 1080p screen, I prefer to use a cursor that is at least 36 pixel. In the Plasma desktop I managed to change the size without problems. However, having the cursor's size small in the SDDM Login screen and getting the right one after the login is not very appealing. After a lot of research, I figured out how to set another size for the SDDM Login. I opened the console and I executed:
SUDO_EDITOR=kwrite sudoedit /etc/environment
Afterwards, I added the line:
XCURSOR_SIZE=36
I think that I also did something else because the tweak wasn't working in the beginning.

There are still some bugs:

a) The size of the cursor is initially the default one. At the smallest mouse/touchpad movement, the size gets fixed. It should start already with the right size.

b) If you press on other Desktop Environments dialog you will have to choose which desktop environment you want to use. When you haven't choosen yet the mouse gets the smaller size again. It gets the right size automatically after choosing a DE. It would be nice if it has always the same size.

c) It seems that when the password to login is given, the password changes again for a vary small span of time.

I think it would be a good idea to fix these bugs and to give a possibility to choose not only which cursor theme to use, but also its dimention in /etc/sddm.conf.
Thank you for your attention and for the effort you devs put into this project!

joshuakraemer:
I do not see changing cursor sizes with sddm-0.19.0_3, so this is probably solved.

PR #1498, opened 2022-01-20
Set XCURSOR_SIZE in XorgDisplayServer::start

### LINKED ISSUE #569: Mouse cursor size when using multimonitor.
I noticed that the mouse is enlarged when using more than one monitor.

I have currently {Left, Middle, Right} identical monitors.
Left Portrait 1080x1920.
Middle/Right Landscape 1920x1080.

When running only with one monitor the cursor is correct size, however when connecting my other monitors it appears enlarged.

dylif:
This fully fixes #569 and #1353, and perhaps even  #965. Although an option to specify a cursor size was given in #1446, this only sets XCURSOR_SIZE in the Greeter; it should also be set during initialisation of the X server to prevent sudden changes in cursor size.

### DIFF
--- a/src/daemon/XorgDisplayServer.cpp
+++ b/src/daemon/XorgDisplayServer.cpp
@@ -96,6 +96,9 @@ namespace SDDM {
         // set process environment
         QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
         env.insert(QStringLiteral("XCURSOR_THEME"), mainConfig.Theme.CursorTheme.get());
+        QString xcursorSize = mainConfig.Theme.CursorSize.get();
+        if (!xcursorSize.isEmpty())
+            env.insert(QStringLiteral("XCURSOR_SIZE"), xcursorSize);
         process->setProcessEnvironment(env);
 
         //create pipe for communicating with X server


aleixpol:
Does this mean that we don't need to set it anymore on Greeter.cpp?
Would you be able to test it with rootless X? i.e. `DisplayServer=x11-user` there

dylif:
We would still need to set it in Greeter.cpp since that takes the user's theme into account.
Edit: If it's appropriate, I could modify my submission to include accounting for the user's theme.

As for rootless X, I have yet to try it on my system, but I could definitely try and get it working and report back here.

aleixpol:
> As for rootless X, I have yet to try it on my system, but I could definitely try and get it working and report back here.

I'd very much appreciate it. It should just work by adding the line I mentioned in your sddm.conf.

If you have any questions, feel free to ask.

dylif:
> I'd very much appreciate it. It should just work by adding the line I mentioned in your sddm.conf.
> 
> If you have any questions, feel free to ask.

For sure, I'll report back here as soon as I give it a try. I appreciate your helpfulness :)

dylif:
Can confirm the fix works as intended when using the `x11-user` DisplayServer. For some reason KDE started on tty2, but I suspect this is an unrelated issue.

For reference here are the relevant config files:

99_cursor.conf:
```
[Theme]
CursorTheme=Breeze_Snow
CursorSize=24
```

99_xuser.conf:
```
[General]
DisplayServer=x11-user
```

Edit: It should be noted that I have **not** set `XCURSOR_SIZE=24` in `/etc/environment`, which I was previously using as a workaround on the latest release. This is good.

plfiorini (approved):

DanielHCoding:
Unfortunately, the solution does not work for me. Setting CursorSize=24 in any of the SDDM Config Files does not change the size. Only ```XCURSOR_SIZE=24``` in ```/etc/environment``` make a change but with the disadvantage that the wrong size is displayed for a short time anyway. 
Is there any way to change the default cursor size for SDDM?

jeois:
"CursorSize=xx" in the various `sddm.conf` paths such as `/etc/sddm.conf.d/` does nothing for me as well.  It doesn't change cursor size the during X init or at the greeter.  

Using `XCURSOR_SIZE` env variable causes jumps in cursor size during a couple phases of startup, like the previous commenter said.

Also, SDDM can't find cursors in `~/.local/share/icons`.  Some cursor themes install there or to ~/.icons per user instead of globally to `/usr/share/icons`.  Sorry, I know this is a related but separate issue, and I'll create an issue if I can't figure it out.  It's disconcerting to be manually installing and breaking files out of packages.  Is there way to add a cursor theme PATH?

Vogtinator:
> Also, SDDM can't find cursors in ~/.local/share/icons. Some cursor themes install there or to ~/.icons per user instead of globally to /usr/share/icons. Sorry, I know this is a related but separate issue, and I'll create an issue if I can't figure it out. It's disconcerting to be manually installing and breaking files out of packages. Is there way to add a cursor theme PATH?

No, you'd have to install the cursor theme either system wide or into the home directory of the `sddm` user.

jeois:
>... into the home directory of the `sddm` user.

Ok, thanks for the info, but what you mean by "home directory of `sddm` user"?  I assume you don't mean directly into `~/` without any subdirectories?

Vogtinator:
No, the home of the sddm user is probably `/var/lib/sddm` or similar.

PR #1506, opened 2022-02-03
WaylandHelper: support non-default display names

alebastr:
Weston v10[1] and some wlroots-based compositors[2],[3] decided to stop using `wayland-0`.
Extend WaylandSocketWatcher to look for any socket matching the `wayland-?` pattern and pass the `WAYLAND_DISPLAY` to the greeter.

[1]: https://gitlab.freedesktop.org/wayland/weston/-/merge_requests/486
[2]: https://github.com/swaywm/sway/commit/65a751a21f61b30808b7e703257c6ca3b71f50eb
[3]: https://github.com/WayfireWM/wayfire/commit/4ee4f3f259be00fd9a3c26960e8fce91ec526290

***
Tested with: sway v1.7 (`wayland-1`), weston v10 (`wayland-1`), weston v8 (`wayland-0`).

### DIFF
--- a/src/helper/waylandsocketwatcher.cpp
+++ b/src/helper/waylandsocketwatcher.cpp
@@ -18,6 +18,7 @@
 * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 ***************************************************************************/
 
+#include <unistd.h>
 #include <QDebug>
 #include <QStandardPaths>
 
@@ -28,8 +29,9 @@ namespace SDDM {
 WaylandSocketWatcher::WaylandSocketWatcher(QObject *parent )
     : QObject(parent)
     , m_runtimeDir(QDir(QStandardPaths::writableLocation(QStandardPaths::RuntimeLocation)))
-    , m_socketPath(m_runtimeDir.absoluteFilePath(QLatin1String("wayland-0")))
 {
+    m_runtimeDir.setFilter(QDir::Files | QDir::System);
+    m_runtimeDir.setNameFilters(QStringList() << QLatin1String("wayland-?"));
 }
 
 WaylandSocketWatcher::Status WaylandSocketWatcher::status() const
@@ -37,9 +39,9 @@ WaylandSocketWatcher::Status WaylandSocketWatcher::status() const
     return m_status;
 }
 
-QString WaylandSocketWatcher::socketPath() const
+QString WaylandSocketWatcher::socketName() const
 {
-    return m_socketPath;
+    return m_socketName;
 }
 
 void WaylandSocketWatcher::start()
@@ -53,8 +55,7 @@ void WaylandSocketWatcher::start()
         // Time is up and a socket was not found
         if (!m_watcher.isNull())
             m_watcher->deleteLater();
-        qWarning("Wayland socket watcher for \"%s\" timed out",
-                 qPrintable(m_socketPath));
+        qWarning("Wayland socket watcher timed out");
         m_status = Failed;
         Q_EMIT failed();
     });
@@ -62,22 +63,28 @@ void WaylandSocketWatcher::start()
     // Check if the socket exists
     connect(m_watcher, &QFileSystemWatcher::directoryChanged, this,
             [this](const QString &path) {
-        qDebug() << "Directory" << path << "has changed, checking for" << m_socketPath;
+        qDebug() << "Directory" << path << "has changed, checking for Wayland socket";
 
-        if (QFile::exists(m_socketPath)) {
-            m_timer.stop();
-            if (!m_watcher.isNull())
-                m_watcher->deleteLater();
-            m_status = Started;
-            Q_EMIT started();
+        m_runtimeDir.refresh();
+        const QFileInfoList fileInfoList = m_runtimeDir.entryInfoList();
+        for (const QFileInfo &fileInfo : fileInfoList) {
+            if (fileInfo.ownerId() == ::getuid()) {
+                qDebug() << "Found Wayland socket" << fileInfo.absoluteFilePath();
+                m_timer.stop();
+                if (!m_watcher.isNull())
+                    m_watcher->deleteLater();
+                m_socketName = fileInfo.fileName();
+                m_status = Started;
+                Q_EMIT started();
+                break;
+            }
         }
     });
 
     // Watch for runtime directory changes
     if (!m_runtimeDir.exists() || !m_watcher->addPath(m_runtimeDir.absolutePath())) {
-        qWarning("Cannot watch directory \"%s\" for Wayland socket \"%s\"",
-                 qPrintable(m_runtimeDir.absolutePath()),
-                 qPrintable(m_socketPath));
+        qWarning("Cannot watch directory \"%s\" for Wayland socket",
+                 qPrintable(m_runtimeDir.absolutePath()));
         m_watcher->deleteLater();
         m_status = Failed;
         Q_EMIT failed();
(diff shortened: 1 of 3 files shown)


Conan-Kudo (approved):

src/helper/waylandsocketwatcher.cpp
```diff
@@ -28,18 +28,20 @@ namespace SDDM {
 WaylandSocketWatcher::WaylandSocketWatcher(QObject *parent )
     : QObject(parent)
     , m_runtimeDir(QDir(QStandardPaths::writableLocation(QStandardPaths::RuntimeLocation)))
-    , m_socketPath(m_runtimeDir.absoluteFilePath(QLatin1String("wayland-0")))
+    , m_socketName(QLatin1String("wayland-?"))
```
aleixpol: I'd leave m_socketName as empty. It's a better way to signal we don't know the name. Don't want to return "wayland-?" on ::socketPath.

src/helper/waylandsocketwatcher.cpp
```diff
@@ -28,18 +28,20 @@ namespace SDDM {
...
+    m_runtimeDir.setFilter(QDir::Files | QDir::System);
+    m_runtimeDir.setNameFilters(QStringList() << m_socketName);
 }
 
 WaylandSocketWatcher::Status WaylandSocketWatcher::status() const
 {
     return m_status;
 }
 
-QString WaylandSocketWatcher::socketPath() const
```
aleixpol, line 40: Let's just remove this method entirely? Nobody is calling socketPath anyway.

aleixpol (changes requested):
I'd say the bottom line is that we want to make sure there is a wayland-* socket in the runtime directory. Let's just trim our API to that.

src/helper/waylandsocketwatcher.cpp
alebastr: `WaylandHelper::startGreeter` is calling `WaylandSocketWatcher::socketName` to get the `WAYLAND_DISPLAY` value. That's actually important, as the Wayland socket has to be explicitly passed to the greeter if it's not `wayland-0`.

src/helper/waylandsocketwatcher.cpp
alebastr: It's there pretty much for logging purposes only - all these qDebugs printing what file the watcher is expecting.

There's a big shortcut I took with not separating m_socketName and m_socketPattern and not making the pattern configurable. I'd love to address that, but I can't figure out if there are ways to pass the value from the config to the helper other than setting the env variable in Greeter.cpp.
I'll do the first part and separate the variables.

(Wayland socket doesn't have to match `wayland-\d+` and some debug tools are actually using different value patterns.)

src/helper/waylandsocketwatcher.h
```diff
@@ -54,7 +54,8 @@ class WaylandSocketWatcher : public QObject
 private:
     Status m_status = Stopped;
     QDir m_runtimeDir;
-    QString m_socketPath;
+    QString m_socketName;
+    QString m_socketPattern;
```
aleixpol: I'd still drop the m_socketPattern. It's not worth keeping only for logging.

src/helper/waylandsocketwatcher.cpp
```diff
@@ -54,20 +56,23 @@ void WaylandSocketWatcher::start()
...
 
     // Check if the socket exists
     connect(m_watcher, &QFileSystemWatcher::directoryChanged, this,
             [this](const QString &path) {
-        qDebug() << "Directory" << path << "has changed, checking for" << m_socketPath;
+        qDebug() << "Directory" << path << "has changed, checking for" << m_socketPattern;
 
-        if (QFile::exists(m_socketPath)) {
+        m_runtimeDir.refresh();
+        QStringList files = m_runtimeDir.entryList();
```
aleixpol: It would make sense to also check that the socket we found is for the correct user. You can easily check that using entryInfoList.
Also remember to make it const.

src/helper/waylandsocketwatcher.h
alebastr: Done.

src/helper/waylandsocketwatcher.cpp
alebastr: Sure, added the check and verified that `wayland-0` owned by root is ignored.

alebastr:
Ping? Anything else I can do to get that merged?

dania83:
I have compiled sddm with the 3 commits -> sddm is launched with weston 10.0.0 to 11.0.0 but there is no more mouse pointer on the login greeter to select the session to start or select the keyboard. The login works. I have also add the commits for  https://github.com/sddm/sddm/pull/1581 Now we have a US flag in the menu but the symptoms remains the same , its impossible to select a keyboard under wayland

dania83:
i have compiled sddm with the 3 commits and now sddm is launched with weston 10.0.0 and 11.0.0 but there is no mouse pointer on the greeter to select the session and the keyboard. The login works. I have also add to sddm the two commits from https://github.com/sddm/sddm/pull/1581 to resolve the keyboard probleme selection under wayland . It doesn't work, the only difference we have a US flag now replacing the AF flag .

src/helper/waylandsocketwatcher.cpp
```diff
@@ -53,31 +55,35 @@ void WaylandSocketWatcher::start()
...
+        qWarning("Wayland socket watcher timed out");
         m_status = Failed;
         Q_EMIT failed();
     });
 
     // Check if the socket exists
     connect(m_watcher, &QFileSystemWatcher::directoryChanged, this,
             [this](const QString &path) {
-        qDebug() << "Directory" << path << "has changed, checking for" << m_socketPath;
+        qDebug() << "Directory" << path << "has changed, checking for Wayland socket";
```
aleixpol, line 66: still print `m_runtimeDir.absolutePath()` so we can diagnose when debugging.

aleixpol:
Added a small detail, would be happy to approve after. Excuse the delay in my review.

src/helper/waylandsocketwatcher.cpp
alebastr: `QFileSystemWatcher` passes to the signal the same value it received via `addPaths`. In this case it would be `m_runtimeDir.absolutePath()`, so the change you suggested will not affect the output.

But I agree that there's not enough info for debugging. Added a message with a full path to the socket we found:
```
Dec 13 20:05:41 sddm-helper-start-wayland[1074]: Starting Wayland process "weston --shell=fullscreen-shell.so" "sddm"
Dec 13 20:05:41 sddm-helper-start-wayland[1074]: started succesfully "weston --shell=fullscreen-shell.so"
<... weston startup logs ...>
Dec 13 20:05:41 sddm-helper-start-wayland[1074]: Directory "/run/user/979" has changed, checking for Wayland socket
Dec 13 20:05:41 sddm-helper-start-wayland[1074]: Directory "/run/user/979" has changed, checking for Wayland socket
Dec 13 20:05:41 sddm-helper-start-wayland[1074]: Found Wayland socket "/run/user/979/wayland-1"
```

aleixpol (approved):

Conan-Kudo:
FYI, this patch is being carried downstream in Fedora for the upcoming Fedora Sway spin.

CoelacanthusHex:
Hope this patch can be merged soon. It's necessary to use sddm on wayland with weston (and this is our **DEFAULT** value) and wlroots-based compositor.

PR #1514, opened 2022-02-14
Set QT_USE_PHYSICAL_DPI if hidpi is enabled

antlarr:
When the EnableHiDPI config option is set, Qt doesn't
automatically change the dpi of the screen. This requires
manual configuration leaving in the meantime too small fonts
for users with a hidpi monitor.

This commit sets the QT_USE_PHYSICAL_DPI environment variable
when the EnableHiDPI config option is set, which makes Qt
read the monitor edid data using xrandr and calculate the
required DPI to use automatically without user intervention
nor manual configuration.

### DIFF
--- a/src/greeter/GreeterApp.cpp
+++ b/src/greeter/GreeterApp.cpp
@@ -328,6 +328,7 @@ int main(int argc, char **argv)
     if (hiDpiEnabled) {
         qDebug() << "High-DPI autoscaling Enabled";
         QGuiApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+        qputenv("QT_USE_PHYSICAL_DPI", "1");
     } else {
         qDebug() << "High-DPI autoscaling not Enabled";
     }


antlarr:
I've also considered to maybe also add an option like "UsePhysicalDPI" which would be enabled by default so sddm would use the EDID information automatically (giving a better user experience by default) but users with monitors that deliver broken EDID data, can still disable that behavior and set the DPI manually by editing the ServerArguments option as it's done now. Would that be wanted?

dobey:
> Would that be wanted?

If I understand correctly, the current PR means that those users with broken EDID would always have no way to fix an issue with the DPI? Also, how does this affect running on Wayland instead of Xorg?

Vogtinator:
I'm not entirely sure what the behaviour of `QT_USE_PHYSICAL_DPI` is on wayland and X11. Does it use the exact DPI or only integer steps? Will it also go below an effective scale of 1?

Issue #1545, opened 2022-05-01
Think about letting RememberLastUser get to UserModel::lastUser()?


UMU618:
https://github.com/sddm/sddm/blob/e67307e4103a8606d57a0c2fd48a378e40fcef06/src/greeter/UserModel.cpp#L177

Such as:

```cpp
if (mainConfig.Users.RememberLastUser.get()) {
    return stateConfig.Last.User.get();
} else {
    return QString();
}
```

#626

PR #1546, opened 2022-05-01
[greeter] UserModel::lastUser() return empty if RememberLastUser=false

UMU618:
#1545


aleixpol:
Can you please elaborate a bit further on what this exactly fixes? The change seems straightforward but the issue doesn't explain what problem you were seeing.

UMU618:
It's a security issue. If user set RememberLastUser=false, it means that lastUser() should be empty.

Vogtinator:
How is this a security issue? The state file is not readable by unprivileged users and the `LastUser` is also not confidential information.

UMU618:
> How is this a security issue? The state file is not readable by unprivileged users and the `LastUser` is also not confidential information.

If themes can get it, they can show it, then this configuration item is meaningless.

Vogtinator:
> > How is this a security issue? The state file is not readable by unprivileged users and the `LastUser` is also not confidential information.
> 
> If themes can get it, they can show it, then this configuration item is meaningless.

`RememberLastUser` determines whether the `Last.User` setting is written or cleared on logins. If it's false and there's a (successful) login, the property is cleared and neither `lastIndex` nor `lastUser` are available.

In any case, I don't see how this can have security implications. Themes are considered trusted.

UMU618:
long time ago, I used a theme, and it showed me my username even I set RememberLastUser=false, and I didn't want it to show, so I fix this.

Now I am using Gnome.......

Vogtinator:
> long time ago, I used a theme, and it showed me my username even I set RememberLastUser=false, and I didn't want it to show, so I fix this.

The fix would've been to just log in once after changing that setting. Afterwards, `Last.User` is empty.

IMO this is just a slightly unexpected behaviour of the configuration option combined with bad documentation.
The behaviour exactly matches the name of the configuration option though.

PR #1548, opened 2022-05-06
Add a ForceUsers option to display users which are not in passwd

flesniak:
The ForceUsers option accepts comma-separated UIDs or Usernames which are resolved using getpwnam/getpwuid. This way, accounts from external authentication providers like LDAP, AD etc. can be forced to be shown.

In addition, this removes the previous behavior of dropping anything after a comma sign from the gecos field. I was not able to find out why this was implemented in the first place - can anyone comment on that? In my case, the displayName LDAP field is formatted like "Surname, First names".

### DIFF
--- a/src/greeter/UserModel.cpp
+++ b/src/greeter/UserModel.cpp
@@ -72,9 +72,24 @@ namespace SDDM {
         const QString iconURI = QStringLiteral("file://%1").arg(
                 QFile::exists(themeDefaultFace) ? themeDefaultFace : defaultFace);
 
-        bool lastUserFound = false;
-
         struct passwd *current_pw;
+
+        for (auto forced_user : mainConfig.Users.ForceUsers.get()) {
+            bool is_uid = false;
+            int uid = forced_user.toInt(&is_uid);
+            if (is_uid)
+                current_pw = getpwuid(uid);
+            else
+                current_pw = getpwnam(forced_user.toLocal8Bit().data());
+
+            if (current_pw == nullptr)
+                continue;
+
+            UserPtr user { new User(current_pw, iconURI) };
+            d->users << user;
+        }
+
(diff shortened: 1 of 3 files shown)


flesniak:
Okay, so originally the GECOS field is a comma-separated list of the user's full name, room and phone number etc. I'm looking for an alternative solution for the 2nd commit, as apparently it should be merged as-is.

Vogtinator:
What's the use-case for this?

Even with LDAP/AD/etc. the list of user accounts should be complete. Are there too many that the full list is disabled?

I wonder whether some other mechanism like always showing the last 5 user accounts or so might fix this in an even more useful way.

flesniak:
My use-case is a large AD environment. We have more than 70.000 users, so user enumeration takes a lot of time and is therefore disabled.  Thus, `getpwent()` does not return any AD/LDAP users and the users list is empty.

However, I agree that remembering the last N user accounts that have been entered manually is much more useful and might also benefit users in other situations. I expect that to be much more work than this small change, so I need to see when I got some spare time to see how to implement it.

We can leave this MR on hold if you prefer and see whether I come up with a better approach, or close it and I'll open a new one in case.

src/greeter/UserModel.cpp
```diff
@@ -72,9 +72,24 @@ namespace SDDM {
...
+            if (is_uid)
+                current_pw = getpwuid(uid);
+            else
+                current_pw = getpwnam(forced_user.toLocal8Bit().data());
+
+            if (current_pw == nullptr)
+                continue;
+
+            UserPtr user { new User(current_pw, iconURI) };
+            d->users << user;
```
Vogtinator, line 89: This should probably be after the loop and only add users which aren't already in the l ist

Issue #930, opened 2017-11-07
[feature] autologin with cached password from systemd-cryptsetup


SebastianS90:
I have a feature proposal regarding autologin:

It is currently not possible to automatically unlock kwallet when using autologin and a non-empty wallet password.
When booting a luks-encrypted system, then systemd prompts for a password. This password [is cached](https://github.com/systemd/systemd/blob/v235/NEWS#L2181-L2195) in the kernel keyring for user root.
When autologin is enabled, then SDDM should retrieve the password from there and inject it into the pam stack, such that `pam_kwallet5.so` can use it. With that addition, the following scenario will be possible:
* When booting the machine, I am prompted for the hard disk encryption password
* After entering that password, the system starts up including starting my desktop session and unlocking the wallet (using the same password as provided during boot) without any further interruption
* With the open kwallet, I can also unlock other applications

GDM already [has that feature](https://github.com/GNOME/gdm/commit/31ed6f2b3f1ab45ae07aad41c13a51ba91fd159d) and it doesn't look very complicated.

Vincent43:
What's practical difference between this and using an empty password?

SebastianS90:
I have checked the *Promt when an application accesses a wallet* option in kwallet, so evil processes cannot just query my open wallet for secret data. But an evil process running under my user account could steal the wallet file from disk. If I have a password set for the wallet and my user account, then the wallet file is encrypted. If the password is empty, then the wallet is not encrypted.

So the difference is added security for the case that something bad happens in my user account.

Vincent43:
I think you greatly overestimate kwallet security. It's barely maintained tool with outdated encryption algorithms. It's better to treat it as management tool not security tool. For protecting user account you can try solutions like selinux,apparmor,firejail,flatpak etc. For secure password manager check [pass](https://www.passwordstore.org/).

https://news.ycombinator.com/item?id=9715432
https://bbs.archlinux.org/viewtopic.php?id=233278

Grief:
@Vincent43 kwallet is on an encrypted partition, this is the question of the usability only. If the feature @SebastianS90 was asking is the implemented then it would be possible to have a single password to securely boot the system from start until unlocked session and kwallet/gnome-keyring.

Vincent43:
@Grief My point is that you can achieve the same thing with empty kwallet password without writing single line of code.

Grief:
[...]

PR #1550, opened 2022-05-15
Add pam module to read cached password from systemd-cryptsetup.

h0cheung:
This solves #930.
Use the source file `pam_gdm.c` in gdm, build it as `pam_sddm.so` to avoid conflicting.
Then edit pam config and systemd units for it to work.


cmake/Findkeyutils.cmake
```diff
@@ -0,0 +1,27 @@
+# Try to find Keyutils
+# Once done, this will define
+#
+# KEYUTILS_FOUND - system has keyutils
+# KEYUTILS_INCLUDE_DIR - the keyutils include directories
+# KEYUTILS_LIBRARIES - link these to use keyutils
+
+find_path(KEYUTILS_INCLUDE_DIR keyutils.h PATHS
```
davidedmundson, line 8: keyutils (on my system at least) has a pkgconfig file.
We should wrap that rather than manual calls where possible, it's more future proof

src/pam_sddm/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+include(KDEInstallDirs)
+
+find_package(PAM REQUIRED)
+find_package(keyutils REQUIRED)
```
davidedmundson, line 4:
```
find_package(keyutils REQUIRED)
add_definitions(-DHAVE_KEYUTILS)
```

something here is off. We're making this required, yet have a code path to make it optional

src/pam_sddm/pam_sddm.c
```diff
@@ -0,0 +1,131 @@
...
+#include <keyutils.h>
+#endif
+
+int
+pam_sm_authenticate (pam_handle_t  *pamh,
+                     int            flags,
+                     int            argc,
+                     const char   **argv)
+{
+#ifdef HAVE_KEYUTILS
```
davidedmundson, line 41: Is there a reason we build this pam module at all if we don't have keyutils?

If not, we could move this to a cmake check above the add_subdirectory

davidedmundson (commented):
Overall it makes sense. I'm always happy to steal things from Gnome.

A few cmake issues

cmake/Findkeyutils.cmake
aleixpol: `pkg_check_modules(Keyutils IMPORTED_TARGET libkeyutils)` would suffice.

Vogtinator:
The commit messages need a lot more explanation

jinliu:
A similar PAM module was added in systemd 255, so we don't need another module. Just add this line to /etc/pam.d/sddm-autologin:
`-auth        optional    pam_systemd_loadkey.so`
or if the LUKS passphrase is bound to TPM:
`-auth        optional    pam_systemd_loadkey.so keyname=luks2-pin`

PR #1562, opened 2022-06-19
Redesign Xauth handling

Kangie:
This is a rebase of #1230 by @Vogtinator

> This commit moves Xauthority handling over to libXau.
> Advantage is that this allows use of FamilyWild, is faster, more reliable
> and easier to read. However, we lose the ability to merge the new cookie into
> an existing Xauthority file, so support for using a non-temporary file is
> dropped. Even if merging was implemented manually, use of FamilyWild would
> "infect" such a file and break it for DMs which don't write it.

### DIFF
--- a/src/common/Configuration.h
+++ b/src/common/Configuration.h
@@ -68,11 +68,9 @@ namespace SDDM {
             Entry(ServerPath,          QString,     _S("/usr/bin/X"),                           _S("Path to X server binary"));
             Entry(ServerArguments,     QString,     _S("-nolisten tcp"),                        _S("Arguments passed to the X server invocation"));
             Entry(XephyrPath,          QString,     _S("/usr/bin/Xephyr"),                      _S("Path to Xephyr binary"));
-            Entry(XauthPath,           QString,     _S("/usr/bin/xauth"),                       _S("Path to xauth binary"));
             Entry(SessionDir,          QString,     _S("/usr/share/xsessions"),                 _S("Directory containing available X sessions"));
             Entry(SessionCommand,      QString,     _S(SESSION_COMMAND),                        _S("Path to a script to execute when starting the desktop session"));
-	    Entry(SessionLogFile,      QString,     _S(".local/share/sddm/xorg-session.log"),   _S("Path to the user session log file"));
-	    Entry(UserAuthFile,        QString,     _S(".Xauthority"),                          _S("Path to the Xauthority file"));
+        Entry(SessionLogFile,      QString,     _S(".local/share/sddm/xorg-session.log"),   _S("Path to the user session log file"));
             Entry(DisplayCommand,      QString,     _S(DATA_INSTALL_DIR "/scripts/Xsetup"),     _S("Path to a script to execute when starting the display server"));
             Entry(DisplayStopCommand,  QString,     _S(DATA_INSTALL_DIR "/scripts/Xstop"),      _S("Path to a script to execute when stopping the display server"));
             Entry(EnableHiDPI,         bool,        false,                                      _S("Enable Qt's automatic high-DPI scaling"));
@@ -82,7 +80,7 @@ namespace SDDM {
             Entry(CompositorCommand,   QString,     _S("weston --shell=fullscreen-shell.so"),   _S("Path of the Wayland compositor to execute when starting the greeter"));
             Entry(SessionDir,          QString,     _S("/usr/share/wayland-sessions"),          _S("Directory containing available Wayland sessions"));
             Entry(SessionCommand,      QString,     _S(WAYLAND_SESSION_COMMAND),                _S("Path to a script to execute when starting the desktop session"));
-	    Entry(SessionLogFile,      QString,     _S(".local/share/sddm/wayland-session.log"),_S("Path to the user session log file"));
+            Entry(SessionLogFile,      QString,     _S(".local/share/sddm/wayland-session.log"),_S("Path to the user session log file"));
             Entry(EnableHiDPI,         bool,        false,                                      _S("Enable Qt's automatic high-DPI scaling"));
         );
 
--- a/src/common/XAuth.cpp
+++ b/src/common/XAuth.cpp
@@ -18,15 +18,20 @@
 * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 ***************************************************************************/
 
+#include <limits.h>
 #include <QDebug>
 #include <QDir>
+#include <QString>
 #include <QUuid>
+#include <random>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <X11/Xauth.h>
 
 #include "Configuration.h"
 #include "Constants.h"
 #include "XAuth.h"
-
-#include <random>
 
 namespace SDDM {
 
@@ -55,7 +60,7 @@ QString XAuth::authPath() const
     return m_authPath;
 }
 
-QString XAuth::cookie() const
+QByteArray XAuth::cookie() const
 {
     return m_cookie;
 }
@@ -77,15 +82,14 @@ void XAuth::setup()
     // Generate cookie
     std::random_device rd;
     std::mt19937 gen(rd());
-    std::uniform_int_distribution<> dis(0, 15);
+    std::uniform_int_distribution<> dis(0, 0xFF);
 
-    // Reseve 32 bytes
-    m_cookie.reserve(32);
+    QByteArray m_cookie;
+    m_cookie.reserve(16);
 
     // Create a random hexadecimal number
-    const char *digits = "0123456789abcdef";
-    for (int i = 0; i < 32; ++i)
-        m_cookie[i] = QLatin1Char(digits[dis(gen)]);
+    for(int i = 0; i < 16; i++)
+        m_cookie[i] = dis(gen);
 }
 
 bool XAuth::addCookie(const QString &display)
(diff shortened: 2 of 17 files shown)


Kangie:
As best as I can tell I've managed to preserve the advantages of the PR without introducing any bugs while rebasing.

The biggest change overall was merging `XAuthHelper.{h,cpp}` into `XAuth`, otherwise it pretty much follows the previous PR.

Everything seems to work as expected; Happy to make changes to address any feedback.

src/daemon/XorgDisplayServer.cpp
```diff
@@ -61,7 +61,7 @@ namespace SDDM {
         return QStringLiteral("x11");
     }
 
-    QString XorgDisplayServer::cookie() const {
+    const QByteArray XorgDisplayServer::cookie() const {
```
staticssleever668, line 64: This probably meant
```suggestion
    const QByteArray &XorgDisplayServer::cookie() const {
```

src/daemon/XorgDisplayServer.h
```diff
@@ -39,7 +39,7 @@ namespace SDDM {
 
         QString sessionType() const;
 
-        QString cookie() const;
+        const QByteArray cookie() const;
```
staticssleever668, line 42: This probably meant
```suggestion
        const QByteArray &cookie() const;
```

plfiorini:
@Vogtinator @davidedmundson

Kangie:
Thanks for the feedback, I'll get it addressed this week!

Apologies, been sidetracked for the last 10 days or so.

src/daemon/XorgDisplayServer.h
aleixpol: There's little value in const& on implicitly shared objects. I'd suggest to just leave it as `QByteArray cookie() const;` as the const has also no value.

aleixpol:
It would make sense to come up with a test plan for this. The code looks okay and it's clearly needed, a clear plan to evaluate that this works better than the status quo would help us integrate the change.

Vogtinator:
The easiest way to test whether this has the desired effect is to change the hostname of the system from inside an X11 session. Without this change, X11 applications won't be able to connect anymore. On top of that, verifying that the file operations are done properly (correct permissions, no races) would be useful.

(I hope I can find some time to review/work on this again soon...)

src/daemon/XorgDisplayServer.h
staticssleever668: I have little experience with Qt, and I didn't know about implicit sharing, thanks!
It obviously won't matter here, but I wonder what would be faster in a hot code section, a reference or a shallow copy? 🤔

By the way, this method is only used in [a single place](https://github.com/sddm/sddm/blob/b7155e6f7db9f34dc9e69c9f525110657bfddd9d/src/helper/UserSession.cpp#L60), and it always makes a copy (although a shallow one since implicit sharing). If my suggestion is the one that ends up being accepted, in this line a const& type should be used. I would also use `auto` there, like so:
```cpp
            const auto& cookie = qobject_cast<HelperApp*>(parent())->cookie();
```

aleixpol: This is no hot section or anything, so it doesn't matter. It should just be QByteArray.

aleixpol:
> (I hope I can find some time to review/work on this again soon...)

Does this mean that this work in progress?

Kangie:
> The easiest way to test whether this has the desired effect is to change the hostname of the system from inside an X11 session. Without this change, X11 applications won't be able to connect anymore.

I'll see what I can do!

a17r:
Any update here?

Kangie:
I've been busy for a few months unfortunately, but this is still on my radar! I'm hoping to have some time to validate the hostname changing before the end of year.

If there's an upcoming release that I'm not tracking I can try and bump this up in my priority list - I've a couple days off in the coming weeks that I should be able to squeeze a few hours of validation & verification into.

a17r:
> If there's an upcoming release that I'm not tracking I can try and bump this up in my priority list

Things seem to be moving again right now towards a new release.

Kangie:
I have next week off. I'll schedule time to recompile this, do the hostname test, and report back!

src/common/Configuration.h
```diff
@@ -68,11 +68,9 @@ namespace SDDM {
...
             Entry(ServerArguments,     QString,     _S("-nolisten tcp"),                        _S("Arguments passed to the X server invocation"));
             Entry(XephyrPath,          QString,     _S("/usr/bin/Xephyr"),                      _S("Path to Xephyr binary"));
-            Entry(XauthPath,           QString,     _S("/usr/bin/xauth"),                       _S("Path to xauth binary"));
             Entry(SessionDir,          QString,     _S("/usr/share/xsessions"),                 _S("Directory containing available X sessions"));
             Entry(SessionCommand,      QString,     _S(SESSION_COMMAND),                        _S("Path to a script to execute when starting the desktop session"));
-	    Entry(SessionLogFile,      QString,     _S(".local/share/sddm/xorg-session.log"),   _S("Path to the user session log file"));
-	    Entry(UserAuthFile,        QString,     _S(".Xauthority"),                          _S("Path to the Xauthority file"));
+        Entry(SessionLogFile,      QString,     _S(".local/share/sddm/xorg-session.log"),   _S("Path to the user session log file"));
```
Vogtinator, line 73: Whitespace error

src/common/XAuth.cpp
```diff
@@ -77,15 +82,14 @@ void XAuth::setup()
     // Generate cookie
     std::random_device rd;
     std::mt19937 gen(rd());
-    std::uniform_int_distribution<> dis(0, 15);
+    std::uniform_int_distribution<> dis(0, 0xFF);
 
-    // Reseve 32 bytes
-    m_cookie.reserve(32);
+    QByteArray m_cookie;
```
Vogtinator, line 87: This declaration is broken, `m_cookie` is a member variable.

src/common/XAuth.cpp
```diff
@@ -99,29 +103,61 @@ bool XAuth::addCookie(const QString &display)
 }
 
 bool XAuth::addCookieToFile(const QString &display, const QString &fileName,
-                            const QString &cookie)
+                            const QByteArray &m_cookie)
```
Vogtinator, line 106: This shadows the member variable.

Either use `cookie` here or remove the parameter.

src/daemon/XorgDisplayServer.cpp
Vogtinator: `XAuth::cookie()` returns a value. Returning a reference to that would result in a dangling pointer.

src/helper/UserSession.cpp
```diff
@@ -50,11 +51,37 @@ namespace SDDM {
...
     bool UserSession::start() {
         auto helper = qobject_cast<HelperApp*>(parent());
-        QProcessEnvironment env = helper->session()->processEnvironment();
+        QProcessEnvironment env = processEnvironment();
 
         bool isWaylandGreeter = false;
         if (env.value(QStringLiteral("XDG_SESSION_TYPE")) == QLatin1String("x11")) {
             QString command;
+            // Create the Xauthority file
+            QByteArray cookie = qobject_cast<HelperApp*>(parent())->cookie();
```
Vogtinator, line 60: Use `helper->cookie()` here

Vogtinator (changes requested):
This is unfortunately broken: The shadowing of `m_cookie` means that it's always empty and the display server doesn't even attempt to start.

I noticed that with the merge of `XAuthUtils` into `XAuth` you removed the copyright info, please add that back.

Please also keep the original commit message and authorship info. You can add `Co-Authored-by: Matt Jolly <your.em@il.addr>` to the description.

davidedmundson:
>This commit moves Xauthority handling over to libXau.

FWIW, In Kwin we have some xauth handling done ourselves instead of even needing the lib:

https://invent.kde.org/plasma/kwin/-/blob/master/src/xwayland/lib/xauthority.cpp

might be worth copy pasting?

Vogtinator:
> > This commit moves Xauthority handling over to libXau.
> 
> FWIW, In Kwin we have some xauth handling done ourselves instead of even needing the lib:
> 
> https://invent.kde.org/plasma/kwin/-/blob/master/src/xwayland/lib/xauthority.cpp
> 
> might be worth copy pasting?

IMO kwin should also use libXau instead of code duplication/NIH.

src/helper/UserSession.cpp
```diff
@@ -50,11 +51,37 @@ namespace SDDM {
...
+
+            if (!m_xauthFile.open()) {
+                qCritical() << "Could not create the Xauthority file";
+                return false;
+            }
+
+            QString display = processEnvironment().value(QStringLiteral("DISPLAY"));
+            qDebug() << "Adding cookie to" << m_xauthFile.fileName();
+
+            if (!XAuth::addCookieToFile(m_xauthFile.fileName(), display, cookie)) {
```
Vogtinator, line 77: Wrong order of parameters:

```suggestion
            if (!XAuth::addCookieToFile(display, m_xauthFile.fileName(), cookie)) {
```

Kangie:
Appreciate the feedback. I'll get that amended and tested over the next week while I'm off (this time for real, I promise!)

Vogtinator:
With the review comments addressed autologin X11 works (FWICT), but the greeter does not start because the HelperApp receives an empty cookie. I didn't test xorg user sessions or wayland greeter sessions yet.

Vogtinator:
I fixed the greeter session and some other issues I found like using `strdup` to copy the binary cookie data and pushed that to #1230.

I didn't test wayland greeter or xorg user sessions yet.

Vogtinator:
#1230 got merged :partying_face:

Issue #434, opened 2015-06-25
RebootCommand option in sddm.conf seems ignored


egreshko:
There is a problem when shutting down a system which uses Wifi as its network connection and has NFS mounts.  The network is dropped before unmounts so if you try to do a normal "systemctl poweroff" the system will take a very long time to shutdown.  So, one must unmount the NFS mounts first.  Running this script from the command line works well.

#!/bin/sh
/usr/bin/umount -a -t nfs4 
/usr/bin/systemctl reboot

So, I placed the script in /usr/sbin and modified sddm.conf to contain

RebootCommand=/usr/sbin/rebooty

However, clicking on "Reboot" from the login screen seems to have no effect.  The system does not unmount the NFS mounts and reboot takes forever.  

For completeness I'm running Fedora F22 with sddm-0.10.0-4 I have filed a BZ with Fedora but the suggestions haven't helped.  https://bugzilla.redhat.com/show_bug.cgi?id=1231413

plfiorini:
Proposal:
* RebootCommand is empty by default
* login1 backend will use RebootCommand if specified otherwise the D-Bus request
* upower backend will use RebootCommand (assuming "reboot" if it's empty)

jleclanche:
Do we do this for 0.12? If we dont, we have to document that RebootCommand isn't used in login1.

jleclanche:
I know we just said on IRC we can release without it, but I'm starting to think we should do this for 0.12.

giddie:
Any progress on this? I'm still seeing this issue in 0.13 (ArchLinux)

flyser:
still present.

giddie:
Still broken :(

felagund:
Indeed. I wanted to set the reboot option to `/bin/systemctl kexec` but it just does not seem to register.

iam-TJ:
The option `RebootCommand` in `sddm.conf` is only valid when the `UPower` backend is being used, and that backend is only used if both DBus services `login1` and `ConsoleKit` are not found at run-time.

The best solution, instead of mixing up `RebootCommand` may be to add `BackendPreferred=` to `sddm.conf` with options `login1|ConsoleKit|UPower`.

In which case instead of the first available backend being used to trigger the reboot `BackendPreferred` would be chosen if available.

Then for users that expect `RebootCommand=` to be used they can set `BackendPreferred=UPower`. There'd need to be a log entry confirming (or not) that the preferred back-end is available at run-time.

[...]

PR #1563, opened 2022-06-21
PowerManager: enable use of kexec with systemd-logind

iam-TJ:
DBus interface has changed, interface XML required re-generation via
introspection, and custom types defining for IN/OUT arguments, all so
login1.Manager.RebootWithFlags() can be used.

Addresses: issue #434.

### DIFF
--- a/src/daemon/PowerManager.cpp
+++ b/src/daemon/PowerManager.cpp
@@ -27,6 +27,7 @@
 #include <QDBusInterface>
 #include <QDBusReply>
 #include <QProcess>
+#include <QDebug>
 
 namespace SDDM {
     /************************************************/
@@ -47,6 +48,7 @@ namespace SDDM {
         virtual void suspend() const = 0;
         virtual void hibernate() const = 0;
         virtual void hybridSleep() const = 0;
+        QString service;
     };
 
     /**********************************************/
@@ -60,7 +62,9 @@ const QString UPOWER_OBJECT = QStringLiteral("org.freedesktop.UPower");
     class UPowerBackend : public PowerManagerBackend {
     public:
         UPowerBackend(const QString & service, const QString & path, const QString & interface) {
+            qDebug() << "sddm QDBusInterface service =" << service << "path =" << path << "interface =" << interface;
             m_interface = new QDBusInterface(service, path, interface, QDBusConnection::systemBus());
+            this->service = service;
         }
 
         ~UPowerBackend() {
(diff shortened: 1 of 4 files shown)


src/daemon/PowerManager.cpp
```diff
@@ -174,8 +181,15 @@ const QString CK2_OBJECT = QStringLiteral("org.freedesktop.ConsoleKit.Manager");
         }
 
         void reboot() const {
-            if (!daemonApp->testing())
-                m_interface->call(QStringLiteral("Reboot"), true);
+            if (!daemonApp->testing()) {
+                if (m_interface->service() == LOGIN1_SERVICE) {
+                    qDebug() << "sddm calling RebootWithFlags";
+                    m_interface->call(QStringLiteral("RebootWithFlags"),
```
davidedmundson, line 187: It's weird that we're updating the interface, only to the use a string interface that doesn't use generated code.

```
m_interface->call(QStringLiteral("RebootWithFlags"),
```

If you've gone to the effort of updating the XML we may as well do:

```
auto message = m_interface->RebootWithFlags();
message << LOGIN1_FLAGS_KEXEC;
m_interface->connection()->call(message);
```

src/daemon/PowerManager.cpp
```diff
@@ -174,8 +181,15 @@ const QString CK2_OBJECT = QStringLiteral("org.freedesktop.ConsoleKit.Manager");
         }
 
         void reboot() const {
-            if (!daemonApp->testing())
-                m_interface->call(QStringLiteral("Reboot"), true);
+            if (!daemonApp->testing()) {
+                if (m_interface->service() == LOGIN1_SERVICE) {
```
davidedmundson, line 185: There is the case of an old systemd.

This will still be Login1, but the method won't exist.
I'm pretty confident this patch as-is would break on Neon for example.

src/daemon/PowerManager.cpp
```diff
@@ -60,7 +62,9 @@ const QString UPOWER_OBJECT = QStringLiteral("org.freedesktop.UPower");
     class UPowerBackend : public PowerManagerBackend {
     public:
         UPowerBackend(const QString & service, const QString & path, const QString & interface) {
+            qDebug() << "sddm QDBusInterface service =" << service << "path =" << path << "interface =" << interface;
```
davidedmundson, line 65: can we drop some of the debug

davidedmundson (changes requested):

aleixpol:
@iam-TJ have you been able to go through David's comments?

Issue #1276, opened 2020-07-23
Import environment variable from systemd user session


wengxt:
GDM has done this in the commit: https://gitlab.gnome.org/GNOME/gdm/commit/448134d3cdbc54e5359ea33d387993b0defdaefa

This allows user to set environment variable based on config in ~/.config/environment.d/

hexchain:
As a desktop user I feel there currently lacks a general way to set environment variables (e.g. for input methods to work) for a whole user session. `.xprofile` only works for X sessions in some cases, and there is no Wayland counterpart. `.pam_environment` is deliberately disabled by default with the latest linux-pam release (1.4.0). IMHO, `environment.d` seems to be a better place for such things (if it could work, like in the GDM way).

and3rson:
Any updates on this?

chmanie:
Is there a workaround for this? Is `.pam_environment` the only option as of now for wayland sessions?

hexchain:
It is possible to modify the SDDM wayland session script to make it source other shell scripts. You might want to make a copy of the stock script and then point `[Wayland].SessionCommand` to the copy.

guihkx:
A simple workaround I found after I migrated from GDM+GNOME to SDDM+Plasma, is to just create a `.sh` file (e.g. `systemd-env.sh`) in the `~/.config/plasma-workspace/env` directory, with the following:

```bash
export $(/usr/lib/systemd/user-environment-generators/30-systemd-environment-d-generator | xargs)
```

Then just re-login and your environment variables should be available to the entire Plasma session.

hexchain:
I think this issue isn't very relevant anymore for full-fledged DEs (i.e. KDE and GNOME). IIUC both GNOME and KDE runs themselves ~~and user applications~~ as systemd services, which naturally inherit everything from `environment.d`.

It is still useful for other lightweight desktops (e.g. i3/sway/...), though.

EDIT: correction, user applications are not launched as systemd services by default in 5.27.5.

Jackenmen:
> I think this issue isn't very relevant anymore for full-fledged DEs (i.e. KDE and GNOME). IIUC both GNOME and KDE runs themselves and user applications as systemd services, which naturally inherit everything from environment.d.

I'm unsure if this is true, see:
https://github.com/sddm/sddm/pull/1616#issuecomment-1415753026

guihkx:
[...]

PR #1574, opened 2022-08-09
add loading environment.d feature to session scripts

Brli:
try to implement #1276

### DIFF
--- a/data/scripts/Xsession
+++ b/data/scripts/Xsession
@@ -60,6 +60,7 @@ OPTIONFILE=/etc/X11/Xsession.options
 USERXSESSION=$HOME/.xsession
 USERXSESSIONRC=$HOME/.xsessionrc
 ALTUSERXSESSION=$HOME/.Xsession
+ENVIRONMENT_DIR=$HOME/.config/environment.d
 
 if [ -d "$xsessionddir" ]; then
     for i in `ls $xsessionddir`; do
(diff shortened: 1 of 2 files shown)


plfiorini:
The GDM commit referenced in #1276 is getting the environment from systemd rather than sourcing ~/.config/environment.d
Am I missing something?

Brli:
the following arguements may not be related to each other, just, listing:

1. Currently, the `environment.d` proposed by [systemd](https://www.freedesktop.org/software/systemd/man/environment.d.html) uses the common format of legacy environmental variable files, such as shellrc(bashrc zshrc etc.), shell profile(bash_profile, zprofile etc.), user profile (.profile, .pam_environment), and graphic profiles (xinitrc, xprofile) etc. Thus for now, a simple `source` would do the trick (to add support for this...standard?). (omit the fact that some files require to use `export`)

3. **But we cannot guarentee the systemd team won't change the `KEY=value` tuple format in the future**

4. My implementation **cannot** avoid possible security risk that malicious code injected fishy shell script to user's directory, which is later being sourced.

5. the GDM implementation is done in calling dbus, with strict type check, thus do fail fast for possible malicious code execution? maybe.

6. But, _THAT IS DBUS_, which would lead to sddm binding harder to systemd. I'm not sure your (the team's) attitude toward this, or you try to support as many use case as possible. -> On this point of view, _this_ "source" method is neutral/independent to shell or initial system.

7. on, the other hand, my code is just a rewrite to sddm's xsession [script](https://github.com/sddm/sddm/blob/develop/data/scripts/Xsession#L64), which, also read without check the mentioned possible flaw AFAIK. (or maybe it is trimmed in other part of the code?)

Brli:
BTW, the CI failure has nothing to do with the commit changes, it is about how zypper/suse dealt with dependencies.

plfiorini:
As you point out, sourcing files from `environment.d` is not safe in case of malicious code. I know we have old code that does that, but I wouldn't add new if we have an alternative.

The attitude towards systemd is to have modular code that allows to support systemd and its alternatives.

I would use the D-Bus API on systemd systems, the way GDM does, instead of sourcing files.

Jackenmen:
This doesn't import the env vars from all `environment.d` directories:
https://www.freedesktop.org/software/systemd/man/environment.d.html

Conan-Kudo:
> But we cannot guarentee the systemd team won't change the KEY=value tuple format in the future

I'm pretty sure they're not going to change it.

aleixpol:
Here's an alternative implementation that does ask the systemd1 dbus service about the environment.
https://github.com/sddm/sddm/pull/1616

As implemented there, we have two advantages:
- we don't rely on a flimsy bash script
- we also initialise the environment on Wayland sessions as well as on the greeters.

PR #1581, opened 2022-09-09
Allow using org.freedesktop.locale1 to switch keyboard layouts on Wayland

### LINKED ISSUE #1528: keyboard layout selections don't work in SDDM with Wayland greeter
_This was originally reported by @adamwill as ***[RHBZ#2062415](https://bugzilla.redhat.com/show_bug.cgi?id=2062415)***_

In Fedora 36 and Rawhide, whatever language/keyboard layout you select during installation, SDDM will show the selected keyboard layout as `af` when you boot up. This is Afrikaans, which AIUI is functionally very similar to the US English layout (I think it supports some additional characters with dead keys).

I tried to test what keyboard layout is *actually* being used. I did an install in UK English, with a user called `test` with password `test`. After install, I logged in, verified the UK English keyboard layout was being used in KDE, and changed the user's password to:

`weakpassword"`

Note that in the UK English keyboard layout, you type the `"` character with <kbd>shift</kbd>+<kbd>2</kbd>. On US English and similar layouts, you type it with <kbd>shift</kbd>+<kbd>'</kbd>, where <kbd>'</kbd> is the key to the left of <kbd>Enter</kbd>. On a US English keyboard, <kbd>shift</kbd>+<kbd>2</kbd> gives the character `@`.

So, if the keyboard layout in use is UK English, I should be able to log in by typing <kbd>shift</kbd>+<kbd>2</kbd>; if it's US English or similar I should type <kbd>shift</kbd>+<kbd>'</kbd>.

If I leave the layout selector in SDDM at `af`, I can log in by typing <kbd>shift</kbd>+<kbd>'</kbd>, which is consistent with the layout being `af` or `us`.

If I change the layout selector in SDDM to `gb`, which should give me a UK English layout, I cannot log in by typing <kbd>shift</kbd>+<kbd>2</kbd> as expected. I still have to type <kbd>shift</kbd>+<kbd>'</kbd>. So it seems like changing the layout does not work.

Executive summary: keyboard layouts in SDDM are just hosed. I think it's either always using `af` or always using `us`, and selecting layouts just does not work. I wonder if this has never actually been hooked up to work properly on Wayland?

In KDE itself, this does work properly. The UK layout is used by default as expected.

aleixpol:
Fixes https://github.com/sddm/sddm/issues/1528

### DIFF
--- a/src/daemon/DaemonApp.cpp
+++ b/src/daemon/DaemonApp.cpp
@@ -34,6 +34,7 @@
 #include <QTimer>
 
 #include <iostream>
+#include <locale1_interface.h>
 
 namespace SDDM {
     DaemonApp *DaemonApp::self = nullptr;
@@ -118,6 +119,12 @@ namespace SDDM {
 
     int DaemonApp::newSessionId() {
         return m_lastSessionId++;
+    }
+
+    void DaemonApp::setKeyboardLayout(const QString &layout) {
+        OrgFreedesktopLocale1Interface locale1(QStringLiteral("org.freedesktop.locale1"), QStringLiteral("/org/freedesktop/locale1"),
+                                                   QDBusConnection::systemBus());
+        locale1.SetX11Keyboard(layout, {}, {}, {}, false, false);
     }
 }
 
--- a/src/daemon/SocketServer.cpp
+++ b/src/daemon/SocketServer.cpp
@@ -181,6 +181,15 @@ namespace SDDM {
                 daemonApp->powerManager()->hybridSleep();
             }
             break;
+            case GreeterMessages::KeyboardLayout: {
+                qDebug() << "Message received from greeter: Keyboard Layout switch";
+
+                // read username, pasword etc.
+                QString layout;
+                input >> layout;
+
+                daemonApp->setKeyboardLayout(layout);
+            }
             default: {
                 // log message
                 qWarning() << "Unknown message" << message;
--- a/src/greeter/GreeterProxy.cpp
+++ b/src/greeter/GreeterProxy.cpp
@@ -108,6 +108,10 @@ namespace SDDM {
         SocketWriter(d->socket) << quint32(GreeterMessages::HybridSleep);
     }
 
+    void GreeterProxy::sendKeyboardLayout(const QString &layout) {
+        SocketWriter(d->socket) << quint32(GreeterMessages::KeyboardLayout) << layout;
+    }
+
     void GreeterProxy::login(const QString &user, const QString &password, const int sessionIndex) const {
         if (!d->sessionModel) {
             // log error
--- a/src/greeter/KeyboardModel.h
+++ b/src/greeter/KeyboardModel.h
@@ -25,6 +25,7 @@
 #include <QString>
 
 namespace SDDM {
+    class GreeterProxy;
     class KeyboardModelPrivate;
     class KeyboardBackend;
 
@@ -45,6 +46,8 @@ namespace SDDM {
     public:
         KeyboardModel();
         virtual ~KeyboardModel();
+
+        void setProxy(GreeterProxy *proxy);
 
     signals:
         void numLockStateChanged();
--- a/src/greeter/waylandkeyboardbackend.cpp
+++ b/src/greeter/waylandkeyboardbackend.cpp
@@ -25,6 +25,7 @@
 #include <QGuiApplication>
 #include <QInputMethod>
 
+#include "GreeterProxy.h"
 #include "KeyboardModel.h"
 #include "KeyboardModel_p.h"
 #include "KeyboardLayout.h"
@@ -41,7 +42,6 @@ WaylandKeyboardBackend::WaylandKeyboardBackend(KeyboardModelPrivate *kmp)
 WaylandKeyboardBackend::~WaylandKeyboardBackend()
 {
 }
-
 
 QList<QObject *> parseRules(const QString &filename, int &current)
 {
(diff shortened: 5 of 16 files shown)


src/greeter/waylandkeyboardbackend.cpp
```diff
@@ -35,14 +43,55 @@ WaylandKeyboardBackend::~WaylandKeyboardBackend()
...
+    if (reader.hasError()) {
+        qWarning() << "Failed to parse the rules file" << file.fileName();
+        return {};
+    }
+    return layouts;
+}
+
+void WaylandKeyboardBackend::init()
+{
+    d->layouts = parseRules(QStringLiteral("/usr/share/X11/xkb/rules/evdev.xml"), d->layout_id);
```
bam80, line 93: `libxkbregistry` is upstream-blessed library for parsing such things. We could use it instead of parsing `evdev.xml` directly:
> The libxkbregistry API to query for available rules, models, layouts, variants and options (RMLVO).
>
> libxkbregistry is a separate library to libxkbcommon.
>
> **This library is the replacement for clients currently parsing evdev.xml directly.** The library is intended to provide easy access to the set of possible MLVO configurations for a given ruleset. It is not a library to apply these configurations, merely to enumerate them. The intended users of this library are the configuration UIs that allow a user to select their keyboard layout of choice. 

https://xkbcommon.org/doc/current/group__registry.html

aleixpol: Do you know if it does the translating or we still need to do it ourselves?

TBF, this code already works (as it's taken from KDE) so I don't think bringing a dependency is worth the hassle.

bam80: > Do you know if it does the translating or we still need to do it ourselves?

I don't know, sorry.

> TBF, this code already works (as it's taken from KDE) so I don't think bringing a dependency is worth the hassle.

That might be, just an option.
We would probably want to use that library in Plasma anyway, just something to keep in mind for the future.

src/daemon/DaemonApp.cpp
```diff
@@ -119,6 +120,12 @@ namespace SDDM {
     int DaemonApp::newSessionId() {
         return m_lastSessionId++;
     }
+
+    void DaemonApp::keyboardLayout(const QString &layout) {
+        OrgFreedesktopLocale1Interface locale1(QStringLiteral("org.freedesktop.locale1"), QStringLiteral("/org/freedesktop/locale1"),
+                                                   QDBusConnection::systemBus());
+        locale1.SetX11Keyboard(layout, {}, {}, {}, false, false);
```
davidedmundson, line 127: Is there a reason the greeter can't make the call directly without proxying through the server?

davidedmundson: >Interactive authentication required.

It needs root, fair enough.

src/daemon/DaemonApp.cpp
```diff
@@ -119,6 +120,12 @@ namespace SDDM {
     int DaemonApp::newSessionId() {
         return m_lastSessionId++;
     }
+
+    void DaemonApp::keyboardLayout(const QString &layout) {
```
davidedmundson: Method name shouldn't sound like a getter.

src/daemon/SocketServer.cpp
```diff
@@ -181,6 +181,15 @@ namespace SDDM {
                 daemonApp->powerManager()->hybridSleep();
             }
             break;
+            case GreeterMessages::KeyboardLayout: {
+                qDebug() << "Requested Keyboard Layout switch";
```
davidedmundson: excessive debug

src/greeter/GreeterProxy.cpp
```diff
@@ -108,6 +108,10 @@ namespace SDDM {
         SocketWriter(d->socket) << quint32(GreeterMessages::HybridSleep);
     }
 
+    void GreeterProxy::keyboardLayout(const QString &layout) {
```
davidedmundson: This sounds like a getter without any verbs

src/greeter/waylandkeyboardbackend.cpp
```diff
@@ -59,7 +108,10 @@ void WaylandKeyboardBackend::dispatchEvents()
 
 void WaylandKeyboardBackend::connectEventsDispatcher(KeyboardModel *model)
 {
-    Q_UNUSED(model);
+    connect(model, &KeyboardModel::currentLayoutChanged, this, [this, model] {
```
davidedmundson: This has nothing to do with connecting event dispatchers. 

This is more akin to Backend::sendChanges which tells backends to sync chosen items out.
Changing also probably means the QObject inheritance is not needed.

src/greeter/KeyboardModel.h
```diff
@@ -46,6 +47,8 @@ namespace SDDM {
         KeyboardModel();
         virtual ~KeyboardModel();
 
+        void setProxy(GreeterProxy *proxy);
```
davidedmundson, line 50: If a member always needs to be set in order to not crash and is constant throughout the lifespan of the object it's better to pass it as an arg in the constructor.

src/daemon/SocketServer.cpp
aleixpol: Literally every switch case has a debug statement, what makes this one excessive?

davidedmundson: It was the only one visible in the diff :)

If the others have similar debugs, ignore this comment.

src/greeter/KeyboardModel.h
aleixpol: Made it so it's not mandatory. We don't yet have the GreeterProxy when we construct the KeyboardModel.

davidedmundson (approved):

PR #1585, opened 2022-09-13
Fix CI, seems like SUSE changed the names of their packages

aleixpol:


davidedmundson (approved):

Vogtinator:
FWICT the issue is a different: The base image is so old that it still includes python 3.8 as default, but Tumbleweed uses 3.10 as default meanwhile (so `python3-docutils` == `python310-docutils`) but installing that conflicts with `python38-docutils` in the old base. So you actually need to update the base image and revert the change here.

aleixpol:
Hi @Vogtinator, do you know of a drop-in replacement?

I do not have access to the ones used right now, I'll see what I can find...

Vogtinator:
FWICT they're pretty much the same as the KDE CI images, so if those can be pulled from here, they should work fine.

Vogtinator:
> FWICT they're pretty much the same as the KDE CI images, so if those can be pulled from here, they should work fine.

Apparently they can: https://github.com/sddm/sddm/pull/1586

I'd +1, but apparently GitHub changed it so that reviewing merged PRs is no longer possible.

PR #1586, opened 2022-09-14
CI: Use KDE's CI images

aleixpol:
The current ones seem to be out of date and these we need to maintain anyway for many other projects.


plfiorini:
I was pushing an update to my image while at the same time wondering if it was possible to use those from KDE!

plfiorini (approved):

PR #1589, opened 2022-09-16
Address shutdown lockups

aleixpol:
* Makes sure our helper processes receive the sigterms as necessary, assert if the API is misused.
* Always react to sigterm signals.
* Do not wait for a process that isn't started to finish.

### DIFF
--- a/src/auth/Auth.cpp
+++ b/src/auth/Auth.cpp
@@ -383,6 +383,10 @@ namespace SDDM {
     }
 
     void Auth::stop() {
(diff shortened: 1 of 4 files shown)


davidedmundson (changes requested):
This change doesn't match the commit title:
" SignalHandler should happen before QCoreApplication ". Then the patch puts it after and asserts that it's after

aleixpol:
@davidedmundson quite right, addressed the commit message.

davidedmundson (approved):

PR #1591, opened 2022-09-19
Display: Do not offer a tty if it's already taken by a logind session

### LINKED ISSUE #1588: User switching breaks
We should look into whether we are switching ttys properly...

https://bugs.kde.org/show_bug.cgi?id=458169

aleixpol:
Otherwise, it all freezes and nothing works as it should.

Fixes #1588


src/daemon/Display.cpp
```diff
@@ -50,7 +54,30 @@
...
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
+
+    int fetchAvailableVt() {
+        const auto vt = VirtualTerminal::currentVt();
+        return approveTty(QStringLiteral("tty%1").arg(vt)) ? vt : VirtualTerminal::setUpNewVt();
```
davidedmundson: On BSD:

currentVt will be -1
approveTTY will return true as we have no logind

so we return -1.  Same for if currentVt fails on Linux.

We want:
```
const auto vt = VirtualTerminal::currentVt();
if (vt > 0) {
    if (approveVt(vt) {
        return vt;
    }
} 
return setupNewVt();
```

src/daemon/Display.cpp
```diff
@@ -50,7 +54,30 @@
 #include "WaylandDisplayServer.h"
 
 
-namespace SDDM {
+namespace SDDM
+{
+    bool approveTty(const QString &desiredTty) {
```
davidedmundson: I don't like the name. "approve"  sounds like a verb, implying it does something active, also it's not a term logind or ioctl's are using

bool isTtyInUse(...)

src/daemon/Display.cpp
aleixpol: good catch, I wonder why it worked on the testing then... :thinking:

plfiorini:
Wonder what happens on systems without logind

davidedmundson (approved):

plfiorini (approved):

arrowd:
On FreeBSD we use SDDM with consolekit, which probably has its own way to find out if the vt is free.

plfiorini:
Yeah, for now `isTtyInUse()` returns false but in the future it can do whatever is needed for consolekit

PR #1603, opened 2022-10-24
disable automatic portal launching

hsitter:
in Qt6 (and the KDE patch collection for Qt 5) genericunixservices will internally attempt to probe the portal tech early on in the app life cycle. this causes the protal system to launch app and then crash because we aren't actually providing a fully functional session. instead opt out of this altogether


Vogtinator:
Which changes (in which components) are involved in this crash? I don't see that here, with a relatively recent Qt patch collection installed.

hsitter:
xdg-desktop-portal-kde also see https://bugs.kde.org/show_bug.cgi?id=458865

Vogtinator:
Ok, so FWICT the crash is because of a packaging issue (?) and this is basically just an optimization.

PR #1613, opened 2022-11-16
Provide a fallback mechanism for failed sessions

aleixpol:
If a session fails to load, we go into the greeter. If the greeter fails to load, we are in trouble.
This change adds a fallback into launch the greeter in a x11-user setup in case that happens as a failsafe.

To test you can configure DisplayServer=wayland and set an unexisting CompositorCommand (e.g. `CompositorCommand=banana`).

An extra test would be to add a faulty autologin session on all of that above.

### DIFF
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -83,40 +83,45 @@ namespace SDDM {
         return VirtualTerminal::setUpNewVt();
     }
 
-    Display::Display(Seat *parent) : QObject(parent),
-        m_auth(new Auth(this)),
-        m_seat(parent),
-        m_socketServer(new SocketServer(this)),
-        m_greeter(new Greeter(this)) {
-
-
-        // Save display server type
+    Display::DisplayServerType Display::defaultDisplayServerType()
+    {
         const QString &displayServerType = mainConfig.DisplayServer.get().toLower();
+        DisplayServerType ret;
         if (displayServerType == QStringLiteral("x11-user")) {
-            m_displayServerType = X11UserDisplayServerType;
-            m_terminalId = fetchAvailableVt();
+            ret = X11UserDisplayServerType;
         } else if (displayServerType == QStringLiteral("wayland")) {
-            m_displayServerType = WaylandDisplayServerType;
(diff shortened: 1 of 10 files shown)


src/daemon/Display.cpp
```diff
@@ -84,40 +84,45 @@ namespace SDDM {
...
-            m_displayServerType = WaylandDisplayServerType;
-            m_terminalId = fetchAvailableVt();
+        if (forceServerType.has_value()) {
+            qDebug() << "Forcing display server type" << *forceServerType;
+            m_displayServerType = *forceServerType;
         } else {
-            if (displayServerType != QLatin1String("x11")) {
-                qWarning("\"%s\" is an invalid value for General.DisplayServer: fall back to \"x11\"",
-                     qPrintable(displayServerType));
+            const QString &displayServerType = mainConfig.DisplayServer.get().toLower();
```
davidedmundson: If we move this code to seat, we can simplify it and get rid of optionals

i.e
```
Display::Display(Seat *parent, DisplayServerType serverType) {
  m_displayServerType = serverType
  .. rest of the code as is...

Seat::createDisplay() {
    const QString &displayServerType = mainConfig.DisplayServer.get().toLower();
    if (displayServerType == QStringLiteral("x11-user"))
       return createDisplay(X11UserDisplayServerType);
       
}

Seat::createDisplay(DisplayServerType serverType)
as-is

```

aleixpol: Not sure it's simpler, but there you have it.

src/daemon/Seat.h
```diff
@@ -33,9 +34,9 @@ namespace SDDM {
         explicit Seat(const QString &name, QObject *parent = 0);
 
         const QString &name() const;
+        void createDisplay(Display::DisplayServerType forceServerType);
```
davidedmundson: This arg can be just serverType now

davidedmundson (approved):

src/daemon/Seat.h
aleixpol: Addressed

PR #1614, opened 2022-11-16
x11: Also run xsessions on the greeter session

aleixpol:
It initializes things that seem necessary at times, like scripts to initialize how to load, like xrandr scripts.

To test, add a script in `/etc/X11/Xsessions.d/` and see it getting executed.


aleixpol:
Closing as this can be done with Xsetup too, with tighter control over what goes where.

PR #1615, opened 2022-11-17
Use github docker container

tiziodcaio:
Cleaner setup


plfiorini (changes requested):
Why not just:

```
runs-on: ubuntu-latest
```

The build script should run on future Ubuntu versions as well.

plfiorini:
Maybe the "Dependencies" step needs sudo in every line, can you see if that works?

plfiorini (approved):

aleixpol:
I'm not sure it's a good idea to be using an older version of the toolchain. I am not sure what this change brings right now.

PR #1616, opened 2022-11-17
Ask systemd for its environment and set it on our sessions

### LINKED ISSUE #1276: Import environment variable from systemd user session

aleixpol:
systemd1 has grown a way to specify environment variables to be used by the system, this change adds support for it.
https://www.freedesktop.org/software/systemd/man/environment.d.html

Fixes #1276

### DIFF
--- a/src/helper/UserSession.cpp
+++ b/src/helper/UserSession.cpp
@@ -27,16 +27,17 @@
 #include "VirtualTerminal.h"
 #include "XAuth.h"
 
-#include <sys/types.h>
-#include <sys/ioctl.h>
+#include <QDBusInterface>
+#include <QDBusPendingReply>
 #include <errno.h>
-#include <string.h>
-#include <unistd.h>
-#include <pwd.h>
-#include <grp.h>
-#include <unistd.h>
 #include <fcntl.h>
+#include <grp.h>
+#include <pwd.h>
 #include <sched.h>
+#include <string.h>
+#include <sys/ioctl.h>
+#include <sys/types.h>
+#include <unistd.h>
 #ifdef Q_OS_FREEBSD
 #include <login_cap.h>
 #endif
@@ -186,6 +187,16 @@ namespace SDDM {
 
             VirtualTerminal::jumpToVt(vtNumber, x11UserSession);
         }
+
+#ifdef HAVE_SYSTEMD
+        QDBusInterface systemd1Iface(
+            QStringLiteral("org.freedesktop.systemd1"),
+            QStringLiteral("/org/freedesktop/systemd1"),
+            QStringLiteral("org.freedesktop.DBus.Properties"),
+            QDBusConnection::systemBus(), this);
+
+        QDBusPendingReply<QVariant> environmentReply = systemd1Iface.asyncCall(QStringLiteral("Get"), QStringLiteral("org.freedesktop.systemd1.Manager"), QStringLiteral("Environment"));
+#endif
 
 #ifdef Q_OS_LINUX
         // enter Linux namespaces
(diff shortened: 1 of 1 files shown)


davidedmundson:
We had issues downstream in Plasma where we have basically this code.

StartPlasma::importSystemdSession

 - Log in, we'd set and export WAYLAND_DISPLAY to the systemd environment
 - Log out
 - Log in again, we import the env, now when kwin starts it sees WAYLAND_DISPLAY is already set, therefore it should run nested! 
 - Chaos ensues


We end up guarding through:
```
    // Check is variable is specific to session.
    return name == "DISPLAY" || name == "XAUTHORITY" || //
        name == "WAYLAND_DISPLAY" || name == "WAYLAND_SOCKET" || //
        name.startsWith("XDG_");
```

and ignoring these

2m:
It would be really nice to get this one merged. Recently Archlinux removed environment loading from `.pam_environment` and now `environment.d` is the preferred way to manage environment variables. I migrated my config to `environment.d` but noticed that it is not propagated to SDDM sessions.

Vogtinator (changes requested):
Shouldn't this ask the user instance instead of the system instance? Otherwise `~/.config/environment.d/*.conf` will be ignored.

aleixpol:
It's unclear to me if this PR is necessary. This is something we also have to do (and do) from Plasma, so I guess it's the DE's responsibility to bring these values. In fact, it's not very clear who is supposed to do this if at all.

Vogtinator:
> It's unclear to me if this PR is necessary. This is something we also have to do (and do) from Plasma, so I guess it's the DE's responsibility to bring these values. In fact, it's not very clear who is supposed to do this if at all.

IMO `pam_systemd` should take care of it, but it doesn't...

Jackenmen:
> It's unclear to me if this PR is necessary. This is something we also have to do (and do) from Plasma, so I guess it's the DE's responsibility to bring these values. In fact, it's not very clear who is supposed to do this if at all.

@aleixpol other popular login managers (namely GDM3 and LightDM) do it. I'm unsure where KDE does this but in my limited experience, the environment variables imported from the systemd user session are not present when opening a terminal app in KDE+SDDM but they are if you combine KDE with a different login manager. Consistently, running GNOME with SDDM similarly doesn't import variables.

Here's one semi-related issue: https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1640514#yui_3_10_3_1_1675424030620_2670

Vogtinator:
> I'm unsure where KDE does this but in my limited experience, the environment variables imported from the systemd user session are not present when opening a terminal app in KDE+SDDM

Here they are. Plasma explicitly imports those variables on start.

rhansen:
> IMO `pam_systemd` should take care of it, but it doesn't...

I agree completely.  I would much rather patch `pam_systemd` than see every display manager query systemd like this PR does.  Then ssh logins would benefit as well.  I opened a feature request: https://github.com/systemd/systemd/issues/35228

> other popular login managers (namely GDM3 and LightDM) do it.

@Jackenmen I don't see anything in the LightDM source code that imports the systemd environment variables.  Do you have a link to a commit or code snippet?

> > I'm unsure where KDE does this but in my limited experience, the environment variables imported from the systemd user session are not present when opening a terminal app in KDE+SDDM
> 
> Here they are. Plasma explicitly imports those variables on start.

@Vogtinator I assume you are referring to:

https://invent.kde.org/plasma/plasma-workspace/-/blob/d4bbb3dbd651923288760a501d4d6ee1b2b1090c/startkde/startplasma.cpp#L273-L290

Jackenmen:
> I don't see anything in the LightDM source code that imports the systemd environment variables. Do you have a link to a commit or code snippet?

Sorry, I wrote it in a way that suggested that LightDM implements this on its own but all it does is just include pam_systemd in the default PAM configuration:
https://github.com/canonical/lightdm/commit/2e31f1dcab907cf7af35a293e5f3c81216f0122a
I didn't actually look at how it's implemented before so I was just commenting on the observed behaviour, not the implementation.

If I add the same line to sddm's PAM configuration files I have in my system ((K)ubuntu 22.04.5), the env vars start to get set properly. To my surprise, on a different system that I have (openSUSE Tumbleweed), sddm actually manages to set env vars properly. I'm not entirely sure in what ways openSUSE's PAM configuration for SDDM deviates from the one that used to be provided by the project (and that Ubuntu's is heavily based on) but it definitely does.
Based on what I could find, it would seem that the SDDM project no longer provides its own PAM configuration (which Ubuntu's seems to be heavily based on) in newer versions (https://github.com/sddm/sddm/commit/ae072f901671b68861da9577e3e12e350a9053d5). This seems to suggest it is on the distributor to provide proper configuration and there's, in fact, nothing for SDDM to do.

PR #1641, opened 2023-01-26
Allow recovering from tty switching failures

### LINKED ISSUE #1636: Logout from KDE session on Rawhide goes to a console/black screen, not SDDM (with SDDM wayland)
*(Reported originally downstream in Fedora as [rhbz#2110801](https://bugzilla.redhat.com/show_bug.cgi?id=2110801) by @AdamWill)*

If you do a clean install of current Fedora Rawhide KDE - e.g. from https://kojipkgs.fedoraproject.org/compose/rawhide/Fedora-Rawhide-20230105.n.0/compose/Spins/x86_64/iso/Fedora-KDE-Live-x86_64-Rawhide-20230105.n.0.iso - then boot the installed system, log in, and log out, you wind up at a console on tty2, not back at SDDM as you should.

As best I can tell, when the system boots, SDDM is running on tty1 and there are consoles on tty2 and tty3. After login, SDDM goes away and tty1 just shows a flashing cursor; KDE starts on tty3. tty2 still has a console. When logging out from KDE, the system puts you on tty2, with the console. tty1 still shows a flashing cursor. tty3 also now has a console again. SDDM doesn't seem to be present on any tty, though it's still *running* according to systemctl and ps.

Another user sees a variation of this problem where they get a black screen with just a flashing text cursor in the top left when logging out. That is, it doesn't switch to tty2, it just sticks to tty1 instead.

This problem shows up when using SDDM in Wayland mode. When using (root) X11, this doesn't occur.

Fedora KDE Rawhide nightlies can be found here: https://openqa.fedoraproject.org/nightlies.html

aleixpol:
If we fail to switch tty when showing the greeter, exit saying so then switch back to tty 1 and try launching the Display again.

Fixes #1636

~~Draft because:~~
- ~~we should avoid looping infinitely.~~ works
- ~~we should do something similar for autologin use cases.~~ works
- ~~we should do something similar for user switching use cases.~~ https://github.com/sddm/sddm/issues/1642

### DIFF
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -51,9 +51,10 @@
 #include "Login1Session.h"
 #include "VirtualTerminal.h"
 #include "WaylandDisplayServer.h"
+#include "config.h"
 
 static int s_ttyFailures = 0;
-
+#define STRINGIFY(x) #x
 
 namespace SDDM {
     bool isTtyInUse(const QString &desiredTty) {
@@ -76,11 +77,12 @@ namespace SDDM {
     }
 
     int fetchAvailableVt() {
+        if (!isTtyInUse(QStringLiteral("tty" STRINGIFY(SDDM_INITIAL_VT)))) {
+            return SDDM_INITIAL_VT;
+        }
         const auto vt = VirtualTerminal::currentVt();
-        if (vt > 0) {
-            if (!isTtyInUse(QStringLiteral("tty%1").arg(vt))) {
-                return vt;
-            }
+        if (vt > 0 && !isTtyInUse(QStringLiteral("tty%1").arg(vt))) {
+            return vt;
         }
         return VirtualTerminal::setUpNewVt();
     }
@@ -159,7 +161,7 @@ namespace SDDM {
             }
             // It might be the case that we are trying a tty that has been taken over by a
             // different process. In such a case, switch back to the initial one and try again.
-            VirtualTerminal::jumpToVt(1, true);
+            VirtualTerminal::jumpToVt(SDDM_INITIAL_VT, true);
             stop();
         });
         connect(m_greeter, &Greeter::displayServerFailed, this, &Display::displayServerFailed);
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -52,6 +52,8 @@
 #include "VirtualTerminal.h"
 #include "WaylandDisplayServer.h"
 
+static int s_ttyFailures = 0;
+
 
 namespace SDDM {
     bool isTtyInUse(const QString &desiredTty) {
@@ -149,10 +151,17 @@ namespace SDDM {
         connect(this, &Display::loginFailed, m_socketServer, &SocketServer::loginFailed);
         connect(this, &Display::loginSucceeded, m_socketServer, &SocketServer::loginSucceeded);
 
-        connect(m_greeter, &Greeter::failed,
-                QCoreApplication::instance(), [] {
-                    QCoreApplication::instance()->exit(23);
-                });
+        connect(m_greeter, &Greeter::failed, this, &Display::stop);
+        connect(m_greeter, &Greeter::ttyFailed, this, [this] {
+            ++s_ttyFailures;
+            if (s_ttyFailures > 5) {
+                QCoreApplication::exit(23);
+            }
+            // It might be the case that we are trying a tty that has been taken over by a
+            // different process. In such a case, switch back to the initial one and try again.
+            VirtualTerminal::jumpToVt(1, true);
+            stop();
+        });
         connect(m_greeter, &Greeter::displayServerFailed, this, &Display::displayServerFailed);
     }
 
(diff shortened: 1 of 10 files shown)


alebastr:
This seems to be working fine in a sane configuration, thanks!

If `ReserveVT=1` is set in `logind.conf` though (as configured in [openSUSE's gdm package](https://build.opensuse.org/package/view_file/openSUSE:Factory/gdm/reserveVT.conf?expand=1)),  the initial boot just quietly fails and subsequent restarts of sddm will either succeed or stuck in an infinite loop.

src/daemon/Display.cpp
```diff
@@ -149,10 +149,11 @@ namespace SDDM {
...
         connect(this, &Display::loginSucceeded, m_socketServer, &SocketServer::loginSucceeded);
 
-        connect(m_greeter, &Greeter::failed,
-                QCoreApplication::instance(), [] {
-                    QCoreApplication::instance()->exit(23);
-                });
+        connect(m_greeter, &Greeter::failed, this, &Display::stop);
+        connect(m_greeter, &Greeter::ttyFailed, this, [this] {
+            VirtualTerminal::jumpToVt(1, true);
+            stop();
```
Vogtinator, line 165: IMO needs a comment here

src/daemon/Display.cpp
```diff
@@ -149,10 +149,11 @@ namespace SDDM {
         connect(this, &Display::loginFailed, m_socketServer, &SocketServer::loginFailed);
         connect(this, &Display::loginSucceeded, m_socketServer, &SocketServer::loginSucceeded);
 
-        connect(m_greeter, &Greeter::failed,
-                QCoreApplication::instance(), [] {
-                    QCoreApplication::instance()->exit(23);
-                });
+        connect(m_greeter, &Greeter::failed, this, &Display::stop);
+        connect(m_greeter, &Greeter::ttyFailed, this, [this] {
+            VirtualTerminal::jumpToVt(1, true);
```
Vogtinator: Why tty1? In openSUSE we have a getty running there, this might break it

aleixpol: getty is everywhere in my experience. Do you have an alternative in mind?

aleixpol: Would you prefer having a default tty setting?

aleixpol:
> If `ReserveVT=1` is set in `logind.conf` though (as configured in [openSUSE's gdm package](https://build.opensuse.org/package/view_file/openSUSE:Factory/gdm/reserveVT.conf?expand=1)), the initial boot just quietly fails and subsequent restarts of sddm will either succeed or stuck in an infinite loop.

Right, it sounds like something that shouldn't be set for sddm's package. Or is it something you'd like to have?

src/daemon/Display.cpp
arrowd: And on FreeBSD we have getty running on ttys `0-7` and X server starts on `8`.

Vogtinator: On openSUSE it's getty on tty1-tty6 by default (started on the first switch) and X on tty7 (sddm.conf has `MinimumVT=7`).

Because getty is started on demand, tty1-tty6 are actually free until the switch and something like OPENQRY can return them for use by e.g. `kwin_wayland`.

It's not entirely clear to me what the VT number specified here effectively does...

aleixpol: > On openSUSE it's getty on tty1-tty6 by default (started on the first switch) and X on tty7 (sddm.conf has MinimumVT=7).

MinimumVT was dropped a while ago. https://github.com/sddm/sddm/commit/f0b6dec66a6fc8cc6e21da57f81190843b483f76

> Because getty is started on demand, tty1-tty6 are actually free until the switch and something like OPENQRY can return them for use by e.g. kwin_wayland.

Well that kind of is the problem. We sometimes want to have a predictable tty for our session and the greeter. openqry can tell us which one is free but getty takes them over and never releases them so you can easily end up with a session that is all over the place.

> It's not entirely clear to me what the VT number specified here effectively does...

Where you want the graphical sessions to start from. I guess not unlike the MinimumVT setting from back then.

Vogtinator: Yep, we'll definitely need a replacement for that. IMO it's important to have this deterministic and predictable. systemd upstream is unfortunately opposed to that...

aleixpol: > In openSUSE we have a getty running there, this might break it

Unless you were patching out this `Conflicts=getty@tty1.service` in the service file, this statement is not true. If you were patching it, I imagine it was conflicting at times?

In any case, I've added a CMake variable so that you can use start from a different tty.

Vogtinator: > Unless you were patching out this Conflicts=getty@tty1.service in the service file, this statement is not true. If you were patching it, I imagine it was conflicting at times?

Yep (https://build.opensuse.org/package/view_file/openSUSE:Factory/sddm/0001-Systemd-service-unit-Use-tty7-by-default.patch?expand=1)! Though we don't actually use `sddm.service` by default, we have `display-manager.service` which starts the displaymanager (configured in `/etc/sysconfig/displaymanager) itself. That'll probably change soon though.

davidedmundson (approved):

PR #1644, opened 2023-02-01
Fix user switching

### LINKED ISSUE #1642: User Switching workflow failure with SDDM on Wayland mode
It's a bit intricate because it's not entirely true that it's not working. It works when:

- user A: from the DE (e.g. Plasma) you ask to switch to a new user.
- you enter user B, ask for the same as the line before
- SDDM lists both A and B, log into either
- :bomb:

aleixpol:
Make sure we don't start a new session if we are switching. Update the terminalId (i.e. tty) Display property as we switch from the greeter to the actual session, as they might be on different terminals now.

~~WIP because we rely on Seat::displayStopped to choose the correct tty to jump to.~~

Fixes #1642

### DIFF
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -67,7 +67,7 @@ namespace SDDM {
             for(const SessionInfo &s : info) {
                 OrgFreedesktopLogin1SessionInterface session(Logind::serviceName(), s.sessionPath.path(), QDBusConnection::systemBus());
                 if (desiredTty == session.tTY() && session.state() != QLatin1String("closing")) {
-                    qDebug() << "tty" << desiredTty << "already in use by" << session.user().path << session.state()
+                    qDebug() << "tty" << desiredTty << "already in use by" << session.user().path.path() << session.state()
                                       << session.display() << session.desktop() << session.vTNr();
                     return true;
                 }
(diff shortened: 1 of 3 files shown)


davidedmundson (approved):

davidedmundson (approved):

src/daemon/Display.cpp
```diff
@@ -428,7 +427,7 @@ namespace SDDM {
         env.insert(QStringLiteral("XDG_SESSION_CLASS"), QStringLiteral("user"));
         env.insert(QStringLiteral("XDG_SESSION_TYPE"), session.xdgSessionType());
         env.insert(QStringLiteral("XDG_SEAT"), seat()->name());
-        env.insert(QStringLiteral("XDG_VTNR"), QString::number(terminalNewSession));
+        env.insert(QStringLiteral("XDG_VTNR"), QString::number(m_terminalId));
```
davidedmundson: we don't want to lose track of which VT X11 is running on, they are separate things

aleixpol: Addressed.

smirkybg:
Not sure what you guys are fixing, but I suppose it's not my case.
If I login to user A, then choose "Switch User" from Plasma 5's lockscreen, I get to the SDDM session just fine.
Then if I login to user B, then logout, I get a black screen with a flashing cursor. I can switch the TTY manually, but this just doesn't work for me for years.

Using Wayland with Plasma 5 and sddm-git.

PR #1646, opened 2023-02-02
ThemeConfig: Provide typed accessors to values

### LINKED ISSUE #1097: Pull booleans from theme.conf into themes QML as such

Vogtinator:
Some methods are added to interpret theme.conf settings as bool, int, real and string explicitly. Previously, most basic types were returned as string and needed manual handling in QML code.

To allow exporting those methods to QML, it was necessary to port the class from QVariantMap over to QQmlPropertyMap. A single method had to be added to make it API compatible with the rest of the code.

Small breaking change: This returns more members when used with Object.keys, e.g. the added methods and QObject's objectName.

Fixes #1097

Test:

theme.conf:

```
[General]
showlogo=hidden
logo=/usr/share/sddm/themes/breeze/default-logo.svg
type=image
color=#1d99f3
fontSize=10
background=components/artwork/1920x1080.jpg
something=007
```

Main.qml:

```
console.log(Object.keys(config))
console.log(config.something)
console.log(config.boolValue("something"))
console.log(config.stringValue("something"))
console.log(config.intValue("something"))
console.log(config.realValue("something"))
```

Output:

```
qml: [objectName,background,color,fontSize,logo,showlogo,something,type,defaultBackground,alwaysShowClock,objectNameChanged,valueChanged,keys,boolValue,intValue,realValue,stringValue,__0,__1,__2,__3,__4,__5,__6,__7,__8]
qml: 007
qml: true
qml: 007
qml: 7
qml: 7
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,7 @@ jobs:
           sudo sed -i '/^#\sdeb-src /s/^#//' "/etc/apt/sources.list"
           sudo apt-get update -y
           DEBIAN_FRONTEND=noninteractive sudo apt-get build-dep sddm -y
-          DEBIAN_FRONTEND=noninteractive sudo apt-get install clang clazy sudo -y
+          DEBIAN_FRONTEND=noninteractive sudo apt-get install clang clazy sudo qml-module-qttest -y
       - name: Build
         run: |
           set -x
@@ -49,3 +49,8 @@ jobs:
             -DENABLE_PAM:BOOL=${{ matrix.pam }}
           make -j $(getconf _NPROCESSORS_ONLN)
           sudo make install
+      - name: Test
+        env:
+          CTEST_OUTPUT_ON_FAILURE: "1"
+        working-directory: build
+        run: make test
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,7 +99,7 @@ find_package(XCB REQUIRED)
 find_package(XKB REQUIRED)
 
 # Qt 5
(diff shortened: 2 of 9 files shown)


aleixpol:
+1 This looks useful.

Since you already seem to have the test prepared, would it be possible to include it for us to run on the CI?

Vogtinator:
> +1 This looks useful.
> 
> Since you already seem to have the test prepared, would it be possible to include it for us to run on the CI?

Done. I used `Qt5::QuickTest`, never tried that before.

Vogtinator:
The CI didn't actually run `make test`, I added that as well.

herostrat (approved):
Looks good! (great that you included the test runs in the CI)

aleixpol (approved):

PR #1653, opened 2023-02-04
Fix terminal logging

Vogtinator:
Previously running `sddm` in a terminal produced no output whatsoever. With these changes it writes to `stderr` if it's a tty.

I'd prefer if we could just use Qt's message handler directly without replacing it with one of our own. It already logs to the journal properly if it's not run from a terminal. That's unfortunately not as easy though because:

* Qt might've been built without journal integration
* Logging to sddm.log is still needed in some cases
* SDDM doesn't use logging categories, only a `DAEMON:`/`HELPER:`/... prefix, which is currently set by the custom message handlers.

### DIFF
--- a/src/common/MessageHandler.h
+++ b/src/common/MessageHandler.h
@@ -71,27 +71,27 @@ namespace SDDM {
     static void standardLogger(QtMsgType type, const QString &msg) {
         static QFile file(QStringLiteral(LOG_FILE));
 
-        // try to open file only if it's not already open
-        if (!file.isOpen()) {
+        // Try to open the log file if we're not outputting to a terminal
+        if (!file.isOpen() && !isatty(STDERR_FILENO)) {
             if (!file.open(QFile::Append | QFile::WriteOnly))
                 file.open(QFile::Truncate | QFile::WriteOnly);
-        }
 
-        // If we can't open the file, create it in a writable location
-        // It will look spmething like ~/.local/share/$appname/sddm.log
-        // or for the sddm user /var/lib/sddm/.local/share/$appname/sddm.log
-        if (!file.isOpen()) {
-            QDir().mkpath(QStandardPaths::writableLocation(QStandardPaths::AppDataLocation));
-            file.setFileName(QStandardPaths::writableLocation(QStandardPaths::AppDataLocation) + QLatin1String("/sddm.log"));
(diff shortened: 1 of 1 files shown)


davidedmundson:
>Qt might've been built without journal integration

Even if it's not, systemd will just redirect any stderr for services.

src/common/MessageHandler.h
```diff
@@ -71,27 +71,27 @@ namespace SDDM {
     static void standardLogger(QtMsgType type, const QString &msg) {
         static QFile file(QStringLiteral(LOG_FILE));
 
-        // try to open file only if it's not already open
-        if (!file.isOpen()) {
+        // Try to open the log file if we're not outputting to a terminal
+        if(!file.isOpen() && !isatty(STDERR_FILENO)) {
```
aleixpol: you dropped a space after the if there.

aleixpol (commented):
Pointed to a small change, looks good although I'm not sure it really fixes anything other than running on the tty which is something we never (?) do.

src/common/MessageHandler.h
Vogtinator: Fixed

Vogtinator:
> Pointed to a small change, looks good although I'm not sure it really fixes anything other than running on the tty which is something we never (?) do.

Correct. Personally I find running from the terminal (e.g. ssh) easier than starting a service, stopping it and then looking through the journal or log files. There's also the `--test-mode` that was supposed to help testing, but it doesn't really work well unfortunately.

aleixpol (approved):

a17r:
MessageHandler.h:75:39: error: ‘STDERR_FILENO’ was not declared in this scope
src/common/MessageHandler.h:75:32: error: ‘isatty’ was not declared in this scope

Missing `#include <unistd.h>`?

Vogtinator:
> MessageHandler.h:75:39: error: ‘STDERR_FILENO’ was not declared in this scope src/common/MessageHandler.h:75:32: error: ‘isatty’ was not declared in this scope
> 
> Missing `#include <unistd.h>`?

Almost: Now it also needs that include in the `!HAVE_JOURNALD` case. I'll fix it.

Edit: https://github.com/sddm/sddm/pull/1661

PR #1661, opened 2023-02-07
Fix missing unistd.h include when building without journald support

Vogtinator:


aleixpol (approved):
Seems like we could get an extra case on the CI, to see we are building with the different configs.

arrowd:
This patch is also part of my PR: https://github.com/sddm/sddm/pull/1662

Vogtinator:
#1662 is more complete, let's continue there instead

PR #1663, opened 2023-02-09
Fix compiler warnings

herostrat:
SSIA

### DIFF
--- a/src/common/VirtualTerminal.cpp
+++ b/src/common/VirtualTerminal.cpp
@@ -210,7 +210,9 @@ out:
 
                 // Clear VT
                 static const char *clearEscapeSequence = "\33[H\33[2J";
-                write(vtFd, clearEscapeSequence, sizeof(clearEscapeSequence));
+                if (write(vtFd, clearEscapeSequence, sizeof(clearEscapeSequence)) == -1) {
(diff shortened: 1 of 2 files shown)


src/common/VirtualTerminal.cpp
```diff
@@ -210,7 +210,9 @@ namespace SDDM {
 
                 // Clear VT
                 static const char *clearEscapeSequence = "\33[H\33[2J";
-                write(vtFd, clearEscapeSequence, sizeof(clearEscapeSequence));
+                if (write(vtFd, clearEscapeSequence, sizeof(clearEscapeSequence)) == -1) {
```
Vogtinator, line 213: Technically checking for success requires `== sizeof(clearEscapeSequence)`, but then errno isn't set and that's even more annoying. So IMO this is fine

Vogtinator (approved):

src/common/VirtualTerminal.cpp
herostrat: You're right, I wasn't sure if it is worth the effort.
I can change it if you prefer :)

Vogtinator: No need!

aleixpol (approved):

herostrat:
Thanks!

PR #1664, opened 2023-02-09
Activate -Wextra and fix warnings: Part 1 GCC

herostrat:
First two commits that activate some warnings included in Wextra.
Let me know what you think. I think the fallthrough were indeed buggy?

### DIFF
--- a/src/greeter/ScreenModel.cpp
+++ b/src/greeter/ScreenModel.cpp
@@ -57,8 +57,8 @@ namespace SDDM {
         return QRect(QPoint(0, 0), d->screen->geometry().size());
     }
 
-    int ScreenModel::rowCount([[maybe_unused]] const QModelIndex &parent) const {
-        return 1;
+    int ScreenModel::rowCount(const QModelIndex &parent) const {
+        return parent.isValid() ? 0 : 1;
     }
 
     QVariant ScreenModel::data(const QModelIndex &index, int role) const {
--- a/src/greeter/ScreenModel.cpp
+++ b/src/greeter/ScreenModel.cpp
@@ -57,7 +57,7 @@ namespace SDDM {
         return QRect(QPoint(0, 0), d->screen->geometry().size());
     }
 
-    int ScreenModel::rowCount(const QModelIndex &parent) const {
+    int ScreenModel::rowCount([[maybe_unused]] const QModelIndex &parent) const {
         return 1;
     }
 
--- a/src/greeter/SessionModel.cpp
+++ b/src/greeter/SessionModel.cpp
@@ -85,7 +85,7 @@ namespace SDDM {
         return roleNames;
     }
 
-    const int SessionModel::lastIndex() const {
+    int SessionModel::lastIndex() const {
         return d->lastIndex;
     }
 
--- a/src/greeter/SessionModel.cpp
+++ b/src/greeter/SessionModel.cpp
@@ -53,7 +53,7 @@ namespace SDDM {
 
         // refresh everytime a file is changed, added or removed
         QFileSystemWatcher *watcher = new QFileSystemWatcher(this);
-        connect(watcher, &QFileSystemWatcher::directoryChanged, [this](const QString &path) {
+        connect(watcher, &QFileSystemWatcher::directoryChanged, [this]([[maybe_unused]] const QString &path) {
             // Recheck for flag to show Wayland sessions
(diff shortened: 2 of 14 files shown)


Vogtinator:
Yeah, that looks like proper bugfixes indeed. Especially the `LoginFailed` one, which fell through to process an info message. I wonder whether this triggered some delay. Did you try sddm with these changes applied? If so, does it behave differently on login failure now?

aleixpol (approved):

herostrat:
No, I did not test it in my system, just compile tested.
I'm not totally sure (yet) how I can change the original binary with the development version and the other way around. I don't want to mess up my system :/

herostrat:
I added another fix for warnings and rebased on master.
I will not any new fixes in this PR but open additional ones in the future.
imho this is now ready to merge.

Edit: Forget what I said, the remaining errors were only few and from type Wmissing-field-initializers
I also activated Werror but I can revert if you prefer.

Edit2: Ok I see additional warning with llvm. I drop Werror.

aleixpol:
So did you manage to test it after all? do you need us to do so?

herostrat:
If you could help me out here I would be very grateful.
I cannot get it to run in my VM (I am to stupid to change the sddm to my compiled version) and I am a bit afraid to try it on my production machine.
Sorry, I am a bit lost with system programs like this -.-

a17r:
fwiw I have been running sddm with this patch applied for a few days without issues. (no obvious ones, anyway)

src/greeter/ScreenModel.cpp
```diff
@@ -57,7 +57,7 @@ namespace SDDM {
         return QRect(QPoint(0, 0), d->screen->geometry().size());
     }
 
-    int ScreenModel::rowCount(const QModelIndex &parent) const {
+    int ScreenModel::rowCount([[maybe_unused]] const QModelIndex &parent) const {
```
Vogtinator: This should actually do

```return parent.isValid() ? 0 : 1;```

Otherwise it represents an infinite tree model

File: src/greeter/SessionModel.cpp [hunk 2]
```diff
@@ -53,7 +53,7 @@ namespace SDDM {
 
         // refresh everytime a file is changed, added or removed
         QFileSystemWatcher *watcher = new QFileSystemWatcher(this);
-        connect(watcher, &QFileSystemWatcher::directoryChanged, [this](const QString &path) {
```
Vogtinator, line 56: Just remove the parameter?

src/greeter/SessionModel.cpp
```diff
@@ -89,7 +89,7 @@ namespace SDDM {
         return d->lastIndex;
     }
 
-    int SessionModel::rowCount(const QModelIndex &parent) const {
```
Vogtinator, line 92: Like above, should do `return parent.isValid() ? 0 : d->sessions.length();`

src/greeter/UserModel.cpp
```diff
@@ -178,7 +178,7 @@ namespace SDDM {
         return stateConfig.Last.User.get();
     }
 
-    int UserModel::rowCount(const QModelIndex &parent) const {
+    int UserModel::rowCount([[maybe_unused]] const QModelIndex &parent) const {
```
Vogtinator: Like above

src/greeter/waylandkeyboardbackend.cpp
```diff
@@ -43,7 +43,7 @@ WaylandKeyboardBackend::~WaylandKeyboardBackend()
 {
 }
 
-QList<QObject *> parseRules(const QString &filename, int &current)
+QList<QObject *> parseRules(const QString &filename, [[maybe_unused]] int &current)
```
Vogtinator: Looks like a genuine bug. The current layout is not set by the wayland backend?

herostrat: I am not sure, this is a bit above my paygrade.
Any idea how to fix this?

Vogtinator: No idea. There is no compositor agnostic API for this. Maybe just

```
// FIXME
current = 0;
```

herostrat:
Fixed all findings and got a VM up and running to test it. For what it is worth, I could not find any problems.

Vogtinator (approved):

PR #1670, opened 2023-02-20
CI: Add RPM distributions and use containers

Conan-Kudo:
This will ensure we have a reasonable ability to catch issues that affect Fedora/CentOS/RHEL/openSUSE before it gets shipped in a release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,23 +17,42 @@ jobs:
   build:
     if: "!contains(github.event.head_commit.message, 'ci skip')"
     strategy:
+      fail-fast: false
       matrix:
+        container-image:
+          - "quay.io/fedora/fedora:rawhide"
+          - "quay.io/centos/centos:stream9"
+          - "registry.opensuse.org/opensuse/tumbleweed-dnf"
+          - "docker.io/library/ubuntu:22.04"
         compiler:
           - gcc
           - clang
         pam:
           - ON
           - OFF
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
+
+    container:
+      image: ${{ matrix.container-image }}
+
     steps:
       - uses: actions/checkout@v3
-      - name: Dependencies
+      - name: Dependencies (rpm-type)
+        if: ${{ contains(matrix.container-image, 'fedora') || contains(matrix.container-image, 'centos') || contains(matrix.container-image, 'opensuse') }}
         run: |
           set -x
-          sudo sed -i '/^#\sdeb-src /s/^#//' "/etc/apt/sources.list"
-          sudo apt-get update -y
-          DEBIAN_FRONTEND=noninteractive sudo apt-get build-dep sddm -y
-          DEBIAN_FRONTEND=noninteractive sudo apt-get install clang clazy sudo qml-module-qttest -y
+          dnf --assumeyes install dnf-plugins-core
+          if [[ "${{ matrix.container-image }}" =~ "centos" ]]; then dnf --assumeyes config-manager --set-enabled crb && dnf --assumeyes install epel-release; fi
+          dnf --assumeyes builddep sddm
+          dnf --assumeyes install clang clazy
+      - name: Dependencies (deb-type)
+        if: ${{ contains(matrix.container-image, 'ubuntu') }}
+        run: |
+          set -x
+          sed -i '/^#\sdeb-src /s/^#//' "/etc/apt/sources.list"
+          apt-get update -y
+          DEBIAN_FRONTEND=noninteractive apt-get build-dep sddm -y
+          DEBIAN_FRONTEND=noninteractive apt-get install clang clazy qml-module-qttest -y
       - name: Build
         run: |
           set -x
@@ -44,11 +63,17 @@ jobs:
             #export CXX=clang++
             export CXX=clazy
           fi
+          if [ -f "/usr/etc/login.defs" ]; then
+            export LOGIN_DEFS_PATH="/usr/etc/login.defs"
+          else
+            export LOGIN_DEFS_PATH="/etc/login.defs"
+          fi
           cmake .. \
             -DBUILD_MAN_PAGES:BOOL=ON \
-            -DENABLE_PAM:BOOL=${{ matrix.pam }}
+            -DENABLE_PAM:BOOL=${{ matrix.pam }} \
+            -DLOGIN_DEFS_PATH:PATH="${LOGIN_DEFS_PATH}"
           make -j $(getconf _NPROCESSORS_ONLN)
-          sudo make install
+          make install
       - name: Test
         env:
           CTEST_OUTPUT_ON_FAILURE: "1"


herostrat:
How difficult would it be to add a BSD, e.g. FreeBSD?
Even just compile testing would be very beneficial I think.

marcdeop:
> How difficult would it be to add a BSD, e.g. FreeBSD? Even just compile testing would be very beneficial I think.

FreeBSD is not officially supported by Github Actions as of now. See [request](https://github.com/actions/runner/issues/385) (here possible solutions in the discussion but... well, they are not the best ;--)

I would keep FreeBSD out for now (at least not in this PR).

.github/workflows/build.yml
```diff
@@ -17,23 +17,41 @@ jobs:
...
       - uses: actions/checkout@v3
-      - name: Dependencies
+      - name: Dependencies (rpm-type)
+        run: |
+          set -x
+          dnf --assumeyes install dnf-plugins-core
+          if [ -f /etc/centos-release ]; then dnf --assumeyes config-manager --set-enabled crb && dnf --assumeyes install epel-release; fi
+          dnf --assumeyes builddep sddm
+          dnf --assumeyes install clang clazy
+        if: ${{ contains(matrix.container-image, 'fedora') || contains(matrix.container-image, 'centos') }}
```
Vogtinator: IMO the `if` should be above `run`.

Vogtinator:
Could you add `registry.opensuse.org/opensuse/tumbleweed` as well? I could open a separate PR but it would depend on this anyway.

Vogtinator:
IMO we don't need the full matrix of clang/gcc and USE_PAM OFF/ON for every OS.

.github/workflows/build.yml
Conan-Kudo: Done.

Conan-Kudo:
> Could you add `registry.opensuse.org/opensuse/tumbleweed` as well? I could open a separate PR but it would depend on this anyway.

I've added `tumbleweed-dnf` since that would work with the steps we have here.

Conan-Kudo:
@Vogtinator why is `curl`, `tar`, and `gzip` missing from the openSUSE containers? Without them, GitHub Actions can't download the sources into the environment...

Conan-Kudo:
Anyway, made an SR to fix that for the Tumbleweed DNF image: https://build.opensuse.org/request/show/1066820

Vogtinator:
> @Vogtinator why is curl, tar, and gzip missing from the openSUSE containers? Without them, GitHub Actions can't download the sources into the environment...

They're meant to be minimal. For actually using the containers, you'll have to install some packages anyway. This is true here as well: Just move the step for installing dependencies before the checkout action and install curl etc. as well.

aleixpol:
Can somebody explain to me how it knows to do the for deb/rpm accordingly?

marcdeop:
> Can somebody explain to me how it knows to do the for deb/rpm accordingly?

It's based on the matrix and the `if`s:
```
if: ${{ contains(matrix.container-image, 'ubuntu') }}
```

In the original submission, those `if`s where at the end of the array which made readability bad (my *opinion*). This has been fixed now.

.github/workflows/build.yml
```diff
@@ -36,6 +36,12 @@ jobs:
       image: ${{ matrix.container-image }}
 
     steps:
+      - name: GitHub Actions dependencies (openSUSE only)
```
Vogtinator: You could just move `- uses: actions/checkout@v3` below the dependency installation. Then there is only one step for deps and it does not need to be temporary.

BTW, you need only curl, tar and gzip here.

Conan-Kudo: I'm deliberately keeping it separate since I expect it to be removed once the containers are updated.

Vogtinator: What's so bad about just moving one step?

marcdeop: I guess because then it can be removed just from here.

marcdeop:
Is there anything really blocking this from being merged?

Vogtinator:
> Is there anything really blocking this from being merged?

The builds should all succeed

Conan-Kudo:
Yup, and I'm working on that.

marcdeop:
> Yup, and I'm working on that.

We could also comment out/remove the failing builds and leave them for another iteration.

You  know, baby steps :-)

Conan-Kudo:
> > Yup, and I'm working on that.
> 
> We could also comment out/remove the failing builds and leave them for another iteration.
> 
> You know, baby steps :-)

Meh. I know what's wrong and how to fix them.

Conan-Kudo:
This is finally now good to land!

.github/workflows/build.yml
```diff
@@ -17,38 +17,62 @@ jobs:
...
+          set -x
+          sed -i '/^#\sdeb-src /s/^#//' "/etc/apt/sources.list"
+          apt-get update -y
+          DEBIAN_FRONTEND=noninteractive apt-get build-dep sddm -y
+          DEBIAN_FRONTEND=noninteractive apt-get install clang clazy qml-module-qttest -y
       - name: Build
         run: |
           set -x
           mkdir -p build
-          cd build
```
aleixpol: Why are we changing from using `cd` to use weird arguments for `cmake` and `make`?

Conan-Kudo: It more closely mimics how we actually build sddm downstream in distributions where those methods are used. I figure given that it looks more like how we build it and is slightly faster with less commands, it should be fine.

aleixpol: It's considerably harder to read and it's not how anyone would do it on their system. I'd prefer if we keep it like it was.

I'd suggest you change how you do it in the packaging too. :)

Conan-Kudo: All distributions wrap CMake invocations through some interface. For RPM packages, we have one-liner macros:

```
%cmake
%cmake_build
%cmake_install
```

We're probably not going to change how those macros are implemented because it also makes it easy to twiddle to Ninja or something else without having to do anything too special. It also guarantees that they build out of tree.

(I could also technically drop the `mkdir -p build` command here too, since CMake will create it for me this way...)

But this isn't a hill I'll die on, so I'll roll it back...

.github/workflows/build.yml
```diff
@@ -44,11 +63,17 @@ jobs:
             #export CXX=clang++
             export CXX=clazy
           fi
+          if [[ "${{ matrix.container-image }}" =~ "opensuse" ]]; then
```
Vogtinator: Instead of hardcoding distros, check for file existence which is what ultimately matters anyway:

```
export LOGIN_DEFS_PATH=/etc/login.defs
[ -e "$LOGIN_DEFS_PATH" ] || export LOGIN_DEFS_PATH=/usr/etc/login.defs
```

marcdeop: I find it easier to read how it's currently implemented. This way it's clear it's Opensuse the  one that is different and if they ever change/fix it we will notice here and we will be able to remove unneded code

Vogtinator: The move to `/usr/etc` is intentional and I expect that at least some other distros follow.

Conan-Kudo: So far, nobody else uses `/usr/etc` for this purpose. Other distributions have been moving files to different locations in `/usr`. I'm also skeptical of anyone else using `/usr/etc` since `/etc` is such a bad path for config files anyway.

Vogtinator: :shrug: If they want to diverge, sure...

Conan-Kudo: Nobody is doing `/usr/etc` except openSUSE. In this regard, openSUSE is the divergence. Everyone else is moving things to either `/usr/share` or some other more appropriate location.

Vogtinator: There was some cross-distro communication but nobody else cared so we went ahead with `/usr/etc`. If someone is moving it to another location after we already use `/usr/etc` for some time now, it's not us causing the divergence.

Conan-Kudo: I could also just touch `/etc/login.defs` and make this work that way universally too. SDDM doesn't know how to deal with `/usr/etc` and `/etc` layering anyway.

Vogtinator: SDDM doesn't actually use `login.defs` during runtime at all, it's just used during buildtime to figure out the default user ID range.

Vogtinator: In any case, my main argument here is that checking for distro names is as bad as checking user agents in web pages. The correct way is to perform feature detection.

Conan-Kudo: Fine, I don't care enough. Fixed.

Vogtinator:
The only remaining topic then is the build matrix. IMO {fedora,centos,tumbleweed,ubuntu}-{clang,gcc}-{pam,no-pam} is a bit too much.

What about disabling PAM OFF builds on most of them? (I wonder what the code is for anyway)

Conan-Kudo:
That actually _adds_ complexity to the pipeline logic. I deliberately didn't bother because it's more work to make that possible.

Vogtinator:
> That actually _adds_ complexity to the pipeline logic. I deliberately didn't bother because it's more work to make that possible.

Might be as simple as this:

```
        container-image:
          - "quay.io/fedora/fedora:rawhide"
          - "quay.io/centos/centos:stream9"
          - "registry.opensuse.org/opensuse/tumbleweed-dnf"
          - "docker.io/library/ubuntu:22.04"
        compiler:
          - gcc
          - clang
        pam:
          - ON
        include:
          - container-image: "docker.io/library/ubuntu:22.04"
            pam: OFF
```

(Somehow I hit Ctrl-Return instead of Shift-Return, that is apparently the shortcut for Comment and close. TIL.)

Conan-Kudo:
Who uses SDDM without PAM, by the way? Even Slackware adopted PAM in 2020: https://alien.slackbook.org/blog/slackware-introduces-pam-into-its-core/

Vogtinator:
Yeah, maybe we can just drop that. I doubt that there is any practical relevance and I don't think it's useful for test/dev setups either.

.github/workflows/build.yml
```diff
@@ -44,11 +60,16 @@ jobs:
             #export CXX=clang++
```
marcdeop, line 63: We can remove this, can we not?

Conan-Kudo: I treat that as a comment note that `clang++` is the default way to do it, since we're using `clazy` instead.

Conan-Kudo:
@Vogtinator I've added a commit to drop the PAM build modes.

Vogtinator:
IMO that should be done together with removing the code itself, which is something for a separate PR. But maybe the other maintainers don't mind? @aleixpol @davidedmundson @plfiorini

Conan-Kudo:
Then do you want me to drop that commit and we can merge this, or what?

Vogtinator:
As long as the PAM code is there, it should be built tested, but IMO not 8 times.

aleixpol:
@Conan-Kudo remove the last commit, then we (I?) can add the commit that removes PAM, so we don't need to be building it 8 times for too long.

Conan-Kudo:
@aleixpol Done.

aleixpol (approved):

aleixpol:
This makes PAM mandatory as discussed https://github.com/sddm/sddm/pull/1685

PR #1673, opened 2023-02-20
Make greeter window fullscreen

### LINKED ISSUE #1672: sddm on kwin_wayland: Multi-monitor opens all greeter windows on one monitor
Can be reproduced on fedora rawhide over libvirt with multiple QXL/Virtio displays.

One display shows the greeter, while the other displays are black. When triggering kwin's windows overview you can see multiple greeter windows were opened, but all on the same monitor.
Attached video should be clear:

https://user-images.githubusercontent.com/8167495/220163420-e870f953-88c0-466c-9de3-911776a3d29b.mp4

aleasto:
Only fullscreen clients can decide the output display on wayland

Fixes #1672

### DIFF
--- a/src/greeter/GreeterApp.cpp
+++ b/src/greeter/GreeterApp.cpp
@@ -228,7 +228,7 @@ namespace SDDM {
 
         // show
         qDebug() << "Adding view for" << screen->name() << screen->geometry();
-        view->show();
+        view->showFullScreen();
 
         // activate windows for the primary screen to give focus to text fields
         if (QGuiApplication::primaryScreen() == screen)


Vogtinator (approved):

Conan-Kudo (approved):

aleixpol:
This fix is wrong though. We are not using xdg shell so the premise of "only fullscreen windows choose output" doesn't apply. It should either be using layer shell or xdg fullscreen, on either this change shouldn't make a difference (at best) or it broke for other use cases.

I'd suggest reverting this and taking some time to understand what you are seeing.

aleasto:
~This change fixes it for `QT_WAYLAND_SHELL_INTEGRATION=layer-shell`~

Vogtinator:
> This fix is wrong though. We are not using xdg shell so the premise of "only fullscreen windows choose output" doesn't apply. It should either be using layer shell or xdg fullscreen, on either this change shouldn't make a difference (at best) or it broke for other use cases.

IMO this fix is not wrong in any case, as we want the greeter to be fullscreen (except maybe for test mode? Debatable). According to the `QWindow::show` doc:

> This is equivalent to calling [showFullScreen](https://doc.qt.io/qt-5/qwindow.html#showFullScreen)(), [showMaximized](https://doc.qt.io/qt-5/qwindow.html#showMaximized)(), or [showNormal](https://doc.qt.io/qt-5/qwindow.html#showNormal)(), depending on the platform's default behavior for the window type and flags.

Currently this defaults to `showNormal()` on the wayland platform.

aleasto:
> This change fixes it for `QT_WAYLAND_SHELL_INTEGRATION=layer-shell`

Well, I went to double check and it doesn't. I must have been trying with `QT_WAYLAND_SHELL_INTEGRATION=xdg-shell` which I can confirm _double-checked_ is fixed by this

Vogtinator:
To clarify, it's still broken with `QT_WAYLAND_SHELL_INTEGRATION=layer-shell` but works with `xdg-shell` now?

aleasto:
Yea. Previously both showed the behaviour of spawning both windows in the same screen, as per my screenrecording in #1672

Vogtinator:
Can be reproduced easily by using `dbus-run-session kwin_wayland --wayland-display $WAYLAND_DISPLAY` and then `WAYLAND_DISPLAY=wayland-1 QT_WAYLAND_SHELL_INTEGRATION=layer-shell ./src/greeter/sddm-greeter --test-mode`.

With `WAYLAND_DEBUG=1`, it shows that it does not specify an output on layer creation:

```zwlr_layer_shell_v1@24.get_layer_surface(new id zwlr_layer_surface_v1@38, wl_surface@43, nil, 2, "window")```

**Comment by Vogtinator** [comment]
Works with this patch to layer-shell-qt:

```diff
diff --git a/src/qwaylandlayersurface.cpp b/src/qwaylandlayersurface.cpp
index dd17aa0..e172b52 100644
--- a/src/qwaylandlayersurface.cpp
+++ b/src/qwaylandlayersurface.cpp
@@ -25,6 +25,10 @@ QWaylandLayerSurface::QWaylandLayerSurface(QWaylandLayerShell *shell, QtWaylandC
 
     wl_output *output = nullptr;
     QScreen *screen = interface->desiredOutput();
+    if (!screen) {
+        screen = window->window()->screen();
+    }
+
     if (screen) {
         auto waylandScreen = dynamic_cast<QtWaylandClient::QWaylandScreen *>(screen->handle());
         // Qt will always assign a screen to a window, but if the compositor has no screens available a dummy QScreen object is created
```

That would be a rather big oversight though, so I wonder whether that was intentionally not done.

aleasto:
Yes that works on my test set up.

Vogtinator:
I opened https://invent.kde.org/plasma/layer-shell-qt/-/merge_requests/26 for discussion

PR #1667, opened 2023-02-18
Port to Qt6

davispuh:
Port to Qt 6. Main things:
* `RegExp` has been removed, instead use `QRegularExpression`
* No more `QProcess::setupChildProcess` but instead `setChildProcessModifier`
* `QSettings` now uses UTF-8
* `QStringRef` has been removed. There is [QStringView](https://doc.qt.io/qt-6/qstringview.html) but that looks useful only for UTF-16 and [QLatin1StringView](https://doc.qt.io/qt-6/qlatin1stringview.html) but we need local encoding/UTF-8 so not usable

Something changed regarding `QMetaType` so need to include `Auth*` stuff aswell otherwise got linking failures
```
[ 70%] Linking CXX executable sddm-helper
/usr/bin/ld: CMakeFiles/sddm-helper.dir/sddm-helper_autogen/mocs_compilation.cpp.o: in function `QtPrivate::MetaObjectForType<SDDM::Auth::Info, void>::metaObjectFunction(QtPrivate::QMetaTypeInterface const*)':
/usr/src/debug/sddm-git/build/src/helper/sddm-helper_autogen/mocs_compilation.cpp:924: undefined reference to `SDDM::Auth::staticMetaObject'
/usr/bin/ld: CMakeFiles/sddm-helper.dir/sddm-helper_autogen/mocs_compilation.cpp.o: in function `QtPrivate::MetaObjectForType<SDDM::Auth::Error, void>::metaObjectFunction(QtPrivate::QMetaTypeInterface const*)':
/usr/src/debug/sddm-git/build/src/helper/sddm-helper_autogen/mocs_compilation.cpp:924: undefined reference to `SDDM::Auth::staticMetaObject'
/usr/bin/ld: CMakeFiles/sddm-helper.dir/sddm-helper_autogen/mocs_compilation.cpp.o: in function `std::enable_if<QtPrivate::IsQEnumHelper<SDDM::Auth::Info>::Value, QDebug>::type operator<< <SDDM::Auth::Info>(QDebug, SDDM::Auth::Info)':
/usr/include/qt6/QtCore/qdebug.h:398: undefined reference to `SDDM::Auth::staticMetaObject'
[...]

PR #1674, opened 2023-02-21
XAuth::writeCookieToFile: Add error messages

davispuh:
While testing #1667 I saw weird failure and logs weren't that useful:
```
Initializing...
Starting...
Logind interface found
Adding new display...
Loaded empty theme configuration
Xauthority path: "/var/run/sddm/xauth_gZSICE"
Using VT 2
Display server starting...
Writing cookie to "/var/run/sddm/xauth_gZSICE"
Failed to write xauth file
Attempt 1 starting the Display server on vt 2 failed
Tried to start Xorg before previous instance exited
Attempt 2 starting the Display server on vt 2 failed
Tried to start Xorg before previous instance exited
Attempt 3 starting the Display server on vt 2 failed
Could not start Display server on vt 2
```

There are several cases when it can fail so this PR adds logging to `XAuth::writeCookieToFile` to make debugging easier.
Also actually SDDM has really poor error handling/error logging because most of time it fails without saying WHY. And `sddm` process still keeps running doing nothing even after failure and user gets just black screen.

Anyway in this case failure was because Q6 changed behavior of `QByteArray` that `qbytearray[i]` doesn't affect `qbytearray.size()` but it still stayed at `0`. This might be Qt6 bug not sure...

### DIFF
--- a/src/common/XAuth.cpp
+++ b/src/common/XAuth.cpp
@@ -116,17 +116,22 @@ bool XAuth::writeCookieToFile(const QString &display, const QString &fileName,
 
     qDebug() << "Writing cookie to" << fileName;
 
-    if(display.size() < 2 || display[0] != QLatin1Char(':') || cookie.size() != 16)
+    if(display.size() < 2 || display[0] != QLatin1Char(':') || cookie.size() != 16) {
+        qWarning().nospace() << "Unexpected DISPLAY='" << display << "' or cookie.size() = " << cookie.size();
         return false;
+    }
 
     // The file needs 0600 permissions
     const int oldumask = umask(077);
 
     // Truncate the file. We don't support merging like the xauth tool does.
     FILE * const authFp = fopen(qPrintable(fileName), "wb");
+    auto error = errno;
     umask(oldumask);
-    if (authFp == nullptr)
+    if (authFp == nullptr) {
+        qWarning().nospace() << "fopen() failed with errno=" << error;
         return false;
+    }
 
     auto fileCloser = qScopeGuard([authFp]{ fclose(authFp); });
     char localhost[HOST_NAME_MAX + 1] = "";
@@ -149,18 +154,27 @@ bool XAuth::writeCookieToFile(const QString &display, const QString &fileName,
     auth.data = cookie.data();
     auth.data_length = cookie.size();
 
+    errno = 0;
     if (XauWriteAuth(authFp, &auth) == 0) {
+        qWarning().nospace() << "XauWriteAuth(FamilyLocal) failed with errno=" << errno;
         return false;
     }
 
     // Write the same entry again, just with FamilyWild
     auth.family = FamilyWild;
     auth.address_length = 0;
+    errno = 0;
     if (XauWriteAuth(authFp, &auth) == 0) {
+        qWarning().nospace() << "XauWriteAuth(FamilyWild) failed with errno=" << errno;
         return false;
     }
 
-    return fflush(authFp) != EOF;
+    if (fflush(authFp) != 0) {
+        qWarning().nospace() << "fflush() failed with errno=" << errno;
+        return false;
+    }
+
+    return true;
 }
 
 } // namespace SDDM


src/common/XAuth.cpp
```diff
@@ -149,18 +154,28 @@ bool XAuth::writeCookieToFile(const QString &display, const QString &fileName,
...
     // Write the same entry again, just with FamilyWild
     auth.family = FamilyWild;
     auth.address_length = 0;
+    errno = 0;
     if (XauWriteAuth(authFp, &auth) == 0) {
+        qWarning().nospace() << "XAuth::writeCookieToFile: XauWriteAuth(FamilyWild) failed with errno=" << errno;
+        return false;
+    }
+
+    auto result = fflush(authFp);
```
Vogtinator: Can be just `if (fflush(authFp) != 0)`

Vogtinator (approved):
IMO a bit verbose (qWarning knows the function, file and line already), but I don't really care that much.

davispuh:
> IMO a bit verbose (qWarning knows the function, file and line already), but I don't really care that much.

Then how can we enable to show that for logs? Because currently if you look at my provided log example then it doesn't output function name. Neither running `sddm` directly nor thru Systemd it does't give extra log info. I would say for errors/warning that is quite useful to quickly find where exactly it failed because same/similar error message can be used in different/multiple locations.

src/common/XAuth.cpp
davispuh: Yeah true. At first I was thinking maybe should log `fflush`  return value aswell and that's why I put it in separate variable but then looking at man-page it didn't look like it would contain anything useful so doesn't seem worth bothering with that.

Vogtinator:
It's currently not visible in logs because SDDM installs its own (rather bad) message handlers which don't respect the message format. See also the discussion in https://github.com/sddm/sddm/pull/1653.

davispuh:
I see, indeed that needs fixing. Fixing logging would make debugging a lot simpler/easier.

I see that for `standardLogger` we completely drop `context` and don't use it at all. And for `journaldLogger` I was hopeful because we do pass `context` and use  `sd_journal_print_with_location` so we should have proper function/file/line info but for some reason something doesn't work with that.

This is what I see
```
$ journalctl -o json -u sddm | jq -r '(.__REALTIME_TIMESTAMP | tonumber / 1000000 | todate) + " " + .SYSLOG_IDENTIFIER + "[" + ._PID + "] " + .CODE_FUNC + "() " + .CODE_FILE + ":" + .CODE_LINE + " - " + .MESSAGE'

2023-02-21T20:58:24Z systemd[1] job_emit_done_message() src/core/job.c:768 - Started Simple Desktop Display Manager.
2023-02-21T20:58:25Z sddm[827] unknown() unknown:0 - Initializing...
2023-02-21T20:58:26Z sddm[827] unknown() unknown:0 - Starting...
[...]
2023-02-21T20:58:45Z sddm-helper[920] unknown() unknown:0 - [PAM] Conversation with 1 messages
2023-02-21T20:58:45Z sddm-helper[920] () : - gkr-pam: unable to locate daemon control file
2023-02-21T20:58:45Z sddm-helper[920] () : - gkr-pam: stashed password to try later in open session
2023-02-21T20:58:45Z sddm-helper[920] unknown() unknown:0 - [PAM] returning.
2023-02-21T20:58:45Z sddm[827] unknown() unknown:0 - Authenticated successfully
2023-02-21T20:58:45Z sddm-helper[920] () : - pam_unix(sddm:session): session opened for user Dāvis(uid=1000) by (uid=0)
2023-02-21T20:58:45Z sddm[827] unknown() unknown:0 - Auth: sddm-helper exited successfully
2023-02-21T20:58:45Z sddm[827] unknown() unknown:0 - Greeter stopped. SDDM::Auth::HELPER_SUCCESS
2023-02-21T20:58:46Z sddm[827] unknown() unknown:0 - Session started true
2023-02-21T20:58:51Z sddm[827] unknown() unknown:0 - Auth: sddm-helper exited successfully
[...]
```

I can see files/lines/functions for systemd itself but for `sddm` it's all `unknown`. Maybe there is some CMake flag to build `sddm` with but note that systemd already came with that info by default so think it does make sense to build with it by default.

herostrat (approved):

davispuh:
I removed function names from messages as they're not needed. Will be there once someone fixes context :D so I think this can be merged now

Vogtinator:
Not sure whether the `errno = 0;` is actually necessary but they won't hurt. Thanks!

Vogtinator:
BTW, the `QByteArray` behaviour change is intentional: https://doc.qt.io/qt-5/qbytearray.html#operator-5b-5d vs. https://doc.qt.io/qt-6/qbytearray.html#operator-5b-5d.

The old behaviour is technically not allowed by Qt 5.15 either (which was IMO a bad idea), but apparently it still works?
Can you make a PR with the `XAuth::setup` change from #1667?

davispuh:
> Not sure whether the `errno = 0;` is actually necessary but they won't hurt. Thanks!

Based on documentation it looked needed because if you check `XauWriteAuth` man page it says:

> XauWriteAuth writes an authorization entry to auth_file.  It returns 1 on success, 0 on failure.

No mention if `errno` is used at all. And if you look at `XauLockAuth`

> LOCK_ERROR
A system error occurred, either a file_name which is too long, or an unexpected failure  from  a  system  call.
errno **may** prove useful.

So basically my understanding from this is that `errno` might be used or it might not in which case it's better to set it as `0` so if it fails there it will just show it as `0` without giving false previous information (eg. from previous `gethostname` call and you could waste tons of time trying to find why such errno was returned..)

davispuh:
Sure, created #1680

PR #1678, opened 2023-03-05
fix: set RUNTIME_DIR to /run/sddm to follow FHS 3.0

marcdeop:
After an upgrade on our Fedora systems, one of our users/testers noticed this on their logs:
```
Line references path below legacy directory /var/run/, updating /var/run/sddm/run/sddm; please update the tmpfiles.d/ drop-in file accordingly.
```

According to [FHS 3.0](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#varrunRuntimeVariableData) and [systemd](https://github.com/systemd/systemd/pull/9019) we should be using directly `/run` rather than `/var/run`.

More [info](https://access.redhat.com/solutions/4154291)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,12 +185,18 @@ if (NOT ELOGIND_FOUND AND NOT SYSTEMD_FOUND)
     set(REBOOT_COMMAND "/sbin/shutdown -r now")
 endif()
 
+if (SYSTEMD_FOUND)
+    set(RUNTIME_DIR_DEFAULT "/run/sddm")
+else()
+    set(RUNTIME_DIR_DEFAULT "${CMAKE_INSTALL_FULL_LOCALSTATEDIR}/run/sddm")
+endif()
+
 
 # Set constants
 set(DATA_INSTALL_DIR            "${CMAKE_INSTALL_FULL_DATADIR}/sddm"                CACHE PATH      "System application data install directory")
 set(DBUS_CONFIG_DIR             "${CMAKE_INSTALL_SYSCONFDIR}/dbus-1/system.d"       CACHE PATH      "DBus config files directory")
 set(STATE_DIR                   "${CMAKE_INSTALL_FULL_LOCALSTATEDIR}/lib/sddm"      CACHE PATH      "State directory")
-set(RUNTIME_DIR                 "${CMAKE_INSTALL_FULL_LOCALSTATEDIR}/run/sddm"      CACHE PATH      "Runtime data storage directory")
+set(RUNTIME_DIR                 "${RUNTIME_DIR_DEFAULT}"                            CACHE PATH      "Runtime data storage directory")
 set(QML_INSTALL_DIR             "${QT_IMPORTS_DIR}"                                 CACHE PATH      "QML component installation directory")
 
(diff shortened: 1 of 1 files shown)


Vogtinator:
There is `CMAKE_INSTALL_FULL_RUNSTATEDIR` set by GNUInstallDirs. Is that set correctly by default?

marcdeop:
> There is `CMAKE_INSTALL_FULL_RUNSTATEDIR` set by GNUInstallDirs. Is that set correctly by default?

I honestly don't know. Not really an expert on this topics unfortunately :(

Vogtinator:
I tried it locally, looks like it's set to `/var/run` still.

Vogtinator:
Is `/run` a thing on FreeBSD as well? If not, might need an OS dependent default

marcdeop:
> Is `/run` a thing on FreeBSD as well? If not, might need an OS dependent default

Seems like you are right. According to [this](https://man.freebsd.org/cgi/man.cgi?query=hier&sektion=7) , Freebsd still uses `/var/run`

Perhaps I can have a look later on how to do this conditional

marcdeop:
@Vogtinator would this be better?

Conan-Kudo (approved):

Vogtinator:
I'm not too fond of the line duplication, maybe use an intermediate `RUNTIME_DIR_DEFAULT` variable?

marcdeop:
ok, third time is the charm. Right? :smile:

Vogtinator (approved):

marcdeop:
I think I force-pushd after @Vogtinator approved my changes (I adjust a bit the commit message but didn't change the code)

CMakeLists.txt
```diff
@@ -185,12 +185,18 @@ if (NOT ELOGIND_FOUND AND NOT SYSTEMD_FOUND)
...
+    set(RUNTIME_DIR_DEFAULT "${CMAKE_INSTALL_FULL_LOCALSTATEDIR}/run")
+endif()
+
 
 # Set constants
 set(DATA_INSTALL_DIR            "${CMAKE_INSTALL_FULL_DATADIR}/sddm"                CACHE PATH      "System application data install directory")
 set(DBUS_CONFIG_DIR             "${CMAKE_INSTALL_SYSCONFDIR}/dbus-1/system.d"       CACHE PATH      "DBus config files directory")
 set(STATE_DIR                   "${CMAKE_INSTALL_FULL_LOCALSTATEDIR}/lib/sddm"      CACHE PATH      "State directory")
-set(RUNTIME_DIR                 "${CMAKE_INSTALL_FULL_LOCALSTATEDIR}/run/sddm"      CACHE PATH      "Runtime data storage directory")
+set(RUNTIME_DIR                 "${RUNTIME_DIR_DEFAULT}/sddm"                       CACHE PATH      "Runtime data storage directory")
```
Vogtinator: Could you move the `/sddm` part into `RUNTIME_DIR_DEFAULT` as well so that `RUNTIME_DIR_DEFAULT` is actually the default of `RUNTIME_DIR`?

marcdeop: should be fixed now

Vogtinator (approved):

Conan-Kudo (approved):

PR #1679, opened 2023-03-06
Make accountsservice data directory overridable via CMake.

Zoid-dr:
On FreeBSD accountsservice stores its data in /var/db/accountsservice. This patch provides a way to change the default /var/lib/accountsservice path to something else while maintaining a sensible default.


arrowd:
Looks good from our (FreeBSD) side.

Vogtinator (approved):
LGTM. Technically it could use `CMAKE_INSTALL_FULL_LOCALSTATEDIR`, but I think for external resources, hardcoding this might actually be better.

PR #1680, opened 2023-03-06
XAuth: Fix `m_cookie` creation for `QByteArray`

davispuh:
Qt6 have changed behaviour that `qbytearray[i]` doesn't affect `size()` calculation.

It's still working fine with Qt `5.15.8` but we can fix it already.


Vogtinator (approved):
LGTM, thanks!

PR #1681, opened 2023-03-08
Enable HiDPI scaling by default

Pointedstick:
This setting works properly out of the box now.

Would obviate https://bugs.kde.org/show_bug.cgi?id=467039 in KDE.

### DIFF
--- a/data/man/sddm.conf.rst.in
+++ b/data/man/sddm.conf.rst.in
@@ -159,7 +159,7 @@ OPTIONS
 `EnableHiDPI=`
 	Enables Qt's automatic HiDPI scaling.
 	Can be either "true" or "false".
-	Default value is "false".
+	Default value is "true".
 
 The `XauthPath=` option is no longer necessary, libxau is used instead.
 
@@ -189,7 +189,7 @@ The `UserAuthFile=` option was removed, the file is always created as
 `EnableHiDPI=`
 	Enables Qt's automatic HiDPI scaling.
 	Can be either "true" or "false".
-	Default value is "false".
+	Default value is "true".
 
 [Users] section:
 
--- a/src/common/Configuration.h
+++ b/src/common/Configuration.h
@@ -74,7 +74,7 @@ namespace SDDM {
             Entry(SessionLogFile,      QString,     _S(".local/share/sddm/xorg-session.log"),   _S("Path to the user session log file"));
             Entry(DisplayCommand,      QString,     _S(DATA_INSTALL_DIR "/scripts/Xsetup"),     _S("Path to a script to execute when starting the display server"));
             Entry(DisplayStopCommand,  QString,     _S(DATA_INSTALL_DIR "/scripts/Xstop"),      _S("Path to a script to execute when stopping the display server"));
-            Entry(EnableHiDPI,         bool,        false,                                      _S("Enable Qt's automatic high-DPI scaling"));
+            Entry(EnableHiDPI,         bool,        true,                                      _S("Enable Qt's automatic high-DPI scaling"));
         );
 
         Section(Wayland,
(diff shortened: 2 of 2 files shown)


Conan-Kudo (approved):

Pointedstick:
One moment; I'm seeing incorrect rotation in SDDM with this set and I need to do some troubleshooting...

Vogtinator:
FYI, in openSUSE we have

```
[X11]
EnableHiDPI=yes
```

for a few years now so we could also include that as default?

Pointedstick:
My issue might be KWin's fault; see https://bugs.kde.org/show_bug.cgi?id=467070.

@Vogtinator how does this setting interact with the `ServerArguments: -dpi [something]` option on X11? Currently our SDDM KCM sets that if you're using any scaling on X11. So if the two settings conflict in any way, we'll need to stop setting it in KDE code, and possible handle in SDDM code the situation where both are set, to account for people who suddenly have both set once they update SDDM. Might be less risk for now to only for it for Wayland.

Pointedstick:
https://bugs.kde.org/show_bug.cgi?id=467070 turned out to be a local configuration error and actually verifies that SDDM is doing the right thing, lol. Un-marking as a draft.

Vogtinator:
> @Vogtinator how does this setting interact with the `ServerArguments: -dpi [something]` option on X11? Currently our SDDM KCM sets that if you're using any scaling on X11. So if the two settings conflict in any way, we'll need to stop setting it in KDE code, and possible handle in SDDM code the situation where both are set, to account for people who suddenly have both set once they update SDDM. Might be less risk for now to only for it for Wayland.

Using `-dpi` without `EnableHiDPI=true` is somewhat broken: it only scales text but not other UI elements. With the Plasma theme that's not immediately noticable because many sizes are derived based on font metrics meanwhile. (Partially because of this issue with SDDM I imagine).

Here's how it looks with the built-in maui theme.

96 DPI:
Screenshot_20230308_212625

192 DPI, `EnableHiDPI=false`:
Screenshot_20230308_212835

Font size doubled, UI elements have same size as with 96 DPI.

192 DPI, `EnableHiDPI=true`:
Screenshot_20230308_212814

Now the UI elements doubled in size as well.

Vogtinator:
It might need some testing how exactly this behaves with a mixed-DPI multimonitor setup. In that case the `-dpi` parameter is broken by design though.

Pointedstick:
Thanks for that. Makes sense. Mixed-DPI multimonitor is already expected to be broken on X11 so I think we can safely ignore it.

Vogtinator:
> Thanks for that. Makes sense. Mixed-DPI multimonitor is already expected to be broken on X11 so I think we can safely ignore it.

With SDDM it did actually work fine, provided that the monitor EDIDs are roughly correct. This was a good reason for us to enable it early, as there was no option to use wayland yet.

Pointedstick:
Interesting, let me test a bit.

Pointedstick:
So what I see is that with X11, `HiDPI=true` alone doesn't seem to do anything. If I add `ServerArguments=-dpi 192`, then everything looks correct at 200% scale on my high DPI monitor, but the low DPI monitor also gets scaled inappropriately and is twice as big as it should be.

But this isn't a regression from the status quo for people who sync their settings; with only `ServerArguments=-dpi 192` set, I see roughly the same thing--with the exception that UI elements that don't scale with the font size being too small, which is fixed with this MR being applied.

TL;DR: I think this should be a safe change. :)

Vogtinator:
> So what I see is that with X11, HiDPI=true alone doesn't seem to do anything.

Yeah, they unfortunately broke that in Qt 5.14 (or 5.15) when they rewrote HiDPI stuff for the third time :-/

Vogtinator:
Since then, the physical DPI of the monitor is ignored by default, unless `QT_USE_PHYSICAL_DPI` is set (see also https://github.com/sddm/sddm/pull/1514). Unfortunately there are no sanity checks around that anymore, so setting that by default will cause SDDM to become unusable on some systems.

aleixpol:
IMO, changing this for Wayland is a yes, no-brainer. For X11, high DPI on X11 is broken on many levels, especially with multiple monitors. Leaving it as is can be good for our collective peace of mind.

Pointedstick:
I tested it on X11 with a high DPI monitor, a low DPI, and a mixed-DPI setup, both with and without the `ServerArgument` setting set appropriately. I didn't find any regressions. And @Vogtinator says they've been shipping this in openSUSE for years.

But I won't argue the point too much, and if you want me to remove the X11 part so this can merge, I will. Let me know.

Conan-Kudo:
I don't think we should be trying to do more with the X11 side of things anymore.

aleixpol (approved):

Vogtinator:
> And @Vogtinator says they've been shipping this in openSUSE for years.

I checked: It was introduced with the update to 0.18.0 in July 2018!

PR #1684, opened 2023-03-13
Prefer using a virtual to ifs and dynamic casts

aleixpol:


Vogtinator:
Missing `override` decls?

aleixpol:
> Missing `override` decls?

I didn't add it because then I had to add it to the other methods in the class which also are overrides. I'd suggest adding a mandatory override warning on a separate MR and add them all at once.

Vogtinator:
Ok

Vogtinator:
IIRC one of the "fix warnings" PRs adds override to most (or all) declarations.

Vogtinator (approved):

herostrat:
You're right, this would be #1664

Vogtinator:
#1664 is merged, can you add `override` here now?

src/daemon/XorgDisplayServer.h
```diff
@@ -41,6 +41,8 @@ namespace SDDM {
 
         const QByteArray cookie() const;
 
+        void setDisplayName(const QString &displayName);
```
Vogtinator, line 44: Missing `override`

Vogtinator (changes requested):

PR #1685, opened 2023-03-13
Remove the Passwd backend, make PAM mandatory

aleixpol:
As far as I know it's unused and thus untested.


Conan-Kudo (approved):

CMakeLists.txt
```diff
@@ -74,14 +73,7 @@ include(FeatureSummary)
...
-    find_package(PAM REQUIRED)
-
-    if(PAM_FOUND)
-        add_definitions(-DUSE_PAM)
-    endif()
-endif()
-add_feature_info("PAM" PAM_FOUND "PAM support")
+find_package(PAM REQUIRED)
 
 # getspnam and shadow(5) support
```
Vogtinator: This block is also no longer needed

herostrat (approved):

a17r:
I guess we'll find out how many people were still using this after 0.20.0 release.

CMakeLists.txt
aleixpol: Addressed

Vogtinator (approved):

PR #1699, opened 2023-04-04
Do not force QT_WAYLAND_DISABLE_WINDOWDECORATION=1 on wayland greeters

### LINKED ISSUE #1441: Can't run with a full colour depth on Wayland
Tested both on Weston and KWin, I don't really get all the colours on my display. It's possibly that my users didn't have some permission enabled but it's something that affects SDDM performance on wayland mode.

A picture, which says more than a thousand words:
IMG_20210827_122953

I really don't know what is wrong and neither kwin or weston's logs are very helpful there.

aleixpol:
This would put Qt in a weird state (see the issue that we are fixing) where it would use the wrong GL configuration.

Arguably it's an error of Qt for having a bogus default pixel format but this change should fix it but also make sure it behaves like other Qt applications and thus more predictably.

Fixes #1441

### DIFF
--- a/src/auth/Auth.cpp
+++ b/src/auth/Auth.cpp
@@ -153,64 +153,66 @@ namespace SDDM {
         Auth *auth = qobject_cast<Auth*>(parent());
         Msg m = MSG_UNKNOWN;
         SafeDataStream str(socket);
-        str.receive();
-        str >> m;
-        switch (m) {
-            case ERROR: {
-                QString message;
-                Error type = ERROR_NONE;
-                str >> message >> type;
-                Q_EMIT auth->error(message, type);
-                break;
-            }
-            case INFO: {
-                QString message;
-                Info type = INFO_NONE;
-                str >> message >> type;
-                Q_EMIT auth->info(message, type);
-                break;
-            }
-            case REQUEST: {
-                Request r;
-                str >> r;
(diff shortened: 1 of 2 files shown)


Vogtinator:
I would assume that at the point it was added, setting the variable was necessary. Did that change or is it still the case with `fullscreen-shell-v1`?

(IMO `QT_WAYLAND_SHELL_INTEGRATION` should also no longer be set - with #1673, the default xdg-shell should also work)

aleixpol:
> I would assume that at the point it was added, setting the variable was necessary. Did that change or is it still the case with `fullscreen-shell-v1`?

I would assume as well. It makes no sense that Qt is rendering decorations on fullscreen-shell-v1 either, we can consider it a limitation of its impementation in Qt.

In practice, I don't think anyone should be using the default behaviour of weston + fullscreen-shell-v1 anyway.

Vogtinator (approved):
I just tried this with weston, works fine without it :shrug:

Vogtinator:
For some reason [48a98c7](https://github.com/sddm/sddm/pull/1699/commits/48a98c7ba45b49cf443ac46f0d218f044933c603) shows up as part of this PR now even though that's in the target branch. I'll hit the button anyway.

PR #1716, opened 2023-05-08
Optionally support builds against Qt6

cordlandwehr:
This is a revisited version if the patch series https://github.com/sddm/sddm/pull/1667 but with build flags to simultaneously support Qt5 and Qt6 builds.
Build tested against Qt5.15 and Qt6.5

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,9 +24,19 @@ jobs:
           - "quay.io/centos/centos:stream9"
           - "registry.opensuse.org/opensuse/tumbleweed-dnf"
           - "docker.io/library/ubuntu:22.04"
+          - "docker.io/library/ubuntu:23.04"
         compiler:
           - gcc
           - clang
+        qt:
+          - "qt5"
+        include:
+          - container-image: "docker.io/library/ubuntu:23.04"
+            compiler: clang
+            qt: "qt6"
+          - container-image: "docker.io/library/ubuntu:23.04"
+            compiler: gcc
+            qt: "qt6"
     runs-on: ubuntu-latest
 
     container:
@@ -49,13 +59,22 @@ jobs:
           sed -i '/^#\sdeb-src /s/^#//' "/etc/apt/sources.list"
           apt-get update -y
           DEBIAN_FRONTEND=noninteractive apt-get build-dep sddm -y
-          DEBIAN_FRONTEND=noninteractive apt-get install clang clazy qml-module-qttest -y
+          if [ "${{ matrix.qt }}" = "qt6" ]; then
+            DEBIAN_FRONTEND=noninteractive apt-get install clang clazy qt6-base-dev qt6-base-dev-tools qt6-tools-dev qt6-declarative-dev qml6-module-qtqml-workerscript qml6-module-qtquick-window -y
+          else
+            DEBIAN_FRONTEND=noninteractive apt-get install clang clazy qml-module-qttest -y
+          fi
       - name: Build
         run: |
           set -x
           mkdir -p build
           cd build
-          if [ "${{ matrix.compiler }}" == "clang" ]; then
+          if [ "${{ matrix.qt }}" = "qt6" ]; then
+            BUILD_WITH_QT6=ON
+          else
+            BUILD_WITH_QT6=OFF
+          fi
+          if [ "${{ matrix.compiler }}" = "clang" ]; then
             export CC=clang
             #export CXX=clang++
             export CXX=clazy
@@ -67,6 +86,7 @@ jobs:
           fi
           cmake .. \
             -DBUILD_MAN_PAGES:BOOL=ON \
+            -DBUILD_WITH_QT6:BOOL=${BUILD_WITH_QT6} \
             -DLOGIN_DEFS_PATH:PATH="${LOGIN_DEFS_PATH}"
           make -j $(getconf _NPROCESSORS_ONLN)
           make install
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,7 @@ option(BUILD_MAN_PAGES "Build man pages" OFF)
 option(ENABLE_JOURNALD "Enable logging to journald" ON)
 option(NO_SYSTEMD "Disable systemd support" OFF)
 option(USE_ELOGIND "Use elogind instead of logind" OFF)
+option(BUILD_WITH_QT6 "Build with Qt 6" OFF)
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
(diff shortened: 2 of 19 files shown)


CMakeLists.txt
```diff
@@ -84,11 +89,19 @@ find_package(XCB REQUIRED)
 # XKB
 find_package(XKB REQUIRED)
 
-# Qt 5
-find_package(Qt5 5.15.0 CONFIG REQUIRED Core DBus Gui Qml Quick LinguistTools Test QuickTest)
+# Qt
+if(BUILD_WITH_QT6)
+    set(QT_MAJOR_VERSION 6)
```
Vogtinator, line 90: Hm, we have `BUILD_WITH_QT6`, `QT6` and `QT_MAJOR_VERSION` now. Maybe just have `QT_MAJOR_VERSION` as the configurable option directly?

cordlandwehr:
Please specifically have a look at the port of QStringRef. I do not see why QStringView should not be applicable here as mentioned in the original patch set.

src/auth/Auth.cpp
```diff
@@ -102,7 +103,7 @@ namespace SDDM {
         static std::unique_ptr<Auth::SocketServer> self;
         if (!self) {
             self.reset(new SocketServer());
-            self->listen(QStringLiteral("sddm-auth-%1").arg(QUuid::createUuid().toString().replace(QRegExp(QStringLiteral("[{}]")), QString())));
+            self->listen(QStringLiteral("sddm-auth-%1").arg(QUuid::createUuid().toString().replace(QRegularExpression(QStringLiteral("[{}]")), QString())));
```
Vogtinator: Not a straight port, but this can be replaced by `QUuid::createUuid().toString(QUuid::WithoutBraces)`

src/common/ConfigReader.cpp
```diff
@@ -27,19 +27,31 @@
...
+#endif
+#include <QtCore/QStringView>
 
 QTextStream &operator>>(QTextStream &str, QStringList &list)  {
     list.clear();
 
     QString line = str.readLine();
-
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+    const auto strings = QStringTokenizer{line, u','};
```
Vogtinator: Is it worth it to use `QStringTokenizer` here or would `QStringView::split` work well for both?

src/common/Session.cpp
```diff
@@ -215,7 +223,11 @@ namespace SDDM {
     {
         QProcessEnvironment env;
 
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+        const auto entryList = QStringTokenizer{list, u','};
```
Vogtinator: Same as above, with `QStringView` this could be the same for both Qt versions

src/auth/Auth.cpp
cordlandwehr: Yeah, missed it, good idea!

CMakeLists.txt
cordlandwehr: Yes, that many options were at least one too much ;) Yet I still would prefer to have a boolean option to a parameter value for better usability.

Vogtinator:
Looks good! Could you try to get some Qt 6 builds into the github actions as well? That's the only missing piece FWICT and would be necessary to ensure that this doesn't get broken by accident.

cordlandwehr:
Sure, will have a look at the github action.
Additional to this, I noticed another minor problem that I have to fix, too. The install logic for the QML components breaks with Qt6.

cordlandwehr:
Regarding the QML issue I recently noticed, it turned out to be a Yocto specific topic. So there is nothing to do.
Please have a look at the pipeline configuration if it fits to your expectations.

Vogtinator:
IMO it's fine to have just a single Qt6 build in addition to the multiple Qt 5 ones, but it would be nicer to not have the code duplicated there for just those two lines (qt6 dep installation, then `BUILD_WITH_QT6`). Can you change that to a `matrix: include` with an `${{ matrix.qt6 }}` statement or similar?

cordlandwehr:
Yeah, that looks nice code-wise! However, I must admit that I am new to GitHub matrix jobs and currently struggle to get it working. My current WIP is here https://github.com/cordlandwehr/sddm/pull/1/commits/04b37c41ba0cdf46bd9409b01ffdd3fec3c6a1ad
if you directly see what is wrong, I would be very interested! It seems that my additional matrix "include:" option is simply ignored and still just 10 jobs are created instead of 12...

cordlandwehr:
Updated PR with conditional matrix syntax: only/exactly the Ubuntu 23.04 image is building with Qt6

.github/workflows/build.yml
```diff
@@ -49,13 +53,22 @@ jobs:
...
           set -x
           mkdir -p build
           cd build
-          if [ "${{ matrix.compiler }}" == "clang" ]; then
+          if [ "${{ matrix.qt }}" = "qt6" ]; then
+            QT6_BUILD=ON
+          else
+            QT6_BUILD=OFF
+          fi
+          if [ "${{ matrix.compiler }}" = "clang" ]; then
```
Vogtinator: For consistency:

```
          if [ "${{ matrix.qt }}" = "qt6" ]; then
            BUILD_WITH_QT6=ON
          else
            BUILD_WITH_QT6=OFF
          fi
          if [ "${{ matrix.compiler }}" = "clang" ]; then
```

.github/workflows/build.yml
```diff
@@ -24,9 +24,13 @@ jobs:
           - "quay.io/centos/centos:stream9"
           - "registry.opensuse.org/opensuse/tumbleweed-dnf"
           - "docker.io/library/ubuntu:22.04"
+          - "docker.io/library/ubuntu:23.04"
         compiler:
           - gcc
           - clang
+        include:
```
Vogtinator, line 33: If you do

```
qt:
   - qt5
```

or similar, you can then use `matrix.qt` in text including the job titles. Currently it's not obvious from the outside that it's building a Qt 6 version

Vogtinator (approved):
LGTM, only nitpicks remaining

.github/workflows/build.yml
```diff
@@ -49,13 +59,22 @@ jobs:
...
+            DEBIAN_FRONTEND=noninteractive apt-get install clang clazy qml-module-qttest -y
+          fi
       - name: Build
         run: |
           set -x
           mkdir -p build
           cd build
-          if [ "${{ matrix.compiler }}" == "clang" ]; then
+          if [ "${{ matrix.qt }}" = "qt6" ]; then
+            QT6_WITH_BUILD=ON
```
Vogtinator: Needs to be `BUILD_WITH_QT6`

cordlandwehr: At least it was consistently wrong, but still embarrassing... thanks for spotting!

Issue #514, opened 2015-10-18
password field does not have focus on dual display setup


markg85:
Hi,

I know recent releases have had some fixes for dual display setups since the login used to be displayed on each monitor. Now only on the primary monitor.

A small side effect of this is that the mouse cursor initial position is in the center of the entire display surface (both screens combined). So for instance if i have two displays (of 1920x1200) next to each other the total resolution is 3840x1200. Now the initial mouse position would be at (or around) 1920x600 (center pixel). However, it seems the cursor is offset just ONE pixel to the right which means that the right display has the cursor (the left display has the login). This minor issue causes the password field to not have focus. It gains focus as soon as i move the mouse to the left part of the screen.

This is a minor annoyance, but annoying nevertheless. Would it be possible to fix this issue in the next sddm release?

markg85:
Note, what i described was with version 0.12.
I just installed the master branch and it shows the greeter on every display again (like it used to be some versions back). It's not a "fix", but it works for me.

plfiorini:
I don't understand what's broken with master, could you elaborate please?

BTW Whether showing the password prompt on all outputs or only on the primary one is up to the theme. SDDM screen model expose it.

markg85:
The greeter that i'm using (just the plasma 5 greeter) shows:
- Only on the primary display with version 0.12.x
- On all displays with master

I thought that might be a bug.

Regarding the password prompt (the reason i opened this bug report) is a theme issue according to you so this bug report can be closed for that.

plfiorini:
I'll check the themes later to see if there something wrong with the screen model.

mikkeloscar:
I experience the same problem with my own theme were I set `visible: primaryScreen` on the UI elements.

The best solution would be to force the cursor to spawn on the primary screen, but I'm not sure that is possible?

Another solution which I have seen in lightdm is to make the UI elements "follow" the mouse, so if the mouse is focusing on the first screen the UI elements appear on that, and if it is focusing on the second they appear on that instead.

The third solution is obviously having the UI elements appear on both screens but that it just confusing because it looks like the screens are mirrored even though that is not the case.

yaomtc:
I have a 4K monitor as my primary on the left, and a 1080p monitor on the right. When SDDM appears, the cursor ends up right-of-center on the 4K monitor - however the password field on the 1080p monitor has focus. So where the cursor ends up doesn't seem to have any affect on where the focus ends up.

avakyan-georgiy:
[...]

PR #1726, opened 2023-06-03
Bring the greeter on primary screen to the front

### LINKED ISSUE #612: Focus is not on password field
After bootup, the focus is not on the user or password field. This is fairly annoying, it means I would have to use the mouse before logging in. ("would" because I actually switched to a different DM to avoid this bug.)

I tried the themes: circles, elarun, maldives, maui.
Breeze is *not* affected, it seems to contain some special code to focus the password (see https://bugs.kde.org/show_bug.cgi?id=337079).

liangjs:
When there are multiple monitors connected, sddm-greeter creates new views for each screen, and activates the primary one to focus on its password field. However, if the monitors are initially setup to display the same content, different views will overlap with each other. The views created by non-primary screens may be placed on top of the primary view. In such cases, the primary view will be focused but not visible to the user.

This is fixed by calling `view->raise` for the primary screen. The primary view should be raised whenever a new screen is added, in order to avoid being covered by the newly created view.

Fixes: https://github.com/sddm/sddm/issues/612 https://github.com/sddm/sddm/issues/514

### DIFF
--- a/src/greeter/GreeterApp.cpp
+++ b/src/greeter/GreeterApp.cpp
@@ -231,8 +231,7 @@ namespace SDDM {
         view->showFullScreen();
 
         // activate windows for the primary screen to give focus to text fields
-        if (QGuiApplication::primaryScreen() == screen)
-            view->requestActivate();
+        activatePrimary();
     }
 
     void GreeterApp::removeViewForScreen(QQuickView *view) {
@@ -291,6 +290,7 @@ namespace SDDM {
         // activate and give focus to the window assigned to the primary screen
         for (QQuickView *view : qAsConst(m_views)) {
             if (view->screen() == QGuiApplication::primaryScreen()) {
+                view->raise();
                 view->requestActivate();
                 break;
             }


arrowd:
How does one configure X server to reproduce this problem?

liangjs:
> How does one configure X server to reproduce this problem?

It can be reproduced by connecting two monitors (for example, one builtin and one external) and then start sddm.

davidedmundson (approved):

arrowd:
I wonder if it is SDDM's own responsibility to start X server either in duplicated or Xinerama mode? Or maybe even a theme may request that from the daemon?

Some themes might be prepared for the Xinerama scenario, for example by drawring only the background on the second screen.

liangjs:
> I wonder if it is SDDM's own responsibility to start X server either in duplicated or Xinerama mode? Or maybe even a theme may request that from the daemon?
> 
> Some themes might be prepared for the Xinerama scenario, for example by drawring only the background on the second screen.

The SDDM greeter creates views for each screen, and then each view is managed by the theme separately.

https://github.com/sddm/sddm/blob/50409a3b11a68794323fefb319aa8078d4afc97a/src/greeter/GreeterApp.cpp#L277-L283

I don't know much about Xinerama. If it configures X server in a way that only one screen will be detected by SDDM, a single view will be created.

Vogtinator:
> The primary view should be raised whenever a new screen is added, in order to avoid being covered by the newly created view.

Not sure I agree with this - effectively this means whenever a screen is plugged in, the primary screen (essentially random) gets focus. Ideally focus shouldn't change, unless the new screen is the primary one, which is what the code does.

> However, if the monitors are initially setup to display the same content, different views will overlap with each other. The views created by non-primary screens may be placed on top of the primary view.

This sounds like the root cause. What about detecting whether a view overlaps anonther one (completely) and then ignoring it?
Can we assume that after a view appears, its position will not change?

lianghwe:
Why do we create multiple views when two monitors showing the same content? I think that is the problem.

Vogtinator:
> Why do we create multiple views when two monitors showing the same content? I think that is the problem.

In the case that all monitors have the same geometry, sharing would be possible, but in any other case it's not. To make it consistent, it's never shared.

Issue #1669, opened 2023-02-19
org.freedesktop.locale1 integration destroys existing keyboard layout configuration


alebastr:
The logic added in https://github.com/sddm/sddm/pull/1581 overwrites keymap settings in systemd-localed with a broken configuration.

Reproduction environment:
Fedora-KDE-Live-x86_64-38-20230219.n.0.iso
sddm-0.19.0^git20230201.3ee57e9-2.fc38.x86_64
sddm-wayland-plasma-5.27.0-3.fc38.noarch (SDDM greeter running under kwin_wayland).

Initial state (as configured in Anaconda during installation):
```
[alebastr@fedora ~]$ localectl 
System Locale: LANG=ru_RU.UTF-8
    VC Keymap: us
   X11 Layout: us,gb,ru
  X11 Variant: ,,
  X11 Options: grp:alt_shift_toggle
```
The system was configured with 3 keyboard layouts. Alt+Shift switches between layouts both in SDDM and in KDE Plasma session. There's no indication of a current layout in SDDM.

Now, we return to the SDDM login screen and select 'Russian' from the combobox.
```
[alebastr@fedora ~]$ localectl 
System Locale: LANG=ru_RU.UTF-8
    VC Keymap: us
   X11 Layout: ru
```

System x11 layout is now set to `ru`. Layout switching is no longer possible both in SDDM and in KDE Plasma session (i.e. if you selected a non-latin layout, you won't be able to input latin characters at all).
The state is preserved after reboot.

The only way to recover from that is to login on another VT and run `localectl set-x11-keymap` with the original settings.

Vogtinator:
IMO SDDM shouldn't use localed to set the keyboard layout. That's a system wide and persistent change.

alebastr:
I believe it would be appropriate to use `org.kde.KeyboardLayouts` here and hide the language selector if the interface is not available (AFAIU, that's what plasma panel uses for an indicator in a full desktop session, correct me if that's wrong).

Even if it's specific to KWin, there are no other widely used display server configs for sddm or commonly accepted interfaces for exposing layout switching. For the first release with Wayland greeter it seems to be good enough.

[...]

PR #1734, opened 2023-06-19
bump version to 0.20.0

hsitter:
in preparation of release


davidedmundson (approved):

Vogtinator:
I guess we should create a formal issue for 0.20 tasks?

hsitter:
Are there any tasks left for 0.20?

Vogtinator:
The issue with kwin_wayland not starting properly half the time after a logout needs to be investigated at least.
Or running the greeter with wayland gets marked as experimental explicitly...

Vogtinator:
> The issue with kwin_wayland not starting properly half the time after a logout needs to be investigated at least. Or running the greeter with wayland gets marked as experimental explicitly...

With that there's also the issue that keyboard layout switching is implemented wrongly (https://github.com/sddm/sddm/issues/1669)

hsitter:
Started https://github.com/sddm/sddm/issues/1735 David and I are thinking that wayland support should probably be treated as experimental for its first release anyway.

Pointedstick:
@sitter Agreed.

PR #1738, opened 2023-06-22
update changelog for 0.20

hsitter:


ChangeLog
```diff
@@ -4,6 +4,30 @@
     - Bug fixes
     * Others
 
+## 0.20.0 - 2023-06-22
+----------------------
+    + Initial Qt6 support
+    + **Experimental** support for Wayland
+    + Enable HiDPI scaling by default
+    + X11 display server without root privileges
```
Vogtinator: Sounds like it would be enabled by default

ChangeLog
```diff
@@ -4,6 +4,30 @@
     - Bug fixes
     * Others
 
+## 0.20.0 - 2023-06-22
+----------------------
+    + Initial Qt6 support
```
Vogtinator:
```suggestion
    + Initial Qt6 support (Will break themes which rely on Qt 5)
```

or something like that

ChangeLog
```diff
@@ -4,6 +4,30 @@
...
+    + **Experimental** support for Wayland
+    + Enable HiDPI scaling by default
+    + X11 display server without root privileges
+    + Greeter: Support setting environment variables
+    + Allow addition env vars to be defined in session files (#1370)
```
Vogtinator: Let's avoid the typo here

```suggestion
    + Allow additional env vars to be defined in session files (#1370)
```

ChangeLog
```diff
@@ -4,6 +4,30 @@
...
+    + Make accountsservice data directory overridable via CMake
+    + Add support for xcursor size configuration
```
Vogtinator: IMO more user friendly:

```suggestion
    + Add support for X11 cursor size configuration
```

Vogtinator (commented):
Some things that should probably be mentioned as well:

- Switch to using libxau with `FamilyWild` (#1230)
- New interface to access config values from themes (#1097)
- Session names are translated now (#1645)

ChangeLog
```diff
@@ -4,6 +4,30 @@
     - Bug fixes
     * Others
 
+## 0.20.0 - 2023-06-22
+----------------------
+    + Initial Qt6 support
+    + **Experimental** support for Wayland
```
Vogtinator:
```suggestion
    + **Experimental** support for running the greeter with Wayland
```

Vogtinator (approved):

PR #1739, opened 2023-06-23
Support multiple theme directories.

### LINKED ISSUE #1561: Enhancement: Support themes installed in two file system locations
Add support for using themes installed in two different file system locations instead of a single one right now.

The objective is to be able to define the two following locations:
- `/usr/share/sddm/themes`: themes installed by packages from Linux distributions
- `/var/lib/sddm/themes`: themes installed by a local user/administrator

This is especially useful on Linux distributions where `/usr` is mounted read only such as Fedora Kinoite.

Will help fix: https://bugs.kde.org/show_bug.cgi?id=454509

Thanks!

alebastr:
Interpret `ThemeDir` option as a comma-separated list of directories. Similar to the recent change in `SessionDir` handling, the option name is preserved for compatibility.

Fixes #1561 

***

This probably should default to `QStandardPaths::locateAll(QStandardPaths::AppDataLocation, QLatin1String("themes"), QStandardPaths::LocateDirectory);` (filtering out `~/*`), but it seems a bit too much of a change when sddm is close to release.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -189,6 +189,13 @@ set(DBUS_CONFIG_FILENAME        "org.freedesktop.DisplayManager.conf"           
 set(COMPONENTS_TRANSLATION_DIR  "${DATA_INSTALL_DIR}/translations"                  CACHE PATH      "Components translations directory")
 set(SDDM_INITIAL_VT             "1"                                                 CACHE STRING    "Initial tty to use")
 
+# Intentionally does not use CMAKE_INSTAL_PREFIX, just like SessionDir in the Configuration.h
+set(DEFAULT_THEME_DIRS "/usr/local/share/sddm/themes" "/usr/share/sddm/themes")
+if (NOT "${DATA_INSTALL_DIR}/themes" IN_LIST DEFAULT_THEME_DIRS)
+    list(APPEND DEFAULT_THEME_DIRS "${DATA_INSTALL_DIR}/themes")
+endif()
+string(REPLACE ";" "," DEFAULT_THEME_DIRS "${DEFAULT_THEME_DIRS}")
+set(THEME_DIRS                  "${DEFAULT_THEME_DIRS}"                             CACHE STRING    "Default theme search directories (comma-separated)")
 
 # Autodetect UID_MIN and UID_MAX from /etc/login.defs
 if(NOT DEFINED LOGIN_DEFS_PATH)
--- a/data/man/sddm.conf.rst.in
+++ b/data/man/sddm.conf.rst.in
@@ -79,8 +79,8 @@ OPTIONS
 [Theme] section:
 
 `ThemeDir=`
-	Path of the directory containing theme files.
-	Default value is "@DATA_INSTALL_DIR@/themes".
+	Comma-separated list of directories containing theme files.
+	Default value is "@THEME_DIRS@".
 
 `Current=`
 	Name of the current theme.
--- a/src/common/Configuration.h
+++ b/src/common/Configuration.h
@@ -50,7 +50,7 @@ namespace SDDM {
         Entry(GreeterEnvironment,  QStringList, QStringList(),                                  _S("Comma-separated list of environment variables to be set"));
         //  Name   Entries (but it's a regular class again)
         Section(Theme,
-            Entry(ThemeDir,            QString,     _S(DATA_INSTALL_DIR "/themes"),             _S("Theme directory path"));
+            Entry(ThemeDir,            QStringList, _S(THEME_DIRS).split(u','),                 _S("Comma-separated list of directories containing themes"));
             Entry(Current,             QString,     _S(""),                                     _S("Current theme name"));
             Entry(FacesDir,            QString,     _S(DATA_INSTALL_DIR "/faces"),              _S("Global directory for user avatars\n"
                                                                                                    "The files should be named <username>.face.icon"));
--- a/src/common/Constants.h.in
+++ b/src/common/Constants.h.in
@@ -29,6 +29,7 @@
 #define RUNTIME_DIR                 "@RUNTIME_DIR@"
 #define STATE_DIR                   "@STATE_DIR@"
 #define ACCOUNTSSERVICE_DATA_DIR    "@ACCOUNTSSERVICE_DATA_DIR@"
+#define THEME_DIRS                  "@THEME_DIRS@"
 
 #define SESSION_COMMAND             "@SESSION_COMMAND@"
 #define WAYLAND_SESSION_COMMAND     "@WAYLAND_SESSION_COMMAND@"
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -339,11 +339,12 @@ namespace SDDM {
         if (themeName.isEmpty())
             return QString();
 
-        QDir dir(mainConfig.Theme.ThemeDir.get());
-
-        // return the default theme if it exists
-        if (dir.exists(themeName))
-            return dir.absoluteFilePath(themeName);
+        // return the configured theme if it exists
+        for (const auto& themeDir: mainConfig.Theme.ThemeDir.get()) {
+            QDir dir(themeDir);
+            if (dir.exists(themeName))
+                return dir.absoluteFilePath(themeName);
+        }
 
         // otherwise use the embedded theme
         qWarning() << "The configured theme" << themeName << "doesn't exist, using the embedded theme instead";
--- a/src/greeter/UserModel.cpp
+++ b/src/greeter/UserModel.cpp
@@ -65,9 +65,14 @@ namespace SDDM {
 
     UserModel::UserModel(bool needAllUsers, QObject *parent) : QAbstractListModel(parent), d(new UserModelPrivate()) {
         const QString facesDir = mainConfig.Theme.FacesDir.get();
-        const QString themeDir = mainConfig.Theme.ThemeDir.get();
         const QString currentTheme = mainConfig.Theme.Current.get();
-        const QString themeDefaultFace = QStringLiteral("%1/%2/faces/.face.icon").arg(themeDir).arg(currentTheme);
+        QString themeDefaultFace;
+        for (const auto& themeDir: mainConfig.Theme.ThemeDir.get()) {
+            if (QDir(themeDir).exists(currentTheme)) {
+                themeDefaultFace = QStringLiteral("%1/%2/faces/.face.icon").arg(themeDir).arg(currentTheme);
+                break;
+            }
+        }
         const QString defaultFace = QStringLiteral("%1/.face.icon").arg(facesDir);
         const QString iconURI = QStringLiteral("file://%1").arg(
                 QFile::exists(themeDefaultFace) ? themeDefaultFace : defaultFace);


Vogtinator:
LGTM, but I wonder whether this is really useful without using this functionality by default.

Currently if something touches the configuration to set `Theme=`, it can already set the `ThemeDir` to wherever it is.

Vogtinator:
> This probably should default to QStandardPaths::locateAll(QStandardPaths::AppDataLocation, QLatin1String("themes"), QStandardPaths::LocateDirectory); (filtering out ~/*), but it seems a bit too much of a change when sddm is close to release.

If ~ is filtered out anyway this can just be hardcoded to `CMAKE_INSTALL_PREFIX "/local/share/sddm/themes/," CMAKE_INSTALL_PREFIX "/share/sddm/themes/"` or similar.

alebastr:
> LGTM, but I wonder whether this is really useful without using this functionality by default.
> 
> Currently if something touches the configuration to set `Theme=`, it can already set the `ThemeDir` to wherever it is.

A reasonable value could be something that resolves into `/usr/local/share/sddm/themes,/usr/share/sddm/themes`. The problem is, which one of those is `DATA_INSTALL_DIR "/themes"`? Does it have to be a third entry in the list? 
With immutable variants of Fedora we know the installation path and I can ask the maintainer to add more directories to the default config patch. I didn't want to make that guess for all supported systems.

> If ~ is filtered out anyway this can just be hardcoded to `CMAKE_INSTALL_PREFIX "/local/share/sddm/themes/," CMAKE_INSTALL_PREFIX "/share/sddm/themes/"` or similar.

There's a small but important difference: `QStandardPaths::AppDataLocation`/`QStandardPaths::GenericDataLocation` will follow `XDG_DATA_DIRS` if set. Although it's more important for `SessionDir`.

Vogtinator:
> > LGTM, but I wonder whether this is really useful without using this functionality by default.
> > Currently if something touches the configuration to set `Theme=`, it can already set the `ThemeDir` to wherever it is.
> 
> A reasonable value could be something that resolves into `/usr/local/share/sddm/themes,/usr/share/sddm/themes`. The problem is, which one of those is `DATA_INSTALL_DIR "/themes"`?

Depends on `CMAKE_INSTALL_PREFIX`. That can be `/usr`, `/usr/local` or something like `/opt/work/sddm-git-builds/foobar`.

> Does it have to be a third entry in the list? With immutable variants of Fedora we know the installation path and I can ask the maintainer to add more directories to the default config patch. I didn't want to make that guess for all supported systems.
> 
> > If ~ is filtered out anyway this can just be hardcoded to `CMAKE_INSTALL_PREFIX "/local/share/sddm/themes/," CMAKE_INSTALL_PREFIX "/share/sddm/themes/"` or similar.
> 
> There's a small but important difference: `QStandardPaths::AppDataLocation`/`QStandardPaths::GenericDataLocation` will follow `XDG_DATA_DIRS` if set. Although it's more important for `SessionDir`.

The issue with that is that following `XDG_DATA_DIRS` can only happen if `ThemeDir` is not set. If it's set, there's no way to indicate that `XDG_DATA_DIRS` should be followed anyway, at least not if it's just used as default value.

muelsyse-az:
Hello, what's the status of this PR?

Vogtinator:
> Hello, what's the status of this PR?

Waiting for my comment to be addressed mostly.

What's your specific usecase for this? Might help in finding a proper default value.

muelsyse-az:
> > Hello, what's the status of this PR?
> 
> 
> 
> Waiting for my comment to be addressed mostly.
> 
> 
> 
> What's your specific usecase for this? Might help in finding a proper default value.


I just wanted to try out Global Themes on Fedora Kinoite (Immutable/Atomic KDE).

With Fedora 40 and Plasma 6 nearing, I was wondering whether or not a fix was already done for the issue where you can't install Global Themes on Immutable/Atomic Desktops. It's not urgent as I've been comfortably using Breeze but wanted to see if there was anyone actively working on the issue.

fenglengshun:
To add, the immutability of `/usr` has caused Global Themes install to not succeed due to some of them including SDDM themes in it. Ideally, it should be someplace in `/var` or `/home` (well, strictly speaking, `/home` is just mounted from `/var` anyways). I don't think `/etc` is the correct directory for the themes, though.

Vogtinator:
> With Fedora 40 and Plasma 6 nearing, I was wondering whether or not a fix was already done for the issue where you can't install Global Themes on Immutable/Atomic Desktops. It's not urgent as I've been comfortably using Breeze but wanted to see if there was anyone actively working on the issue.

That's an issue with kcm_sddm, it cannot be addressed in sddm.

fenglengshun:
@Vogtinator regarding Global Themes, true. The main thing is that I'd like to have a way to add sddm themes without relying on making a package and layering that package due to `/usr` immutability.

I'd say either `/etc/sddm` or `/var/usrlocal/share/sddm` make sense as a new default if you want to account for immutable distro - I don't think `$HOME/.local/share/sddm` make sense as a default, but as a non-default it could be added.

Once this is done, we can then move to `kcm_sddm` for either adding directory choice to install `sddm` themes in, or just request for the default to be changed to mutable `/var`, `/etc`, or `/home` directories.

Vogtinator:
> @Vogtinator regarding Global Themes, true. The main thing is that I'd like to have a way to add sddm themes without relying on making a package and layering that package due to `/usr` immutability.

You can already do that.

Just put the theme into , `/etc/sddm-themes/whatever`/`/var/lib/sddm-themes/whatever` and set

```
[Theme]
ThemeDir=/var/lib/sddm-themes/
Current=whatever
```

Heus-Sueh:
> > @Vogtinator regarding Global Themes, true. The main thing is that I'd like to have a way to add sddm themes without relying on making a package and layering that package due to `/usr` immutability.
> 
> You can already do that.
> 
> Just put the theme into , `/etc/sddm-themes/whatever`/`/var/lib/sddm-themes/whatever` and set
> 
> ```
> [Theme]
> ThemeDir=/var/lib/sddm-themes/
> Theme=whatever
> ```

Is this possible in the current version?
image

Vogtinator:
> > > @Vogtinator regarding Global Themes, true. The main thing is that I'd like to have a way to add sddm themes without relying on making a package and layering that package due to `/usr` immutability.
> > 
> > 
> > You can already do that.
> > Just put the theme into , `/etc/sddm-themes/whatever`/`/var/lib/sddm-themes/whatever` and set
> > ```
> > [Theme]
> > ThemeDir=/var/lib/sddm-themes/
> > Theme=whatever
> > ```
> 
> Is this possible in the current version? image

It's not `Theme=`, but `Current=`, I edited my comment. Other than that it works like that in pretty much any version.

You should put the theme somewhere outside of your home directory for security reasons.

Heus-Sueh:
In relation to security, do themes really have the power to cause such danger while being on the user's home? Has this already been discussed here or elsewhere?

Vogtinator:
> In relation to security, do themes really have the power to cause such danger while being on the user's home?

The concern here is that the theme code is executed as the sddm user but another unprivileged user has write permissions.

Heus-Sueh:
> > In relation to security, do themes really have the power to cause such danger while being on the user's home?
> 
> The concern here is that the theme code is executed as the sddm user but another unprivileged user has write permissions.

I see, thanks for the clarification

CMakeLists.txt
```diff
@@ -189,6 +189,13 @@ set(DBUS_CONFIG_FILENAME        "org.freedesktop.DisplayManager.conf"
 set(COMPONENTS_TRANSLATION_DIR  "${DATA_INSTALL_DIR}/translations"                  CACHE PATH      "Components translations directory")
 set(SDDM_INITIAL_VT             "1"                                                 CACHE STRING    "Initial tty to use")
 
+# Intentionally does not use CMAKE_INSTAL_PREFIX, just like SessionDir in the Configuration.h
+set(DEFAULT_THEME_DIRS "/usr/local/share/sddm/themes" "/usr/share/sddm/themes")
```
travier, line 193: I think we should try `/usr` first, and then `/var/lib/sddm/themes`:

```suggestion
set(DEFAULT_THEME_DIRS "${CMAKE_INSTALL_FULL_DATADIR}/sddm/themes" "${CMAKE_INSTALL_FULL_LOCALSTATEDIR}/lib/sddm/themes")
```

CMakeLists.txt
```diff
@@ -189,6 +189,13 @@ set(DBUS_CONFIG_FILENAME        "org.freedesktop.DisplayManager.conf"
...
+if (NOT "${DATA_INSTALL_DIR}/themes" IN_LIST DEFAULT_THEME_DIRS)
+    list(APPEND DEFAULT_THEME_DIRS "${DATA_INSTALL_DIR}/themes")
```
travier, line 195: Not needed with above change.

CMakeLists.txt
```diff
@@ -189,6 +189,13 @@ set(DBUS_CONFIG_FILENAME        "org.freedesktop.DisplayManager.conf"
...
+endif()
+string(REPLACE ";" "," DEFAULT_THEME_DIRS "${DEFAULT_THEME_DIRS}")
+set(THEME_DIRS                  "${DEFAULT_THEME_DIRS}"                             CACHE STRING    "Default theme search directories (comma-separated)")
```
travier, line 198: Code from above can be merged here.

travier (commented):
Suggestion for the default paths to set

Vogtinator (changes requested):
SDDM does not necessarily need support for multiple theme directories, but for having theme.conf.user in a separate location.

travier:
> SDDM does not necessarily need support for multiple theme directories, but for having theme.conf.user in a separate location.

I don't understand. On some systems (Fedora Kinoite for example), the location for system installed themes is read only, so we do need another location to write user installed themes.

Vogtinator:
> > SDDM does not necessarily need support for multiple theme directories, but for having theme.conf.user in a separate location.
> 
> I don't understand. On some systems (Fedora Kinoite for example), the location for system installed themes is read only, so we do need another location to write user installed themes.

Just install the theme somewhere writable (e.g. in `/var/`) and set `ThemeDir=`.

travier:
From https://pagure.io/fedora-kde/SIG/issue/282#comment-916293:

> If we change the default config to point to `/var/lib/sddm/themes`, then we need all the defaults themes to be there as well.
>
> Say we copy them there on first boot, then appart from the fact that this duplicate files, this will create issues on updates. If the user modifies any of the config / files that are part of the theme, then those modifications are lost on updates as we re-copy files again.
>
> If we don't update the themes on updates then users will likely face subtle bugs as their theme go outdated and mis-match the version of other packages.
>
> We can not install the themes files directly on `/var` either as this is local state, that is not available at compose time for Atomic variants and overall in conflict with the policy that all distribution content should live in `/usr`.
>
> Moreover, we want the defaults themes to be read-only and automatically updated with the rest of the system to make sure that there is always a working fallback if something goes wrong.

Vogtinator:
System-provided themes go into `/usr`, user provided ones into `/var` or elsewhere, just set `ThemeDir` according to the configured one. We could also make `Current` take an absolute path to the theme, which is the exact same just slightly more convenient.

The only issue is theme.conf.user for system themes.

v3xed-v0id:
i hope that this change goes through and gets merged, this is a desperately needed feature on basically every immutable distro, sddm themes dont work correctly on them without this fix, i try on immutable fedora and have issues so i hope this change goes through, thanks for reading.

Vogtinator:
>  hope that this change goes through and gets merged, this is a desperately needed feature on basically every immutable distro, sddm themes dont work correctly on them without this fix, i try on immutable fedora and have issues so i hope this change goes through, thanks for reading.

It's not needed - sddm already supports arbitrary locations for themes. It just needs to be configured properly for immutable distros, which is up to the distro and/or desktop environment.

KDE Plasma's sddm-kcm hardcodes `/usr/share/sddm/themes`, which is the issue here.

v3xed-v0id:
> It's not needed - sddm already supports arbitrary locations for themes. It just needs to be configured properly for immutable distros, which is up to the distro and/or desktop environment.
> 
> KDE Plasma's sddm-kcm hardcodes `/usr/share/sddm/themes`, which is the issue here.

Interesting, i was not aware, i hope that sddm-kcm fixes it then

luvis:
I started using Kinoite just yesterday and ran into this pretty quickly. I understand it’s tied to the immutable setup but tons of other settings are able to stick. KDE, or rather Linux in general is known for its customization, so I find it surprising that we can’t easily change the SDDM background or apply a custom theme and have it persist in a straight-forward way. 

From what I’ve seen, there are some workarounds, so clearly it's possible. Wouldn't it be reasonable for the system to support something like, “if this directory exists, load settings from here instead of the defaults”?

I’m not a developer, so I might be missing something — but I’m just wondering why a cleaner, more persistent approach hasn’t been built in by now. It seems like a problem that could have a more straightforward solution, especially with how long it's been around.

Vogtinator:
> Wouldn't it be reasonable for the system to support something like, “if this directory exists, load settings from here instead of the defaults”?

SDDM already does that for its config files.

It's the SDDM KCM that needs to install themes outside of the "immutable" part of the OS.

luvis:
> > Wouldn't it be reasonable for the system to support something like, “if this directory exists, load settings from here instead of the defaults”?
> 
> SDDM already does that for its config files.
> 
> It's the SDDM KCM that needs to install themes outside of the "immutable" part of the OS.

I see. so for now the only way to install custom themes without all the hassle is to set the themedir in a conf file in /etc/sddm.conf.d/?
It took me quite a while to understand this as I'm no power user and most discussions revolve around very complicated procedures imo. Why can't SDDM KCM just install the themes in an /etc/sddm/themes folder or for that matter just use that folder by default, effectively fixing this issue? I'm sorry if this has already been addressed but I feel like this is a super easy fix, can't any immutable system just have a separate configuration by default?

Vogtinator:
> Why can't SDDM KCM just install the themes in an /etc/sddm/themes folder or for that matter just use that folder by default, effectively fixing this issue?

It could, but currently it just doesn't. MRs welcome I guess: https://invent.kde.org/plasma/sddm-kcm

Idesmi:
> KDE Plasma's sddm-kcm hardcodes `/usr/share/sddm/themes`, which is the issue here.

As of today, this is not the case any more. SDDM KCM installs new themes in `QStandardPaths::locate(QStandardPaths::GenericDataLocation, QStringLiteral("sddm/themes"), QStandardPaths::LocateDirectory)`  
(see https://invent.kde.org/plasma/sddm-kcm/-/blob/f07f5b9e0ec040dc257c832bc2cf5d34c72e0aeb/sddmauthhelper.cpp#L378)

Hence it looks in `~/.local/share/`, `/usr/local/share/` and `/usr/share/` for `sddm/themes`.  
To be written into, `~/.local/share/sddm/themes/` or `/usr/local/share/sddm/themes/` must exist, even if empty.

Idesmi:
I have opened a MR for sddm-kcm: https://invent.kde.org/plasma/sddm-kcm/-/merge_requests/95
Relevant issue on openSUSE Bugzilla: https://bugzilla.opensuse.org/show_bug.cgi?id=1208163

PR #1743, opened 2023-06-25
gitignore: Drop filter on "user" files

Conan-Kudo:
This causes git-archive(1) to filter out the sysuser files from the tarball.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 /build/
 *~
-*.user


Vogtinator:
Which files in particular? FWICT this should only match files ending in `.user` which there aren't any

Conan-Kudo:
The tarball was missing `services/sddm-sysuser.conf.in`, which led to a broken build. https://koji.fedoraproject.org/koji/taskinfo?taskID=102500963

```
CMake Error: File /builddir/build/BUILD/sddm-0.20.0/services/sddm-sysuser.conf.in does not exist.
CMake Error at services/CMakeLists.txt:5 (configure_file):
  configure_file Problem configuring file
```

Vogtinator:
Sounds like a bug somewhere. If the `.gitignore` file would actually match that, it would not be part of the repo.
`git archive` also includes it here:

```
> git archive -v -o test.tar v0.20.0 |& grep user; tar tvaf test.tar | grep user
data/themes/elarun/images/system_switch_user.png
data/themes/elarun/images/user_icon.png
services/sddm-sysuser.conf.in
src/helper/xorguserhelper.cpp
src/helper/xorguserhelper.h
-rw-rw-r-- root/root      1175 2023-06-23 14:28 data/themes/elarun/images/system_switch_user.png
-rw-rw-r-- root/root      1239 2023-06-23 14:28 data/themes/elarun/images/user_icon.png
-rw-rw-r-- root/root       112 2023-06-23 14:28 services/sddm-sysuser.conf.in
-rw-rw-r-- root/root      7949 2023-06-23 14:28 src/helper/xorguserhelper.cpp
-rw-rw-r-- root/root      1888 2023-06-23 14:28 src/helper/xorguserhelper.h
```

Vogtinator:
Where did you get the broken tarball from? https://src.fedoraproject.org/rpms/sddm/blob/rawhide/f/sddm.spec mentions

```
# Broken: https://github.com/sddm/sddm/pull/1743
#Source0:        %{url}/releases/download/v%{version}/%{name}-%{version}.tar.xz
```
But that's 404.

Conan-Kudo:
It looks like they were deleted from https://github.com/sddm/sddm/releases/tag/v0.20.0, but they were there two days ago.

Vogtinator:
Maybe because of that?

CC @hsitter

hsitter:
Yep, I've thrown away the manual tarballs because they were missing files. That said, I don't really think we need them, all they did was remove stuff making us have two divergent tarball variants.

FWIW the offending line is rather https://github.com/sddm/sddm/blob/develop/CMakeLists.txt#L22 I think

With that in mind I'm kinda leaning towards throwing away the dist target.

Vogtinator:
> Yep, I've thrown away the manual tarballs because they were missing files. That said, I don't really think we need them, all they did was remove stuff making us have two divergent tarball variants.
> 
> With that in mind I'm kinda leaning towards throwing away the dist target.

Agreed.

> FWIW the offending line is rather https://github.com/sddm/sddm/blob/develop/CMakeLists.txt#L22 I think

Yep, missing a `$`.

PR #1746, opened 2023-06-26
Session: Parse .desktop files manually again

Vogtinator:
Using QSettings::IniFormat doesn't quite work. Implement a custom parser for those files to handle them according to the specification.

Fixes #1745

PR #1747, opened 2023-06-26
drop dist target

hsitter:
it serves no real purpose since auto-generated release artifacts are expected to work correctly, so anything we do in the dist target only serves to diverge from the tag and potentially break things


Vogtinator (approved):

Vogtinator:
I'd merge, but the xdg check removed by #1746 is still marked as required...

PR #1752, opened 2023-06-30
OS Name and Version as Theme Properties

Videowelder:
Working on a theme that displays the detected OS as a graphical design element, much like some distros did with lightdm in the past.
Figured these properties would be of use on the official project for other theme designers.
These use built-in Qt stuff, so wide support and no maintenance required.

### DIFF
--- a/docs/THEMING.md
+++ b/docs/THEMING.md
@@ -21,6 +21,10 @@ We provide a proxy object, called as `sddm` to the themes as a context property.
 
 **hostName:** Holds the name of the host computer.
 
+**hostOsName:** Holds the operating system product name. "unknown", if not a recognised OS.
+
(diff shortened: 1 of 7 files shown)


Vogtinator (changes requested):
I think this can be implemented solely in the greeter, no need to touch the daemon at all

Vogtinator:
IMO `QSysInfo::prettyProductName()` is the most important one and should be available as well.

Then we have three new properties, at which point it might be a good idea to use an object with properties instead, allowing for use from QML like this:

```
console.log("OS: " + sddm.os.prettyName + " Version: " + sddm.os.version)
```

Vogtinator:
No feedback provided for over a year, closing. Please reopen or open a new PR if there's still interest.

PR #1756, opened 2023-07-06
Deal with absolute paths to log files

mikhailnov:
If path begins with "/", consider it to be an absolute path, not relative to $HOME

Example: SessionLogFile=/dev/null

Co-authored-by: Artem Proskurnev <ProskurnevAS@edu.mos.ru> (@temaps)
Signed-off-by: Mikhail Novosyolov <m.novosyolov@rosalinux.ru>

### DIFF
--- a/src/helper/UserSession.cpp
+++ b/src/helper/UserSession.cpp
@@ -353,11 +353,19 @@ namespace SDDM {
             //we want to redirect after we setuid so that the log file is owned by the user
 
             // determine stderr log file based on session type
-            QString sessionLog = QStringLiteral("%1/%2")
-                    .arg(QString::fromLocal8Bit(pw.pw_dir))
-                    .arg(sessionType == QLatin1String("x11")
-                         ? mainConfig.X11.SessionLogFile.get()
-                         : mainConfig.Wayland.SessionLogFile.get());
+            QString sessionLog;
+            if (sessionType == QLatin1String("wayland"))
+            {
+                sessionLog = mainConfig.Wayland.SessionLogFile.get();
+            } else {
+                sessionLog = mainConfig.X11.SessionLogFile.get();
+            }
+            if (sessionLog.at(0) != QLatin1Char('/'))
+            {
+                sessionLog = QStringLiteral("%1/%2")
+                             .arg(QString::fromLocal8Bit(pw.pw_dir))
+                             .arg(sessionLog);
+            }
 
             // create the path
             QFileInfo finfo(sessionLog);
--- a/data/man/sddm.conf.rst.in
+++ b/data/man/sddm.conf.rst.in
@@ -135,7 +135,8 @@ OPTIONS
 	Default value is "@SESSION_COMMAND@".
 
 `SessionLogFile=`
-        Path to the user session log file, relative to the home directory.
+        Path to the user session log file.
+        Relative to the home directory, if does not begin with "/".
         Default value is ".local/share/sddm/xorg-session.log".
 
 `DisplayCommand=`
@@ -183,7 +184,8 @@ The `UserAuthFile=` option was removed, the file is always created as
 	Default value is "@WAYLAND_SESSION_COMMAND@".
 
 `SessionLogFile=`
-        Path to the user session log file, relative to the home directory.
+        Path to the user session log file.
+        Relative to the home directory, if does not begin with "/".
         Default value is ".local/share/sddm/wayland-session.log".
 
 `EnableHiDPI=`


src/helper/UserSession.cpp
```diff
@@ -353,11 +353,20 @@ namespace SDDM {
             //we want to redirect after we setuid so that the log file is owned by the user
 
             // determine stderr log file based on session type
-            QString sessionLog = QStringLiteral("%1/%2")
-                    .arg(QString::fromLocal8Bit(pw.pw_dir))
-                    .arg(sessionType == QLatin1String("x11")
-                         ? mainConfig.X11.SessionLogFile.get()
-                         : mainConfig.Wayland.SessionLogFile.get());
+            QString sessionLog;
+            if ( (sessionType == QLatin1String("x11")) ||
```
Vogtinator: You can invert this and check for `wayland` instead

Vogtinator:
LGTM but I wonder whether there is any use for absolute paths other than `/dev/null`. All users would end up with the same absolute path to the log file so they would just overwrite each other.

src/helper/UserSession.cpp
mikhailnov: Done

mikhailnov:
> LGTM but I wonder whether there is any use for absolute paths other than `/dev/null`. All users would end up with the same absolute path to the log file so they would just overwrite each other.

It's a good question, but maybe someone would want to configure autologin and write logs to somewhere ouside of user home directory.

Actually I did all that to set X11.SessionCommand "/usr/bin/systemd-cat /usr/share/X11/xdm/Xsession" by default in ROSA Linux package (https://abf.io/import/sddm). I dislike logging into files and want logs to be stored in systemd-journald as in gdm.

Vogtinator:
> Actually I did all that to set X11.SessionCommand "/usr/bin/systemd-cat /usr/share/X11/xdm/Xsession" by default in ROSA Linux package (https://abf.io/import/sddm). I dislike logging into files and want logs to be stored in systemd-journald as in gdm.

Sounds like a better feature. Just not redirecting at all should do the trick, because sddm already outputs into the journal.

mikhailnov:
Hm, thanks for noting this, I did not notice that sddm already outputs into the journal even without systemd-cat, maybe I tried an old version

Vogtinator:
> Hm, thanks for noting this, I did not notice that sddm already outputs into the journal even without systemd-cat, maybe I tried an old version

SDDM itself does, but for the session it redirects the output to a file so that is lost. It could be configurable whether it redirects at all.

Or sddm-helper could learn how to communicate with systemd to log into the user journal. Not sure how involved that would be.

mikhailnov:
Seems that my mind is messed up. systemd-cat does not make stderr in e.g. openbox to appear in journald. Byt it appears both with and wihtout it in KDE, probably because KDE does it itself (or by running apps via systemd).

Vogtinator:
> Byt it appears both with and wihtout it in KDE, probably because KDE does it itself (or by running apps via systemd).

Yes, Plasma itself and everything started by it runs as/within systemd units

mikhailnov:
Which DE can it be tested with? I tried openbox, SessionLogFile (not /dev/null) is empty...

src/helper/UserSession.cpp
mikhailnov: A colleague of mine noted that this is session type (x11 or wayland), not x11/x11-user/wayland, sorry. I've fixed the code and the commit message.

Vogtinator:
> Which DE can it be tested with? I tried openbox, SessionLogFile (not /dev/null) is empty...

Most small DEs should work, there any started application should have that as stdout+stderr

evelikov:
Thinking out loud:

On my Arch box sddm-helper is running as root. So I wonder if allowing for absolute path can backfire in some odd way - say my mistake you set `/home/user/some/path/file` where you effectively destroy some semi-meaningful file that belongs to others.

Does that make sense? Or I guess the proper way forward is to run all sddm programs as non-root?

Vogtinator:
> Thinking out loud:
> 
> On my Arch box sddm-helper is running as root. So I wonder if allowing for absolute path can backfire in some odd way - say my mistake you set `/home/user/some/path/file` where you effectively destroy some semi-meaningful file that belongs to others.

That's basically https://github.com/sddm/sddm/pull/1756#issuecomment-1623795850

> Does that make sense? Or I guess the proper way forward is to run all sddm programs as non-root?

The logfile is opened after setuid, so permission checks are done as the target user.

PR #1759, opened 2023-07-08
Let GreeterEnvironment option take the precedence over PAM environment

### LINKED ISSUE #1758: Should the PAM's default environment override the GreeterEnvironment option?
pam_putEnv doesn't override PAM's environment loaded with pam_env.so so [this](https://github.com/sddm/sddm/blob/07631f2ef00a52d883d0fd47ff7d1e1a6bc6358f/src/daemon/Greeter.cpp#L183-L191) and [this](https://github.com/sddm/sddm/blob/07631f2ef00a52d883d0fd47ff7d1e1a6bc6358f/src/helper/HelperApp.cpp#L170-L181) gets overriden thanks to [this](https://github.com/sddm/sddm/blob/07631f2ef00a52d883d0fd47ff7d1e1a6bc6358f/src/helper/backend/PamBackend.cpp#L276). Is all that intended or should the imported environment variables from sddm process and/or GreeterEnvironment options take the precedence?

ilya-fedin:
Fixes #1758

### DIFF
--- a/src/helper/HelperApp.cpp
+++ b/src/helper/HelperApp.cpp
@@ -167,19 +167,6 @@ namespace SDDM {
         m_user = m_backend->userName();
         QProcessEnvironment env = authenticated(m_user);
 
-        if (env.value(QStringLiteral("XDG_SESSION_CLASS")) == QLatin1String("greeter")) {
-            // Qt internally may load the xdg portal system early on, prevent this, we do not have a functional session running.
-            env.insert(QStringLiteral("QT_NO_XDG_DESKTOP_PORTAL"), QStringLiteral("1"));
-            for (const auto &entry : mainConfig.GreeterEnvironment.get()) {
-                const int index = entry.indexOf(QLatin1Char('='));
-                if (index < 0) {
-                    qWarning() << "Malformed environment variable" << entry;
-                    continue;
-                }
-                env.insert(entry.left(index), entry.mid(index + 1));
-            }
-        }
-
         if (!m_session->path().isEmpty()) {
             env.insert(m_session->processEnvironment());
             m_session->setProcessEnvironment(env);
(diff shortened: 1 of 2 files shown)


Vogtinator:
On one hand having some greeter specific code in backend.cpp is wrong, but on the other hand it's at least set in the same place as other environment variables.

The environment handling is spaghetti anyways so it doesn't really make a difference.

I'll have a closer look next week, but from a quick glance it looks good.

ilya-fedin:
> I'll have a closer look next week

ping? :thinking:

src/helper/HelperApp.cpp
```diff
@@ -167,19 +167,6 @@ namespace SDDM {
...
-                    qWarning() << "Malformed environment variable" << entry;
-                    continue;
-                }
-                env.insert(entry.left(index), entry.mid(index + 1));
-            }
-        }
-
         if (!m_session->path().isEmpty()) {
             env.insert(m_session->processEnvironment());
             m_session->setProcessEnvironment(env);
```
Vogtinator, line 172: What do you think about just moving those assignments here, after the `env.insert(m_session->processEnvironment());` call? That should have the same effect but is IMO a slightly more fitting place

ilya-fedin: Not entirely the same, `getpwnam` and `login_getpwclass` will have the priority over GreeterEnvironment but will still help with the exact case of QT_PLUGIN_PATH in PAM though.

Vogtinator: Yeah, I don't really know what `setclassenvironment` does on FreeBSD. Maybe it should override GreeterEnvironment, maybe it shouldn't...

ilya-fedin: I'd say it has sense for GreeterEnvironment to override anything else

Vogtinator: Ok.

Vogtinator (approved):

PR #1775, opened 2023-07-31
Handle autologin auth attempt errors

SgAkErRu:
If the autologin auth attempt failed for any reason, then start socket server and greeter (launch sddm theme, like as if there was no autologin at all).

### DIFF
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -394,12 +394,6 @@ namespace SDDM {
             qCritical() << "Failed to find command for session" << session.fileName();
             return false;
         }
-        // check if the user even exists
-        struct passwd *pw = getpwnam(qPrintable(user));
-        if (!pw) {
-            qCritical() << "Failed to find the user" << user;
-            return false;
-        }
 
         m_reuseSessionId = QString();
 
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -220,11 +220,20 @@ namespace SDDM {
             return false;
         }
 
+        const QString username = mainConfig.Autologin.User.get();
+
+        // check if the user even exists
+        struct passwd *pw = getpwnam(qPrintable(username));
+        if (!pw) {
+            qCritical() << "Failed to find the autologin user:" << username;
+            return false;
+        }
+
         Session session;
         session.setTo(sessionType, autologinSession);
 
         m_auth->setAutologin(true);
-        startAuth(mainConfig.Autologin.User.get(), QString(), session);
+        startAuth(username, QString(), session);
 
         return true;
     }
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -224,12 +224,7 @@ namespace SDDM {
         session.setTo(sessionType, autologinSession);
 
         m_auth->setAutologin(true);
-        const bool result = startAuth(mainConfig.Autologin.User.get(), QString(), session);
-
-        if (!result)
-            m_auth->setAutologin(false);
-
-        return result;
+        return startAuth(mainConfig.Autologin.User.get(), QString(), session);
     }
 
     void Display::startSocketServerAndGreeter() {
@@ -260,6 +255,12 @@ namespace SDDM {
         m_started = true;
     }
 
+    void Display::handleAutologinFailure() {
+        qWarning() << "Autologin failed!";
+        m_auth->setAutologin(false);
+        startSocketServerAndGreeter();
+    }
+
     void Display::displayServerStarted() {
         // check flag
         if (m_started)
@@ -279,12 +280,11 @@ namespace SDDM {
             // set flags
             m_started = true;
 
-            bool success = attemptAutologin();
-            if (success) {
-                return;
-            } else {
-                qWarning() << "Autologin failed!";
-            }
+            const bool autologinStarted = attemptAutologin();
+            if (!autologinStarted)
+                handleAutologinFailure();
+
+            return;
         }
 
         startSocketServerAndGreeter();
@@ -515,9 +515,7 @@ namespace SDDM {
         qWarning() << "Authentication error:" << error << message;
 
         if (m_auth->autologin()) {
-            qWarning() << "Autologin auth failed!";
-            m_auth->setAutologin(false);
-            startSocketServerAndGreeter();
+            handleAutologinFailure();
             return;
         }
 
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -397,7 +397,7 @@ namespace SDDM {
         // check if the user even exists
         struct passwd *pw = getpwnam(qPrintable(user));
         if (!pw) {
-            qCritical() << "Failed to find the autologin user" << user;
+            qCritical() << "Failed to find the user" << user;
             return false;
         }
 
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -220,22 +220,44 @@ namespace SDDM {
             return false;
         }
 
-        const QString username = mainConfig.Autologin.User.get();
-
-        // check if the user even exists
-        struct passwd *pw = getpwnam(qPrintable(username));
-        if (!pw) {
-            qCritical() << "Failed to find the autologin user:" << username;
-            return false;
-        }
-
         Session session;
         session.setTo(sessionType, autologinSession);
 
         m_auth->setAutologin(true);
-        startAuth(username, QString(), session);
+        const bool result = startAuth(mainConfig.Autologin.User.get(), QString(), session);
 
-        return true;
+        if (!result)
+            m_auth->setAutologin(false);
+
+        return result;
+    }
+
+    void Display::startSocketServerAndGreeter() {
+        // start socket server
+        m_socketServer->start(m_displayServer->display());
+
+        if (!daemonApp->testing()) {
+            // change the owner and group of the socket to avoid permission denied errors
+            struct passwd *pw = getpwnam("sddm");
+            if (pw) {
+                if (chown(qPrintable(m_socketServer->socketAddress()), pw->pw_uid, pw->pw_gid) == -1) {
+                    qWarning() << "Failed to change owner of the socket";
+                    return;
+                }
+            }
+        }
+
+        m_greeter->setSocket(m_socketServer->socketAddress());
+        m_greeter->setTheme(findGreeterTheme());
+
+        // start greeter
+        m_greeter->start();
+
+        // reset first flag
+        daemonApp->first = false;
+
+        // set flags
+        m_started = true;
     }
 
     void Display::displayServerStarted() {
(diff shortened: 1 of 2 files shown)


Vogtinator:
IMO this is the wrong place for such a check. If the auth attempt fails for any reason, it should be handled gracefully.

SgAkErRu:
It is attempt to autologin, why we need to try to auth, if we know right away that it won't work? A little higher in this function, `session` is checked in a similar way.

It seems to me there is no point in making such an attempt. But anyway I see problem with handling auth attempt after autologin in that `m_socketServer` (which, as it seemed to me, is engaged in handling the failed authorization signal) is not even launched before autologin attempt (but I can be wrong).

Also, proposed patch is similar [to this case](https://github.com/sddm/sddm/blob/4c3825f3e09d0b6043171c759f19d0eafd337b0f/src/daemon/Display.cpp#L327C45-L327C45) (there we don't even try to log in if it doesn't make sense).

SgAkErRu:
But I think I can trying to move that check into [`Display::startAuth`](https://github.com/sddm/sddm/blob/4c3825f3e09d0b6043171c759f19d0eafd337b0f/src/daemon/Display.cpp#L377), following the example of sanity session checks there. 

What do you think about this idea?

Vogtinator:
> It is attempt to autologin, why we need to try to auth, if we know right away that it won't work? A little higher in this function, `session` is checked in a similar way.

That's actually for a different reason: The session has to be found to set `sessionType` correctly.

> It seems to me there is no point in making such an attempt. But anyway I see problem with handling auth attempt after autologin in that m_socketServer (which, as it seemed to me, is engaged in handling the failed authorization signal) is not even launched before autologin attempt (but I can be wrong).

If that's the case, that should probably be changed. IMO autologin should be as close to regular login as possible.

> Also, proposed patch is similar [to this case](https://github.com/sddm/sddm/blob/4c3825f3e09d0b6043171c759f19d0eafd337b0f/src/daemon/Display.cpp#L327C45-L327C45) (there we don't even try to log in if it doesn't make sense).

Yeah, but that's also a case of doing an early check. Even if you remove this condition, it'll behave the same way. Login will just be refused by the PAM backend instead of the daemon.

> But I think I can trying to move that check into [Display::startAuth](https://github.com/sddm/sddm/blob/4c3825f3e09d0b6043171c759f19d0eafd337b0f/src/daemon/Display.cpp#L377), following the example of sanity session checks there.
> 
> What do you think about this idea?

Not wrong, but IMO it's still important to handle autologin auth failures in any case, not just for wrong usernames.

Vogtinator:
The logic right now looks good, but what I don't really like is that there are now two ways autologin can fail: Synchronosly (returns false) and asynchronously (auth failure signal). Any idea how this could be unified? That avoids subtle errors like somehow ending up without `m_auth->setAutologin(false);` in some weird circumstance.

SgAkErRu:
> ...two ways autologin can fail: Synchronosly (returns false) and asynchronously (auth failure signal). Any idea how this could be unified?

I think, still, they have a different nature: synchronous failure is a failure of early checks, and async failure is the authorization failure. 

But, I think we can try to reduce the difference between them, if we put failure handling in a separate function (i mean that handling):
```
qWarning() << "Autologin failed!";
m_auth->setAutologin(false);
startSocketServerAndGreeter();
```
And will use that function in `Display::slotAuthError` and `Display::attemptAutologin()` directly, without returning boolean from `Display::attemptAutologin()` in `Display::displayServerStarted()`, so it will be just:
```
attemptAutologin();
return;
```
instead of
```
bool success = attemptAutologin();
if (success) {
    return;
} else {
    qWarning() << "Autologin failed!";
}
 ```
in `Display::displayServerStarted()` function.

**Comment by Vogtinator** [comment]
> > ...two ways autologin can fail: Synchronosly (returns false) and asynchronously (auth failure signal). Any idea how this could be unified?
> 
> I think, still, they have a different nature: synchronous failure is a failure of early checks, and async failure is the authorization failure.

IMO that's an "implementation detail" (as far as that is possible inside the same class). The result is the same. I'm thinking along the lines of emitting `slotAuthError` within `attemptAutologin` for instance, though that can be misleading in other ways.

> But, I think we can try to reduce the difference between them, if we put failure handling in a separate function (i mean that handling):
> 
> ```
> qWarning() << "Autologin failed!";
> m_auth->setAutologin(false);
> startSocketServerAndGreeter();
> ```
> 
> And will use that function in `Display::slotAuthError` and `Display::attemptAutologin()` directly, without returning boolean from `Display::attemptAutologin()` in `Display::displayServerStarted()`, so it will be just:
> 
> ```
> attemptAutologin();
> return;
> ```
> 
> instead of
> 
> ```
> bool success = attemptAutologin();
> if (success) {
>     return;
> } else {
>     qWarning() << "Autologin failed!";
> }
> ```
> 
> in `Display::displayServerStarted()` function.

Ok.

**Comment by SgAkErRu** [comment]
> I'm thinking along the lines of emitting `slotAuthError` within `attemptAutologin` for instance, though that can be misleading in other ways.

Yes, I agree, because a failure of early checks didn't perform authorization process (through `m_auth`), so it can be misleading indeed.

I have put handling in separate function, but I have decided to leave `Display::attemptAutologin()` returning boolean, because it can return `false` if `unable to find autologin session entry`. So, `true` value of `Display::attemptAutologin()`  just means that autologin has been started and this is expressed in the name of the variable `autologinStarted` instead of `success`. I think this will emphasized that the autologin can be failed asynchronously.

**Comment by SgAkErRu** [comment]
@Vogtinator can you check, please?

File: src/daemon/Display.cpp [hunk 1]
```diff
@@ -365,27 +372,33 @@ namespace SDDM {
...
         }
         if (session.exec().isEmpty()) {
             qCritical() << "Failed to find command for session" << session.fileName();
-            return;
+            return false;
+        }
+        // check if the user even exists
+        struct passwd *pw = getpwnam(qPrintable(user));
+        if (!pw) {
+            qCritical() << "Failed to find the autologin user" << user;
```
Vogtinator:
```suggestion
            qCritical() << "Failed to find the user" << user;
```

This path isn't autologin specific

Vogtinator:
Except for that nitpick it looks good

src/daemon/Display.cpp
```diff
@@ -365,27 +372,33 @@ namespace SDDM {
...
+            return false;
         }
         if (session.exec().isEmpty()) {
             qCritical() << "Failed to find command for session" << session.fileName();
-            return;
+            return false;
+        }
+        // check if the user even exists
+        struct passwd *pw = getpwnam(qPrintable(user));
+        if (!pw) {
```
Vogtinator: I think with this check the cause of the failure is never displayed to the user. Can you try to log in manually with an invalid username? It'll probably just fail quickly without any indication of the error

src/daemon/Display.cpp
```diff
@@ -499,6 +514,11 @@ namespace SDDM {
     void Display::slotAuthError(const QString &message, Auth::Error error) {
         qWarning() << "Authentication error:" << error << message;
 
+        if (m_auth->autologin()) {
+            handleAutologinFailure();
+            return;
```
Vogtinator, line 513: What happens if you add `m_socketServer->informationMessage(m_socket, message);` after the `handleAutologinFailure()`? My hope is that the greeter will react to it like a manual login failed, giving some indication to the user

src/daemon/Display.cpp
SgAkErRu: Hmm.. that's right. Signal 'loginFailed' will not be emitted, so sddm-theme just stuck (disabled UI, can't press buttons), waiting signal with result of login. 

I also found out that it happened with other sanity checks, like "Failed to find command for session", even without this proposed changes. So this is a different problem. I think we need handle that in `Display::Login` and emit `loginFailed` if auth was not started ([here](https://github.com/sddm/sddm/blob/d090b034d68acc882fb9f897e95b0e297a416b41/src/daemon/Display.cpp#L331C44-L331C44)).

Also, I think maybe we should move user existence check to `Display::attemptAutologin` for security reasons, because for manual login case, user existence (is login correct) can be determined by speed of login failing. But I think that check will be alright for autologin case.

src/daemon/Display.cpp
SgAkErRu: That's crashed. Because `m_socket == nullptr`. If indication of failed login in autologin case is important, we can create bool variable `m_autologinFailed` in `Display` and connect to `SocketServer::connected` signal (that signal also have to pass `socket` pointer, like `login` signal do). And it will be like:
```
void Display::slotSocketServerConnected(QLocalSocket *socket) {
        m_socket = socket;

        if (m_autologinFailed && m_socket)
            emit loginFailed(m_socket);
}
```
I locally tried that and it's work, but I'm not sure that is 100% proper solution. I think maybe passing `socket` in `connected` signal will be even more relevant than passing this through `login` signal, but I don't know all the subtleties of this.

src/daemon/Display.cpp
Vogtinator: User existence check is IMO not a sanity check. This check here can just be removed now, as all autologin auth failures are handled now.

src/daemon/Display.cpp
Vogtinator: Ok. In that case let's just leave it like it is. If autologin is broken the user will notice in any case and have a look at the logs.

src/daemon/Display.cpp
SgAkErRu: Ok, I will remove user existence check, but what's about handling other sanity checks? If we won't handle that, in case of manual login, the UI of sddm theme will hang and no error message will be displayed (because it waits result of authorization).

We can emit `loginFailed` signal and theme will display error indication, also we can try to say more precise reason for failure by using `m_socketServer->informationMessage(m_socket, message);` (and I think it is more suitable) with message like `System error. Authorization is not possible` or something like that. But I found out that not all sddm-themes support that `informationMessage` signal, `maya` definitely do, but `plasma breeze` handles only `loginFailed` and `loginSucceeded`.

But anyway, with `loginFailed` or `informationMessage`, I think we need to handle this, because otherwise there is no error indication (that caused by 4 sanity checks) at all in any of the sddm themes.

Vogtinator: > But anyway, with loginFailed or informationMessage, I think we need to handle this, because otherwise there is no error indication (that caused by 4 sanity checks) at all in any of the sddm themes.

Agreed. Sanity checks aren't really supposed to fail though so fixing this isn't high priority IMO. Should be done, but doesn't necessarily have to be in this PR.

SgAkErRu: > Agreed. Sanity checks aren't really supposed to fail though so fixing this isn't high priority IMO. Should be done, but doesn't necessarily have to be in this PR.

Sorry, I didn't understand you completely about sanity checks. How it should be done, like I described or in some other way?

Vogtinator: Yes, like you described. At least `loginFailed` is mandatory.

SgAkErRu: User existence check is removed

Vogtinator (approved):

Issue #1768, opened 2023-07-21
No Wayland desktop reached when user shell is not "bash"


RafaelLinux:
Yesterday, I solved for another Linux (specifically openSUSE) user a problem that was also a headache for me months ago. Coincidentally, I found the solution to my problem thanks to another user who had the same problem as me before. This problem was that when trying to use Wayland for the first time, he was frustrated because after the authentication screen, instead of the Plasma desktop, a black screen appeared. The system wasn't crashing, it was just impossible to do anything. This user commented that after uninstalling the "fish" shell, the problem was solved, and as I also had it installed, I did the same and, voila, the Plasma desktop appeared using Wayland.

After some research, what I found out is that all users (of any distribution) using a shell other than "Bash", have the same problem when trying to use a session with the Wayland protocol.

At this point, and after comments from other developers, they suggest that the problem is SSDM, so I open this thread hoping that you shed light on this problem and that, if it is a bug in SSDM, it will be solved as soon as possible, since the affected users are wasting time looking for solutions in Wayland forums, KDE forums and the different shells. 

I look forward to hearing from you.

Thanks

Vogtinator:
See https://github.com/sddm/sddm/issues/1059.

Is this about using `fish` as login shell or as `/bin/sh`?

RafaelLinux:
I installed "fish" from [repositories](https://download.opensuse.org/repositories/shells:fish:release:3/openSUSE_Tumbleweed/shells:fish:release:3.repo) and then changed with `chsh -s /bin/fish` as other `fish-shell` users affected by this issue.

Vogtinator:
I finally had a look and reproduced the issue locally. FWICT, the implementation of the `xsession` and `wayland-session` scripts is just wrong. Instead of actually invoking a login shell, it sources the profile files itself as `/bin/sh`, then tries to get the environment of the configured login shell, imports that into its `/bin/sh` environment and then execs the session process. 

In this particular case, re-importing the fish environment into bash fails because it tries to set the `$PROFILEREAD` variable, which is readonly.

Instead of re-importing the environment, the session process should be executed by the login shell itself.

Vogtinator:
FTR, the reason that the X11 session works is that openSUSE has a downstream script for X11 which does it correctly (`/usr/etc/X11/xdm/Xsession`)

RafaelLinux:
[...]

PR #1779, opened 2023-08-04
Redesign login shell use in session scripts

Vogtinator:
See https://github.com/sddm/sddm/issues/1768#issuecomment-1666211742

New approach, which does not suffer from the xprofile behaviour change:

Avoid the re-import of variables into sh by using the user's login shell
to invoke the session.

For the wayland-session script it's straightforward: Just exec the session
command directly within the login shell.

For the Xsession script it's a bit more complex: The login shell environment
has to be loaded first, then the xprofile files, so that they can overwrite
the login shell environment. To achieve this, the script runs in two stages:
The first uses the login shell to exec the second stage, which then reads
xprofile and execs the session.

~Draft because only tested outside of a system installation so far.~

### DIFF
--- a/data/scripts/Xsession
+++ b/data/scripts/Xsession
@@ -5,39 +5,34 @@
 # This file is extracted from kde-workspace (kdm/kfrontend/genkdmconf.c)
 # Copyright (C) 2001-2005 Oswald Buddenhagen <ossi@kde.org>
 
-# Note that the respective logout scripts are not sourced.
-case $SHELL in
-  */bash)
-    [ -z "$BASH" ] && exec $SHELL --login $0 "$@"
-    shopt -q login_shell || exec $SHELL --login $0 "$@"
-    set +o posix
-    ;;
-  */zsh)
-    [ -z "$ZSH_NAME" ] && exec $SHELL --login $0 "$@"
-    [[ -o login ]] || exec $SHELL --login $0 "$@"
-    emulate -R sh
-    ;;
-  */csh|*/tcsh)
-    # [t]cshrc is always sourced automatically.
-    # Note that sourcing csh.login after .cshrc is non-standard.
-    xsess_tmp=`mktemp /tmp/xsess-env-XXXXXX`
-    $SHELL -c "if (-f /etc/csh.login) source /etc/csh.login; if (-f ~/.login) source ~/.login; /bin/sh -c 'export -p' >! $xsess_tmp"
-    . $xsess_tmp
-    rm -f $xsess_tmp
-    ;;
-  */fish)
-    [ -f /etc/profile ] && . /etc/profile
-    [ -f $HOME/.profile ] && . $HOME/.profile
-    xsess_tmp=`mktemp /tmp/xsess-env-XXXXXX`
-    $SHELL --login -c "/bin/sh -c 'export -p' > $xsess_tmp"
-    . $xsess_tmp
-    rm -f $xsess_tmp
-    ;;
-  *) # Plain sh, ksh, and anything we do not know.
-    [ -f /etc/profile ] && . /etc/profile
-    [ -f $HOME/.profile ] && . $HOME/.profile
-    ;;
-esac
+# Import the login environment for some shells by starting them as login shell
+# and executing this script. Inside, SDDM_XSESSION_PROFILE_READ will be set and
+# it continues below.
+if [ -z "$SDDM_XSESSION_PROFILE_READ" ]; then
+  export SDDM_XSESSION_PROFILE_READ=1
+
+  case $SHELL in
+    */bash|*/zsh)
+      exec $SHELL --login -c 'exec "$@"' - $0 "$@"
+      ;;
+    */csh|*/tcsh)
+      exec $SHELL -c 'if (-f /etc/csh.login) source /etc/csh.login; if (-f ~/.login) source ~/.login; exec $argv' $0 "$@"
+      ;;
+    */fish)
+      [ -f /etc/profile ] && . /etc/profile
+      [ -f $HOME/.profile ] && . $HOME/.profile
+      exec $SHELL --login -c 'exec $argv' $0 "$@"
+      ;;
+    *) # Plain sh, ksh, and anything we do not know.
+      [ -f /etc/profile ] && . /etc/profile
+      [ -f $HOME/.profile ] && . $HOME/.profile
+      exec $0 "$@"
+      ;;
+  esac
+
+  exit 1
+fi
+unset SDDM_XSESSION_PROFILE_READ
 
 [ -f /etc/xprofile ] && . /etc/xprofile
 [ -f /usr/local/etc/xprofile ] && . /usr/local/etc/xprofile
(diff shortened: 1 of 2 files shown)


Vogtinator:
> Draft because it's late here and so I didn't test it, but wanted to get it published anyway to show the idea.

Works as expected. Tested with bash, fish, csh and zsh.

bash: `/etc/profile` is sourced.

fish: `/etc/profile` is sourced, `fish.config` is also read.

csh: `.cshrc` is read, but not `/etc/profile` (same as before)

zsh: `/etc/profile` is sourced, `.zprofile` is also read.

arrowd:
If I understand it right, the environment from `/etc/profile` now overrides the environment set in `~/.xsession`? Would that be a POLA violation?

Vogtinator:
> If I understand it right, the environment from `/etc/profile` now overrides the environment set in `~/.xsession`? Would that be a POLA violation?

I think so.

Might be possible to just source it unconditionally at the top and assume it won't be sourced again by sh compatible shells through some mechanism like `$PROFILEREAD`.

arrowd:
We don't seem to have `PROFILEREAD` anywhere on FreeBSD. Is this some sort of well-known env var?

Vogtinator:
> We don't seem to have `PROFILEREAD` anywhere on FreeBSD. Is this some sort of well-known env var?

I don't know. The `/etc/profile` I have here does

```
#
# Avoid overwriting user settings if called twice
#
if test -z "$PROFILEREAD" ; then
    readonly PROFILEREAD=true
    export PROFILEREAD
fi
```

and surrounds most code with `test -z "$PROFILEREAD`

arrowd:
> Might be possible to just source it unconditionally at the top and assume it won't be sourced again by sh compatible shells 

This won't work on FreeBSD, unfortunately.

I'm confused, why do we try to pull env from bash/fish/zsh at all if now we call `$SHELL $@` instead of `exec $@`?

CoelacanthusHex:
> > Draft because it's late here and so I didn't test it, but wanted to get it published anyway to show the idea.
> 
> Works as expected. Tested with bash, fish, csh and zsh.
> 
> bash: `/etc/profile` is sourced.
> 
> fish: `/etc/profile` is sourced, `fish.config` is also read.
> 
> csh: `.cshrc` is read, but not `/etc/profile` (same as before)
> 
> zsh: `/etc/profile` is sourced, `.zprofile` is also read.

As [zsh doc](https://zsh.sourceforge.io/Doc/Release/Files.html#Startup_002fShutdown-Files) said, if zsh is invoked with 'zsh', it will not source `/etc/profile` but `/etc/zprofile`. Some distro land a `/etc/zprofile` with `emulate sh -c 'source /etc/profile'`.
If zsh is invoked with 'sh', [it will has same behavior as 'sh'](https://zsh.sourceforge.io/Doc/Release/Invocation.html#Compatibility).

Vogtinator:
> As [zsh doc](https://zsh.sourceforge.io/Doc/Release/Files.html#Startup_002fShutdown-Files) said, if zsh is invoked with 'zsh', it will not source `/etc/profile` but `/etc/zprofile`. Some distro land a `/etc/zprofile` with `emulate sh -c 'source /etc/profile'`. If zsh is invoked with 'sh', [it will has same behavior as 'sh'](https://zsh.sourceforge.io/Doc/Release/Invocation.html#Compatibility).

That means, if we source `/etc/profile` explicitly and the distro also sources it from `/etc/zprofile`, it ends up getting sourced twice (which may or may not be an issue). If we don't (like now), `/etc/profile` might not get sourced at all.

HerrNaN:
Tried to incorporate this in my Fedora Sericea build but to no avail, then I saw that the Fedora tests are not passing. Is there any work planned to get this working and into the main branch?

Vogtinator:
> Tried to incorporate this in my Fedora Sericea build but to no avail, then I saw that the Fedora tests are not passing. Is there any work planned to get this working and into the main branch?

The Fedora CI failures are for Fedora reasons, not SDDM reasons.

HerrNaN:
> > Tried to incorporate this in my Fedora Sericea build but to no avail, then I saw that the Fedora tests are not passing. Is there any work planned to get this working and into the main branch?
> 
> The Fedora CI failures are for Fedora reasons, not SDDM reasons.

I see, I haven't been able to pin-point exactly what those are, do you know?

Vogtinator:
> > > Tried to incorporate this in my Fedora Sericea build but to no avail, then I saw that the Fedora tests are not passing. Is there any work planned to get this working and into the main branch?
> > 
> > 
> > The Fedora CI failures are for Fedora reasons, not SDDM reasons.
> 
> I see, I haven't been able to pin-point exactly what those are, do you know?

On clang internal compiler errors, i.e. compiler bugs. The current failures on develop are new, the CI workflow needs to be adjusted after fedora switched sddm to Qt 6 (I assume).

Vogtinator:
I implemented a different approach which should work everywhere without noticable behaviour changes. Needs testing.

arrowd:
It looks good to me, but I didn't test it at the runtime.

Vogtinator:
Tested with fish and bash, works.

Pogogo007:
Any activity? I had a bug that caused sddm to fail to load any session while using fish and this mr completly fixed the issue

Vogtinator:
It's ready for merge, just needs approval from a maintainer.

bernhardreiter:
> It's ready for merge, just needs approval from a maintainer.

Meanwhile you could sync it with the base branch.

Thanks for having put work in towards a solution!

jfroy:
For what it's worth, I was really puzzled why sddm failed to create sessions on my system until I found and tested this PR. It does indeed fix the problem if your shell is fish.

LDprg:
I had problems with sddm wayland on nixos with fish as shell. This fixes all my problems. Any updates of the merge of this great patch? @Vogtinator maybe someone could be pinged for reviewing, since this PR has been around for a while.

Vogtinator:
Sure, I can ping again...

@aleixpol @davidedmundson

SalahAdDin:
We are waiting for yet!

Vogtinator:
It's been another month, so:

@aleixpol @davidedmundson

mcwillzz:
Still waiting for this to come through!

BeiyanYunyi:
Still waiting. For Arch Linux users, I wrote [a PKGBUILD](https://github.com/BeiyanYunyi/sddm-1779-PKGBUILD) to patch sddm with this PR, using it to replace the sddm from Arch Linux repository makes my sddm work with fish.

SalahAdDin:
> Still waiting. For Arch Linux users, I wrote [a PKGBUILD](https://github.com/BeiyanYunyi/sddm-1779-PKGBUILD) to patch sddm with this PR, using it to replace the sddm from Arch Linux repository makes my sddm work with fish.

where can we up the PR?

ccryx:
For the pacman users there's also the option of using a pacman hook.
Place this in `/etc/pacman.d/hooks/somefile.hook`. The file `sddm_fish_wayland-session` contains the patched script from this PR.
```
[Trigger]
Operation = Install
Operation = Upgrade
Type = Package
Target = sddm

[Action]
Description = Hook: Use fixed wayland-session script for fish as the default shell
When = PostTransaction
Exec = /usr/bin/cp --backup -S ".original" /etc/pacman.d/hooks/sddm_fish_wayland-session /usr/share/sddm/scripts/wayland-session
```

davidedmundson (approved):

garrett-edwards:
Just came across this issue and was sad to see no merge in months, then I see @davidedmundson approve today! Wonderful!

PR #1789, opened 2023-09-09
Provide coinstallability for Qt 5 and Qt 6 themes

Vogtinator:
sddm-greeter is already a separate process, so it's actually rather simple to support themes using Qt 5 and themes using Qt 6 on the same system. This PR achieves this by installing the greeter specific files with a -qt6 suffix when `BUILD_FOR_QT6=ON` and dynamically choosing which greeter to execute based on the `QtVersion=` property in the theme `metadata.desktop`. For backwards compatibility, this defaults to 5 if not present.

This means that all current Qt 6 themes (i.e. just breeze?) will have to add `QtVersion=6`, otherwise the Qt 5 greeter will be used.

Open topics:
- [ ] Are translations in .qm format compatible across Qt major versions? I couldn't find any info for this. If they are compatible, then the -qt suffix for the translation dir can be avoided. Themes can also provide .qm files.
- [ ] Should the `QtVersion` property allow to specify multiple versions as fallback? I'm not sure whether it's practical to have a theme which supports multiple major versions, especially if .qm files are not necessarily compatible.
- [ ] What should the example themes specify as Qt version? elarun currently has a broken background with Qt 6.
- [ ] For after the merge: Document the `QtVersion` property and behaviour

With https://github.com/sddm/sddm/pull/1790 on top, the Qt 6 build will automatically build and install the Qt 5 greeter as well.

### DIFF
--- a/src/common/ThemeMetadata.cpp
+++ b/src/common/ThemeMetadata.cpp
@@ -28,6 +28,7 @@ namespace SDDM {
         QString mainScript { QStringLiteral("Main.qml") };
         QString configFile;
         QString translationsDirectory { QStringLiteral(".") };
+        int qtVersion = 5;
     };
 
     ThemeMetadata::ThemeMetadata(const QString &path, QObject *parent) : QObject(parent), d(new ThemeMetadataPrivate()) {
@@ -50,11 +51,16 @@ namespace SDDM {
         return d->translationsDirectory;
     }
 
+    int ThemeMetadata::qtVersion() const {
+        return d->qtVersion;
+    }
+
     void ThemeMetadata::setTo(const QString &path) {
         QSettings settings(path, QSettings::IniFormat);
         // read values
         d->mainScript = settings.value(QStringLiteral("SddmGreeterTheme/MainScript"), QStringLiteral("Main.qml")).toString();
         d->configFile = settings.value(QStringLiteral("SddmGreeterTheme/ConfigFile"), QStringLiteral("theme.conf")).toString();
         d->translationsDirectory = settings.value(QStringLiteral("SddmGreeterTheme/TranslationsDirectory"), QStringLiteral(".")).toString();
+        d->qtVersion = settings.value(QStringLiteral("SddmGreeterTheme/QtVersion"), 5).toInt();
     }
 }
--- a/src/greeter/CMakeLists.txt
+++ b/src/greeter/CMakeLists.txt
@@ -1,3 +1,14 @@
+if(QT_MAJOR_VERSION EQUAL "5")
+    # Keep the unversioned name for Qt5. When upgrading SDDM, the old daemon
+    # might still be running and only know about "sddm-greeter". Keeping the
+    # previous name around also helps users calling it directly.
+    set(GREETER_TARGET sddm-greeter)
+else()
+    set(GREETER_TARGET sddm-greeter-qt${QT_MAJOR_VERSION})
+endif()
+
+message(STATUS "Building greeter for Qt ${QT_MAJOR_VERSION} as ${GREETER_TARGET}")
+
 include_directories(
     "${CMAKE_SOURCE_DIR}/src/common"
     "${CMAKE_BINARY_DIR}/src/common"
@@ -28,18 +39,17 @@ configure_file("theme.qrc" "theme.qrc")
 
 qt_add_resources(RESOURCES ${CMAKE_CURRENT_BINARY_DIR}/theme.qrc)
 
-add_executable(sddm-greeter ${GREETER_SOURCES} ${RESOURCES})
-target_link_libraries(sddm-greeter
+add_executable(${GREETER_TARGET} ${GREETER_SOURCES} ${RESOURCES})
+target_link_libraries(${GREETER_TARGET}
                       Qt${QT_MAJOR_VERSION}::Quick
                       ${LIBXCB_LIBRARIES}
                       ${LIBXKB_LIBRARIES})
 
 if(JOURNALD_FOUND)
-    target_link_libraries(sddm-greeter ${JOURNALD_LIBRARIES})
+    target_link_libraries(${GREETER_TARGET} ${JOURNALD_LIBRARIES})
 endif()
 
 # Translations
-add_dependencies(sddm-greeter components-translation)
-add_dependencies(sddm-greeter themes-translation)
+add_dependencies(${GREETER_TARGET} components-translation themes-translation)
 
-install(TARGETS sddm-greeter DESTINATION "${CMAKE_INSTALL_BINDIR}")
+install(TARGETS ${GREETER_TARGET} DESTINATION "${CMAKE_INSTALL_BINDIR}")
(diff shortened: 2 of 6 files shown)


src/greeter/CMakeLists.txt
```diff
@@ -1,3 +1,14 @@
+if(QT_MAJOR_VERSION EQUAL "5")
+    # Keep the unversioned name for Qt5. When upgrading SDDM, the old daemon
+    # might still be running and only know about "sddm-greeter". Keeping the
+    # previous name around also helps users calling it directly.
+    set(GREETER_TARGET sddm-greeter)
```
Conan-Kudo, line 5: Why not just handle the unversioned name via symlink for this purpose? I'd rather us just change to versioned names unless we intend to quickly drop Qt5 support.

Conan-Kudo (changes requested):

src/greeter/CMakeLists.txt
Vogtinator: What's the benefit of using a symlink instead?

I decided against it because it's one more file and installing a symlink with cmake needs a custom command...

Conan-Kudo: It does? `file(CREATE_LINK)` has been available since CMake 3.14...

Vogtinator: Would need to check whether bumping the requirement would be ok.

Conan-Kudo: `set(CMAKE_CXX_STANDARD 17)` was added in CMake 3.20, so you already require that as a minimum.

Vogtinator: That was most likely an oversight, but nobody complained so it's probably ok :shrug:

Vogtinator: > set(CMAKE_CXX_STANDARD 17) was added in CMake 3.20, so you already require that as a minimum.

That was actually in CMake 3.8: https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD

Conan-Kudo: Oh welp, right, C++20 was in CMake 3.20. My bad.

evelikov:
Does the daemon need a similar -qt6 handling or is it expected that mixing qt5 daemon and qt6 greeter (or vice-versa) will still work?

Vogtinator:
> Does the daemon need a similar -qt6 handling or is it expected that mixing qt5 daemon and qt6 greeter (or vice-versa) will still work?

Both combinations are valid. https://github.com/sddm/sddm/pull/1790 makes that even more explicit.

Vogtinator:
@aleixpol @davidedmundson Could you provide input on the open topics?

Vogtinator:
> @aleixpol @davidedmundson Could you provide input on the open topics?

Ping! This needs to be done before anyone develops themes without QtVersion=6 which will then inadvertedly break after this is merged.

davidedmundson:
> Are translations in .qm format compatible across Qt major versions? I couldn't find any info for this. If they are compatible, then the -qt suffix for the translation dir can be avoided. Themes can also provide .qm files.

Diff implies that pragmatically they are.
The DTD looks the same https://doc.qt.io/qt-6/linguist-ts-file-format.html  and that can't change on a whim.

>Should the QtVersion property allow to specify multiple versions as fallback? I'm not sure whether it's practical to have a theme which supports multiple major versions, especially if .qm files are not necessarily compatible.

I would say no, the point of this is to have SDDM do the work so theme devs don't have to. QML is not ifdef-able and trying to support multiple versions has not been easy for other places. 

> What should the example themes specify as Qt version? elarun currently has a broken background with Qt 6.

The default theme (elarun) *has* to work in both. That gets baked in via qrc into both executables.
On hitting a QML error we fallback to the default theme in process without restarting the greeter. 

We also have the case at a cmake level of not building a Qt6 greeter at which point we want these to work. I would suggest leave them alone as-is for now then bump when we make Qt6 mandatory.

I could not reproduce the Image issue.

>For after the merge: Document the QtVersion property and behaviour

Definitely, especially the part for packagers

Vogtinator:
> > Are translations in .qm format compatible across Qt major versions? I couldn't find any info for this. If they are compatible, then the -qt suffix for the translation dir can be avoided. Themes can also provide .qm files.
> 
> Diff implies that pragmatically they are. The DTD looks the same https://doc.qt.io/qt-6/linguist-ts-file-format.html and that can't change on a whim.

That's for TS files. QM files are basically serialized Qt hashes and I have no idea whether that's meant to be stable... Arguably if they're compatible now, then using the Qt 5 files for Qt 6 is fine, but not necessarily the other way around.

> > Should the QtVersion property allow to specify multiple versions as fallback? I'm not sure whether it's practical to have a theme which supports multiple major versions, especially if .qm files are not necessarily compatible.
> 
> I would say no, the point of this is to have SDDM do the work so theme devs don't have to. QML is not ifdef-able and trying to support multiple versions has not been easy for other places.

Yep...

> > What should the example themes specify as Qt version? elarun currently has a broken background with Qt 6.
> 
> The default theme (elarun) _has_ to work in both. That gets baked in via qrc into both executables. On error we fallback to the default theme on QML error in process without restarting the greeter.
> 
> We also have the case at a cmake level of not building a Qt6 greeter at which point we want these to work. I would suggest leave them alone as-is for now then bump when we make Qt6 mandatory.
> 
> I could not reproduce the Image issue.

Ok.

> > For after the merge: Document the QtVersion property and behaviour
> 
> Definitely, especially the part for packagers

Can do. Currently the wiki is the best documentation source or should I put that into README?

Vogtinator:
Ping

Vogtinator:
Ping.

evelikov:
I patched the sddm package on my Arch box and installed it. Works like a charm and the code is simple enough.

Feel free to consider this "Tested-by" and "Reviewed-by" fwiw o/

davidedmundson (approved):

src/common/ThemeMetadata.cpp
```diff
@@ -50,11 +51,16 @@ namespace SDDM {
...
+        return d->qtVersion;
+    }
+
     void ThemeMetadata::setTo(const QString &path) {
         QSettings settings(path, QSettings::IniFormat);
         // read values
         d->mainScript = settings.value(QStringLiteral("SddmGreeterTheme/MainScript"), QStringLiteral("Main.qml")).toString();
         d->configFile = settings.value(QStringLiteral("SddmGreeterTheme/ConfigFile"), QStringLiteral("theme.conf")).toString();
         d->translationsDirectory = settings.value(QStringLiteral("SddmGreeterTheme/TranslationsDirectory"), QStringLiteral(".")).toString();
+        d->qtVersion = settings.value(QStringLiteral("SddmGreeterTheme/QtVersion"), 5).toInt();
```
evelikov, line 64: Shouldn't we set `QtVersion` for the themes we ship? Casual grep suggests that these would need an update:

```
data/themes/elarun/metadata.desktop
data/themes/maldives/metadata.desktop
data/themes/maya/metadata.desktop
src/greeter/theme/metadata.desktop
```

While I haven't tried the Qt6 side, just yet. I've build this with Qt5 and it works like a charm.

With the above .desktop/QtVersion handled, I think we're good to merge.

Vogtinator: > Shouldn't we set QtVersion for the themes we ship? Casual grep suggests that these would need an update:

Kind of. The builtin themes are only semi-compatible with Qt 6 :-/

That's also mentioned in the PR message:

> - [ ] What should the example themes specify as Qt version? elarun currently has a broken background with Qt 6.

PR #1805, opened 2023-10-02
Prevent tmpfiles from removing "old" xauth files

Rob--W:
Starting from sddm 0.20 (by https://github.com/sddm/sddm/pull/1230), the xauth file is at `/tmp/xauth_*` instead of `~/.Xauthority`. The new location is subject to systemd-tmpfiles's cleanup, which deletes files from /tmp/ that are older than 10 days: https://github.com/systemd/systemd/blob/v254/tmpfiles.d/tmp.conf

The xauth file should not be removed while the session is active, because it is needed for starting X applications. When removed, X applications won't start any more.

This patch fixes the issue by declaring these files as "ignored" in sddm-tmpfiles.conf, so that systemd-tmpfiles doesn't remove them.

Minimal test case:

```sh
touch /tmp/xauth_testonly
SYSTEMD_LOG_LEVEL=debug systemd-tmpfiles --prefix=/tmp/ --clean
```

Pre-patch output:

> File "/tmp/xauth_testonly": change time Mon 2023-10-02 01:27:38.395466 CEST is too new.

(meaning that the file was considered for removal)

Post-patch output:

> Ignoring "/tmp/xauth_testonly": a separate glob exists.

(meaning that the file is ignored and not considered for removal)

### DIFF
--- a/services/sddm-tmpfiles.conf.in
+++ b/services/sddm-tmpfiles.conf.in
@@ -6,3 +6,9 @@ d	${RUNTIME_DIR}	0711	root	root
 r!	/tmp/sddm-auth*
 # xauth files passed to user sessions
 r!	/tmp/xauth_*
+# "r!" above means to remove the files if existent (r), but only at boot (!).
+# tmpfiles.d/tmp.conf declares a periodic cleanup of old /tmp/ files, which
+# would ordinarily result in the deletion of our xauth files. To prevent that
+# from happening, explicitly tag these as X (ignore).
+X	/tmp/sddm-auth*
+X	/tmp/xauth_*


Vogtinator:
Wow, so apparently `r!` does not override any statement for `/tmp` earlier? IMO that's a systemd bug. Can you start a discussion there or link to an existing one?

Rob--W:
> Wow, so apparently `r!` does not override any statement for `/tmp` earlier? IMO that's a systemd bug. Can you start a discussion there or link to an existing one?

While I was initially surprised about the behavior, I don't think that it is a systemd bug. The purpose of "r" is not to prevent removal, but to schedule removal.

"r" means: remove immediately (independently of age) when the task is run. To limit this effect to boot, "!" can be appended. There is a feature request to support age with "r" at https://github.com/systemd/systemd/issues/1633 (but this does not help at all with the issue here)

There are other types to schedule removal dependent on age, such as "q". This is defined for /tmp by something else, and is exactly why we need a way to cancel the removal task (as explained in the PR).

"x" is explicitly documented to not affect "r". It can cancel other age-based cleanup tasks.

For the full overview of the syntax, see https://www.man7.org/linux/man-pages/man5/tmpfiles.d.5.html

evelikov:
The `!` is a type modifier, where the documentation states:

`If the exclamation mark ("!") is used, this line is only safe to execute during boot, and can break a running system.`

As-is I'm not sure the change makes sense. What is the output of `systemd-analyze cat-config tmpfiles.d` on the affected system?

Vogtinator:
AFAICT both the behaviour expected by the current sddm tmpfiles config (`r!` overrides entries for parent directories) as well as the behaviour implemented by systemd-tmpfiles (it doesn't) are valid interpretations of the documentation.

The main issue is that systemd-tmpfiles is what we have less influence on and this issue needs to be addressed in some way, so I think we'll have to take this at least as a workaround, and ideally get this fixed (or at least clarified) by systemd upstream.

My argument against systemd's current behaviour is that currently `r!` just breaks because some other config file out of our control and that we might not even be aware of, has an entry for a parent directory. In this particular case it's not that surprising as `/tmp` is well known, but there can be more complex ones.

Rob--W:
> The `!` is a type modifier, where the documentation states:
> 
> `If the exclamation mark ("!") is used, this line is only safe to execute during boot, and can break a running system.`

Side note: this last part can be interpreted (wrongly) as "Line with ! can break a running system". This is incorrect, and an unambiguous version could be "An exclamation mark ("!") can be used to restrict the effect to boot only, in case the operation can break a running system when run after boot."

> As-is I'm not sure the change makes sense.

I don't see why the change of the PR would not make sense. I explained in the PR and commit message why the change is needed, and what the effect of not having the patch is.

>What is the output of `systemd-analyze cat-config tmpfiles.d` on the affected system?

It includes sddm's config and systemd's tmp.conf, both of which I have linked from the PR and commit message.

> AFAICT both the behaviour expected by the current sddm tmpfiles config (`r!` overrides entries for parent directories) as well as the behaviour implemented by systemd-tmpfiles (it doesn't) are valid interpretations of the documentation.
> 
> The main issue is that systemd-tmpfiles is what we have less influence on and this issue needs to be addressed in some way, so I think we'll have to take this at least as a workaround, and ideally get this fixed (or at least clarified) by systemd upstream.

Again, I don't think that this is necessarily a bug in systemd, as I elaborated at https://github.com/sddm/sddm/pull/1805#issuecomment-1744257471 . Changing the behavior of "r!" could result in broken behavior elsewhere, if that logic relies on the current behavior.

Regardless of the thoughts on "desired" systemd behavior, it should be obvious that the current sddm setup is broken and resulting in observably broken behavior. For that reason, I believe that the PR should be merged without being blocked on the philosophical discussion on the desired vs documented vs implemented behavior (or documentation work in systemd). Even if systemd tmpfiles's behavior were to change, it would complement rather than conflict with the PR here.

Vogtinator:
> Regardless of the thoughts on "desired" systemd behavior, it should be obvious that the current sddm setup is broken and resulting in observably broken behavior. For that reason, I believe that the PR should be merged without being blocked on the philosophical discussion on the desired vs documented vs implemented behavior (or documentation work in systemd). Even if systemd tmpfiles's behavior were to change, it would complement rather than conflict with the PR here.

Correct, but this topic needs to be raised with systemd upstream in any case. I'm busy with other stuff currently though so I won't do that myself in the next days at least.

evelikov:
Apologies, seems like I misread my system output. Namely - I could swear that the `r!` entries were preserved.

Rob--W:
Can the patch be approved and merged? It has been one month since the last comment, and the reason for inaction seems lack of time PLUS your perceived issue in systemd. As I argued in https://github.com/sddm/sddm/pull/1805#issuecomment-1744257471, this is NOT a systemd-tmpfiles issue.

[Your expectation](https://github.com/sddm/sddm/pull/1805#issuecomment-1742481275) of `r!` overriding the `/tmp` cleanup behavior is NOT supported by the documentation ([man tmpfiles.d](https://www.man7.org/linux/man-pages/man5/tmpfiles.d.5.html)):

1. `r` is documented to remove the file
2. `!` means to limit the action to at boot time.
3. `r!` thus defines a delete action at boot time.
4. The above does NOT imply that the file stays around forever at non-boot time. As this issue/PR shows, a different line can trigger the removal. And that different line can be undone by `X`, which is explicitly documented to ignore a path during cleaning.

Relevant excerpts from the documentation (removed lines that aren't relevant):

```
#Type Path                                     Mode User Group Age         Argument
q     /subvolume-or-directory/to/create        mode user group cleanup-age -
r     /path-or-glob/to/remove                  -    -    -     -           -
X     /path-or-glob/to/ignore                  -    -    -     cleanup-age -
```

For comparison, the [tmp.conf](https://github.com/systemd/systemd/blob/v254/tmpfiles.d/tmp.conf), original config and this PR combined looks like this:

```
q     /tmp                                     1777 root root 10d  # from systemd tmp.conf
r!    /tmp/xauth_*                                                 # original sddm
X     /tmp/xauth_*                                                 # added to sddm by this PR
```

The types (letters at the start of each line) are documented as follows:
```
       d
           Create a directory. The mode and ownership will be adjusted
           if specified. Contents of this directory are subject to
           time-based cleanup if the age argument is specified.
       q
           (...) On non-btrfs file systems, this line type is identical to d.
       X
           Ignore a path during cleaning. Use this type to exclude paths
           from clean-up as controlled with the Age parameter. Unlike x,
           this parameter will not exclude the content if path is a
           directory, but only directory itself. Note that lines of this
           type do not influence the effect of r or R lines. Lines of
           this type accept shell-style globs in place of normal path
           names.
       r
           Remove a file or directory if it exists. This may not be used
           to remove non-empty directories, use R for that. Lines of
           this type accept shell-style globs in place of normal path
           names. Does not follow symlinks.
```

```
Type Modifiers
       If the exclamation mark ("!") is used, this line is only safe to
       execute during boot, and can break a running system. Lines
       without the exclamation mark are presumed to be safe to execute
       at any time, e.g. on package upgrades.
```

Vogtinator:
> Can the patch be approved and merged? It has been one month since the last comment, and the reason for inaction seems lack of time PLUS your perceived issue in systemd. As I argued in https://github.com/sddm/sddm/pull/1805#issuecomment-1744257471, this is NOT a systemd-tmpfiles issue.

It's something that needs to be discussed with upstream.

I have the strong impression that if this gets merged without the discussion having been started, it won't happen in the future either.

Rob--W:
> It's something that needs to be discussed with upstream.
> 
> I have the strong impression that if this gets merged without the discussion having been started, it won't happen in the future either.

I disagree with the need for having an upstream discussion. In my last comment I put all sources and documentation together to show why the behavior understandably makes sense. My findings directly contradict your asserted expectation on the behavior.

If you still believe that there is an issue with the configuration behavior or documentation, then I'd like to see some more substance. For example reasoning supported by documentation on why the xauth file is expected to be kept with sddm's tmpfiles.d config. I cannot start an upstream discussion when I'm not convinced of your premise in the first place.

And again, this patch fixes a bug in sddm. A bug that results in broken sessions and crashing applications. What's more important: fixing this bug or having a discussion? Even if the premise of the discussion is valid, I wouldn't expect a quick resolution: a discussion and a backwards-incompatible rollout need time. If you want to wait for that to settle, then sddm is going to be unstable for a very long time.

Vogtinator:
> > It's something that needs to be discussed with upstream.
> > I have the strong impression that if this gets merged without the discussion having been started, it won't happen in the future either.
> 
> I disagree with the need for having an upstream discussion. In my last comment I put all sources and documentation together to show why the behavior understandably makes sense. My findings directly contradict your asserted expectation on the behavior.

Why not just do that upstream then?

> If you still believe that there is an issue with the configuration behavior or documentation, then I'd like to see some more substance. For example reasoning supported by documentation on why the xauth file is expected to be kept with sddm's tmpfiles.d config. I cannot start an upstream discussion when I'm not convinced of your premise in the first place.

Package A has `!r /some/location/some/where`. Package B at some point adds `q /some/where 1777 root root 10d`. This broke package A, without being package A's fault.

> And again, this patch fixes a bug in sddm. A bug that results in broken sessions and crashing applications. What's more important: fixing this bug or having a discussion?

I want to fix it properly, such that other projects also benefit.

> Even if the premise of the discussion is valid, I wouldn't expect a quick resolution: a discussion and a backwards-incompatible rollout need time. If you want to wait for that to settle, then sddm is going to be unstable for a very long time.

I just want to get the ball rolling. Once there's an issue at systemd/systemd, I'm happy.

Rob--W:
> Package A has `!r /some/location/some/where`. Package B at some point adds `q /some/where 1777 root root 10d`. This broke package A, without being package A's fault.

If A didn't want its file to be deleted, then A should use the option that is documented to fulfill that task: X ("Ignore a path during cleaning").

`r!` specifies when the path should be deleted, NOT when it should be kept.

> I just want to get the ball rolling. Once there's an issue at systemd/systemd, I'm happy.

What is your specific request?

"r! path should not be cleaned up by other tasks"

"Need option to ensure that a file doesn't get deleted"? (The X option already does that)

Vogtinator:
> > I just want to get the ball rolling. Once there's an issue at systemd/systemd, I'm happy.
> 
> What is your specific request?
> 
> "r! path should not be cleaned up by other tasks"

Yes. At least systemd needs to warn about this, as using `r!` while some other option might delete it when the system is up doesn't really make sense.

> "Need option to ensure that a file doesn't get deleted"? (The X option already does that)

Vogtinator:
> > Even if the premise of the discussion is valid, I wouldn't expect a quick resolution: a discussion and a backwards-incompatible rollout need time. If you want to wait for that to settle, then sddm is going to be unstable for a very long time.
> 
> I just want to get the ball rolling. Once there's an issue at systemd/systemd, I'm happy.

Nobody bothered so far, so I did it now: https://github.com/systemd/systemd/issues/30361

With that out of the way, this PR can be merged.

Vogtinator (approved):

Issue #1628, opened 2022-12-14
sddm and wayland: arch solution


slynobody:
what do you think of this solution a) in general and b) security-wise? 
https://wiki.archlinux.org/title/SDDM#Running_under_Wayland

and what would be the best solution to enter after '[Wayland] CompositorCommand='?
i fiddled around a bit and observed a few hickups here and there when using f.e. this:

CompositorCommand=kwin_wayland --no-lockscreen --no-kactivities --drm --lock

sometimes there seems to be a mouse-corner in the upper left and you sometimes could trigger something that reminds me of rudimentaries of a window-changer through alt-tab. to put it positive: could this be a first step to a solution to also integrate interfaces to audio, bluetooth, screen-backlight and even screensavers to sddm (like in gdm)?

aleixpol:
Here it's described what we suggest from Plasma:
https://invent.kde.org/plasma/plasma-workspace/-/tree/master/sddm-wayland-session

Agreed, there's opportunity to do a bit more, like launching powerdevil which shouldn't be too hard.

slynobody:
Thank you for your answer.

What do you think security-wise on the sketched solution? esp. in situations when you have to restart the sddm-service via root-shell because it crashed before (common since mesa 22.3 as well as mesa 23) ?

And  could you please explain in simple words what this will do (does not work here, needs another shell (is layer-shell old?)?
QT_WAYLAND_SHELL_INTEGRATION=layer-shell

what could be benefits of current alternatives:
-    ivi-shell
-    wl-shell
-    xdg-shell
-    fullscreen-shell-v1

in general: are there / what are the plans to 'do a bit more' from here on when it comes to power-devil- / sound- / background-lighting-integration?

aleixpol:
`layer-shell` is not old, it's what we use for fullscreen shells. It's developed by wlroots and we implement it through https://invent.kde.org/plasma/layer-shell-qt.

The alternatives are all fairly old and unused. We considered using xdg-shell too which is what we use for most windows but managing fullscreen windows with xdg-shell has proven a challenge to us, thus the development and adoption of layer-shell-qt.

We should find a way to opt in the services that we think would be acceptable in the session. I've tested powerdevil before and it worked alright.

slynobody:
is there any sddm development in the making / planning anymore? or do you really think the current state feature-wise is 'final'?

PR #1831, opened 2023-11-20
Correctly integrate the greeter with layer-shell

aleasto:
When using Wayland as the display server it is possible and very likely that we are requested to use the layer-shell protocol to display the greeter, as it better fits its windowing model. This is selected with QT_WAYLAND_SHELL_INTEGRATION=layer-shell.

With Plasma 6 LayerShellQt has changed the default keyboard policy to not receive any inputs, now following the protocol spec. That breaks the ability type in a username/password, unless we request it to enable keyboard events again.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,7 +51,7 @@ jobs:
           dnf --assumeyes install dnf-plugins-core
           if [[ "${{ matrix.container-image }}" =~ "centos" ]]; then dnf --assumeyes config-manager --set-enabled crb && dnf --assumeyes install epel-release; fi
           dnf --assumeyes builddep sddm
-          dnf --assumeyes --allowerasing --nobest install clang clazy
+          dnf --assumeyes --allowerasing --nobest install clang clazy layer-shell-qt-devel
       - name: Dependencies (deb-type)
         if: ${{ contains(matrix.container-image, 'ubuntu') }}
         run: |
@@ -62,7 +62,7 @@ jobs:
           if [ "${{ matrix.qt }}" = "qt6" ]; then
             DEBIAN_FRONTEND=noninteractive apt-get install clang clazy qt6-base-dev qt6-base-dev-tools qt6-tools-dev qt6-declarative-dev qml6-module-qtqml-workerscript qml6-module-qtquick-window -y
           else
-            DEBIAN_FRONTEND=noninteractive apt-get install clang clazy qml-module-qttest -y
+            DEBIAN_FRONTEND=noninteractive apt-get install clang clazy liblayershellqtinterface-dev qml-module-qttest -y
           fi
       - name: Build
         run: |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,6 +73,9 @@ find_package(XCB REQUIRED)
 # XKB
 find_package(XKB REQUIRED)
 
+# LayerShell integration
+find_package(LayerShellQt REQUIRED)
+
 # Qt
 if(BUILD_WITH_QT6)
     set(QT_MAJOR_VERSION 6)
--- a/src/greeter/CMakeLists.txt
+++ b/src/greeter/CMakeLists.txt
@@ -31,6 +31,7 @@ qt_add_resources(RESOURCES ${CMAKE_CURRENT_BINARY_DIR}/theme.qrc)
 add_executable(sddm-greeter ${GREETER_SOURCES} ${RESOURCES})
 target_link_libraries(sddm-greeter
                       Qt${QT_MAJOR_VERSION}::Quick
+                      LayerShellQt::Interface
                       ${LIBXCB_LIBRARIES}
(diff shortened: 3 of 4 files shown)


aleasto:
So this is a bit problematic, as the CI is finding out, because `LayerShellQt` only targets one Qt version and it is tied to the KDE Plasma version... Indeed, regardless of this change, if you want to use the wayland greeter with the [Plasma configuration](https://invent.kde.org/plasma/plasma-workspace/-/blob/master/sddm-wayland-session/plasma-wayland.conf#L3) (that is using layer-shell), SDDM must be compiled against the same Qt version as KWin.

I know this setting can be changed via qml instead, but don't know how. @aleixpol able to help?

Vogtinator:
> So this is a bit problematic, as the CI is finding out, because LayerShellQt only targets one Qt version and it is tied to the KDE Plasma version...

There just need to be `LayerShellQt5` and `LayerShellQt6` cmake modules IMO.

> Indeed, regardless of this change, if you want to use the wayland greeter with the [Plasma configuration](https://invent.kde.org/plasma/plasma-workspace/-/blob/master/sddm-wayland-session/plasma-wayland.conf#L3) (that is using layer-shell), SDDM must be compiled against the same Qt version as KWin.

No, why?

Vogtinator:
IMO this should not need linking to LayerShellQt. Either this gets done by default in the layer-shell integration plugin or sddm should just use xdg-shell.

aleasto:
> > Indeed, regardless of this change, if you want to use the wayland greeter with the [Plasma configuration](https://invent.kde.org/plasma/plasma-workspace/-/blob/master/sddm-wayland-session/plasma-wayland.conf#L3) (that is using layer-shell), SDDM must be compiled against the same Qt version as KWin.
> 
> No, why?

Sorry, not really kwin but plasma-workspace, kscreen, kscreenlocker, powerdevil... so Plasma in general.

Because `QT_WAYLAND_SHELL_INTEGRATION=layer-shell` makes the greeter load LayerShellQt at runtime, so that must be compiled against the same Qt version as the greeter. Then you need LayerShellQt to be compiled for the same Qt version as plasma-workspace because that links it.
Since LayerShellQt is not coinstallable for both Qt5 and Qt6, the sddm greeter and plasma-workspace must be compiled against the same version of Qt

Vogtinator:
> Since LayerShellQt is not coinstallable for both Qt5 and Qt6,

The plugin is coinstallable for Qt 5 and Qt 6.

aleasto:
Mmmh the plugin itself, yes it is.
But it links to `libLayerShellQtInterface` as a system library, and that's not coinstallable

Vogtinator:
> But it links to libLayerShellQtInterface as a system library, and that's not coinstallable

IMO it should be. Something for upstream. In any case, this is a problem caused by Plasma (sddm itself does not use layer-shell) and needs to be fixed there.

aleasto:
@zzag can we hook changing the keyboard policy with a generic qt interface? or perhaps fetch the default from an env variable?

zzag:
layer-shell-qt could map Qt::WindowDoesNotAcceptFocus to keyboard interactivity flags. But the problem is that layer shell's keyboard interactivity is not just an on/off switch. On the other hand, it's possible that the exclusive mode is going to be dropped in v2 so we could map Qt's window flag to keyboard interactivity. Twisting layer-shell-qt until sddm can use it is also not great... Do you know why xdg-shell doesn't suffice? 

edit: setScreen() + showFullsSreen() should work, which sddm seems to do already

aleasto:
Yes making the views fullscreen was necessary for xdg-shell multi-screen setups (https://github.com/sddm/sddm/pull/1673)
@aleixpol should know if there were more reasons to go for layer-shell

Vogtinator:
https://github.com/sddm/sddm/issues/1628#issuecomment-1361765043 is the current rationale for layer-shell. IMO this should be reconsidered, xdg-shell is what is used for applications everywhere anyway and should work fine.

aleasto:
So one issue with xdg-shell is that kwin allows you to Meta+drag the greeter window from one display to the other

zzag:
Okay... It might be more annoying to prevent kwin from doing that. Let's then follow up on syncing DoesNotAcceptFocus with keyboard interactivity flags.

zzag:
Alternatively making OnDemand the default flag again, i.e. reverting https://invent.kde.org/plasma/layer-shell-qt/-/commit/be6378388894f2fa6bd01a2711a5b0c2053809ac

aleasto:
I think reverting the default is reasonable.
It may not match the upstream protocol, but it better fits the model of a "pluggable" Qt shell backend

aleasto:
Later, if/when `ext-layer-shell` is standardized we can integrate it with `Qt::WindowDoesNotAcceptFocus`

aleasto:
The default was flipped to this shouldn't be necessary anymore.
Thanks everyone

PR #1833, opened 2023-11-22
Fix incorrect detection of debian during cross-compilation

Lorac:
If using Debian as a host system during a yocto build, that condition will be true since it will check the host system. During a yocto build, it shouldn't rely on host configuration.


evelikov:
The relevant code was removed with https://github.com/sddm/sddm/pull/1856

So this PR is safe to close /cc @hsitter @Vogtinator

PR #1842, opened 2023-12-28
cmake: Add option to disable installing PAM files

Conan-Kudo:
This is primarily useful for development purposes.


Conan-Kudo:
FYI @aleixpol and @davidedmundson, I finally got around to doing this...

services/redhat.sddm-autologin.pam
```diff
@@ -0,0 +1,20 @@
+ #%PAM-1.0
```
Vogtinator: Whitespace?

services/redhat.sddm.pam
```diff
@@ -0,0 +1,23 @@
+auth     [success=done ignore=ignore default=bad] pam_selinux_permit.so
```
Vogtinator: No `#%PAM-1.0`?

Vogtinator (changes requested):
Whitespace error.

I wonder what the benefit of pam files here in sddm instead of in downstream packaging is. It's just more bureaucracy without much benefit, as upstream reviewers can't really review distro specific changes.

services/redhat.sddm-autologin.pam
Conan-Kudo: Fixed.

services/redhat.sddm.pam
Conan-Kudo: Fixed.

Conan-Kudo:
> I wonder what the benefit of pam files here in sddm instead of in downstream packaging is. It's just more bureaucracy without much benefit, as upstream reviewers can't really review distro specific changes.

I was asked to submit them so that folks hacking on SDDM could easily do so on Fedora.

Vogtinator:
> > I wonder what the benefit of pam files here in sddm instead of in downstream packaging is. It's just more bureaucracy without much benefit, as upstream reviewers can't really review distro specific changes.
> 
> I was asked to submit them so that folks hacking on SDDM could easily do so on Fedora.

In that case just not overwriting the distro provided PAM files with `cmake --install` might be the best option. That's how I did it so far, maybe that just needs to be made easier and documented?

Conan-Kudo:
Yeah. I think that would make sense too.

Conan-Kudo:
@Vogtinator I've updated this to add an option to not install PAM files instead.

Vogtinator (approved):
LGTM. Though I wonder whether the default should be "OFF" or even the PAM services removed completely to avoid mismatches. The current fallback is Arch specific and won't work anywhere else.

PR #1854, opened 2024-01-20
Add nushell-specific commands to Xsession and wayland-session

aionescu:
Add special cases in `Xsession` and `wayland-session` to load nushell's login environment if the user's shell is `nu`.

### DIFF
--- a/data/scripts/Xsession
+++ b/data/scripts/Xsession
@@ -23,6 +23,11 @@ if [ -z "$SDDM_XSESSION_PROFILE_READ" ]; then
       [ -f $HOME/.profile ] && . $HOME/.profile
       exec $SHELL --login -c 'exec $argv' $0 "$@"
       ;;
+    */nu)
+      [ -f /etc/profile ] && . /etc/profile
+      [ -f $HOME/.profile ] && . $HOME/.profile
+      exec $SHELL --login -c "exec $0 '$@'"
+      ;;
     *) # Plain sh, ksh, and anything we do not know.
       [ -f /etc/profile ] && . /etc/profile
       [ -f $HOME/.profile ] && . $HOME/.profile
--- a/data/scripts/Xsession
+++ b/data/scripts/Xsession
@@ -26,7 +26,7 @@ if [ -z "$SDDM_XSESSION_PROFILE_READ" ]; then
     */nu)
       [ -f /etc/profile ] && . /etc/profile
       [ -f $HOME/.profile ] && . $HOME/.profile
-      exec $SHELL --login -c "exec $0 '$@'"
+      exec $SHELL --login -c "exec $0 '$*'"
       ;;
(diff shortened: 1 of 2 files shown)


Vogtinator:
Currently there's a discussion to just remove the user shell environment reading altogether as it's too fragile and users shouldn't rely on it, see the various PRs and issues about issues with fish...

Vogtinator:
Blocked by #1779

Vogtinator:
> Blocked by https://github.com/sddm/sddm/pull/1779

Got merged - could you rebase?

aionescu:
Just rebased.

Nu doesn't support passing arguments to commands when executed with `nu -c`, so I had to expand the arguments directly into the string passed to `-c`.

I've done some quick testing and the expansion seems to work correctly, but please double-check.

Vogtinator:
> Nu doesn't support passing arguments to commands when executed with nu -c, so I had to expand the arguments directly into the string passed to -c.

That's highly annoying. I don't think the quoting is correct:

`exec $SHELL --login -c "exec $@"`

If $@ is e.g. `(foo bar)`, then this would result in `exec $SHELL --login -c "exec foo" "bar"`

`exec $SHELL --login -c "exec $0 '$@'"`

If $@ is e.g. `(foo bar)`, then this would result in `exec $SHELL --login -c "exec 'foo" "bar'"`

aionescu:
Oh yeah, you're right. It needs to be `$*` instead of `$@` inside the quotes.

~~But even then, that only works if the arguments have no spaces (otherwise e.g. `"foo bar" baz` gets split to `foo bar baz`). Can we guarantee that about the script's arguments?~~ Nevermind, arguments with spaces aren't handled properly anyway, since the whole command-line is in a single string.

Nu allows passing arguments to script _files_, so a more proper alternative would be to write a script into a temp file, then `exec $SHELL --login $tmp $0 "$@"`. The script would first `rm $tmp` (cause Nu already loaded it into memory) before `exec`ing its arguments.

aionescu:
On that note, isn't it a problem that spaces in arguments aren't preserved? Regardless of the Nu stuff, running Xsession with `'echo "foo   bar" baz'` as the command-line will print `"foo bar" baz` instead.

Vogtinator:
> Nu allows passing arguments to script _files_, so a more proper alternative would be to write a script into a temp file, then `exec $SHELL --login $tmp $0 "$@"`. The script would first `rm $tmp` (cause Nu already loaded it into memory) before `exec`ing its arguments.

Can you use `/dev/stdin` as script file? e.g.

```
exec $SHELL --login /dev/stdin $0 "$@" <<EOF
exec foo bar
EOF
```

aionescu:
> Can you use `/dev/stdin` as script file?

Unfortunately not, Nu says `Could not access file`.

PR #1058, opened 2018-07-23
Fix kde and gnome wallet apps auto open in pam

oniGino:
This is the fix for this issue as described in #1004

PR #1856, opened 2024-01-24
Remove distro specific PAM modules

evelikov:
In light of https://github.com/sddm/sddm/pull/1842 instead of getting more distro-specific PAM modules in sddm, we obtained a way to not install the default ones. Here we remove the Debian and FreeBSD variants.

As result upstream devs does not need to worry if https://github.com/sddm/sddm/pull/1058 is applicable for the distro PAM modules, workarounds like https://github.com/sddm/sddm/pull/1833/ become obsolete. Distributions can change their PAM policy without any patching or PRs.

/cc @arrowd and @carlosdem who introduced them.

Team looking at both distros - Debian and FreeBSD - it seems like they both patch a) some paths and b) the default theme. To me it seems like these can be made `cmake -Dfoobar=baz` modules?

HTH o/


arrowd:
I don't get it. What are we supposed to do on the FreeBSD side now?

evelikov:
> I don't get it. What are we supposed to do on the FreeBSD side now?

- build with `cmake -DINSTALL_PAM_CONFIGURATION=OFF ....`
- keep the complete pam files locally in freebsd ports

In case it's not obvious, I'm not 100% sure if this will fly with maintainers. So consider their input more than mine ;-)

arrowd:
I don't see the rationale for this change. It also breaks installing SDDM directly from source (well not installing, but using SDDM installed from source).

Conan-Kudo:
If we were going to do this, we'd drop _all_ of them including the Arch ones. The reason why I did #1842 wasn't because I agreed with the base premise, but because it makes sense to offer the ability to not install them and use what the system already has.

That said, even GDM keeps PAM files for the various operating systems that are supported because PAM configs vary wildly and having the ability to update them when the GDM code changes is valuable.

carlosdem:
debian specific pam modules were introduced in https://github.com/sddm/sddm/pull/1440 by @hsitter .  the original premise is still valid i think : " this prevents people from ending up with
a broken setup because we install unsuitable pam configs."

evelikov:
Are you saying that the nameless ones (aka sddm.pam and co) are used solely by Arch? My assumption was that they're used across multiple distros, although I could be wrong. Especially since Arch doesn't use elogind.

Glancing across GDM - as an example, fwiw - it has Arch, Exterbo, LFS, OE and RH/Fedora PAM config in-tree.

Debian seemingly generates a [pseudo(?) config in their packaging](https://salsa.debian.org/gnome-team/gdm/-/blob/e6a5685fd88a10da6b5ef109ccde9562eaf9292c/debian/rules?ref_type=heads#L71) which then gets updated via their custom tool, [which gets called by a patched gdm.service](https://salsa.debian.org/gnome-team/gdm/-/blob/e6a5685fd88a10da6b5ef109ccde9562eaf9292c/debian/patches/92_systemd_unit.patch?ref_type=heads). Overall it seems a bit over-engineered although I could easily be missing something.

Regardless: my initial train of thought stands, although take it with a pinch of salt.

Possible alternative is to rename `INSTALL_PAM_CONFIGURATION` to `PAM_CONFIGURATION`, which takes freebsd, debian, arch and none as tokens. With the default being "none".

Conan-Kudo:
> Are you saying that the nameless ones (aka sddm.pam and co) are used solely by Arch? My assumption was that they're used across multiple distros, although I could be wrong. Especially since Arch doesn't use elogind.

Yes. They are exclusively used for Arch and are not used by any other distribution. They cannot be, since they depend on how Arch sets up PAM.

evelikov:
Fwiw: Glancing over at Gentoo - it uses the default (Arch?) PAM with a trivial change `s/include/substack/`

hsitter:
The basic requirement here is that if you install sddm from source you should not ever end up with a broken pam setup; that was the original intent behind multiple configs. The sddm.pam didn't work 100% on Debian so installing it on Debian just broke things.

The way I see it we have a fork in the road:

**Option a)** Drop all configs. 99% of the time the user has sddm already installed and will have a distro provided config. No need to install one really.

**Option b)** Keep things mostly the same and maintain configs in-tree like GDM does. Also following Emil's idea of renaming `INSTALL_PAM_CONFIGURATION` to `PAM_CONFIGURATION=none|debian|freesbd|...`. With one additional thought: I'd still default to detect the system if possible. I wouldn't want to put it on the user|contributor to know that they have to set things up - things should just work.

@davidedmundson @Vogtinator @aleixpol @plfiorini do you have a take on this?

Conan-Kudo:
I'm fine with either path, and if we do option B, then I'll submit to include the redhat/fedora PAM configuration too.

Vogtinator:
I'm in favor of a), see also my comments in https://github.com/sddm/sddm/pull/1842.

plfiorini:
I'm in favor of a) because maintaining distro specific stuff here is difficult

evelikov:
Thanks for the input everyone.

Rather busy this week. When I'm back I will:
 - update the PR removing all the PAM modules - option a)
 - poke the distribution maintainers so they're aware -> reduce odds of accidental breakage

evelikov:
Updated the PR to also drop the the final (Arch) PAM files - aka implemented a)

Tagging some people:
 - Debian @delta-one @hsitter @ptoscano
 - Gentoo @a17r
 - FreeBSD @arrowd
 - Arch @antonio-rojas @felixonmars

Team, with this PR the PAM files are removed from SDDM as the general inclination is that those are better suited within the distribution packaging/repo. Please consider updating the SDDM package and sharing this news with fellow package maintainers.

Thanks in advance o/

Conan-Kudo (approved):

hsitter (approved):

edt-xx:
This may be a stupid question.  What has to change in the package for this change to work?  I am on Arch and the aur package maintainer seems to think the problem is here, and here you are saying the package needs updates...  Please provide a little more info so this can be resolved.

Vogtinator:
The package has to provide PAM service files suitable for the target distro. SDDM no longer provides any.

Vogtinator:
(There should ideally be some documentation around this, but for git builds it's always possible to have breaking changes)

edt-xx:
I've no problems with builds breaking.  I'd just love to see docs - it okay for builds to break, its a real PITA when there is a lack of info on how to fix them.  For now, guess I will need to wait (and use an older sddm-git build) for someone with more pam/sddm/arch know how to fix things on the Arch side.  BTW your very prompt replies are appreciated :-)

felixonmars:
The AUR maintainer appears to have applied a quick revert of this commit for the time being: https://aur.archlinux.org/cgit/aur.git/commit/?h=sddm-git&id=6b182d021decff2983182ecbe2e49eb26d93f154 so updating should be fine for now.

The Arch official repository package will be adjusted to provide the pam files later whenever a new release appears.

sirlucjan:
> The AUR maintainer appears to have applied a quick revert of this commit for the time being: https://aur.archlinux.org/cgit/aur.git/commit/?h=sddm-git&id=6b182d021decff2983182ecbe2e49eb26d93f154 so updating should be fine for now.
> 
> The Arch official repository package will be adjusted to provide the pam files later whenever a new release appears.

I got the first feedback and the temporary solution works.

Vogtinator:
> (There should ideally be some documentation around this, but for git builds it's always possible to have breaking changes)

https://github.com/sddm/sddm/pull/1883 adds at least some info.

PR #1858, opened 2024-01-28
Try to fix the CI

Vogtinator:
liri-project/qmllint silently switched to Qt 6 and the status output is also broken.

Pin it to a specific hash and maybe fix the output by setting the env only for that job.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -14,17 +14,7 @@ on:
       - reopened
       - edited
 
-env:
-  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
 jobs:
-# No longer works as of 2020-11-03 due to npm 7.0.3 -> 7.0.7
-#  wip:
-#    if: "!contains(github.event.head_commit.message, 'ci skip') && github.event_name == 'pull_request'"
-#    runs-on: ubuntu-latest
-#    steps:
-#      - name: Work in progress
-#        uses: wip/action@master
   qml:
     if: "!contains(github.event.head_commit.message, 'ci skip')"
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Vogtinator:
@Conan-Kudo Could you have a look at the broken runs for CentOS and Fedora Rawhide as well?
CentOS has an unresolvable clazy and Fedora a segfaulting clang. The latter is surprising, I wouldn't expect such a broken compiler in a distro. Maybe it ended up installing conflicting packages somehow?

Conan-Kudo:
Rawhide's clang segfault is caused by clazy failing to build in Rawhide. @grulja, can you take a look at it?

EPEL 9 Next's clazy needs to be rebuilt. @tdawson, can you look into this?

Conan-Kudo:
Gah, wrong Jan Grulich! @grulja, Rawhide's clang segfault is caused by clazy failing to build in Rawhide, can you look into it?

grulja:
> Gah, wrong Jan Grulich! @grulja, Rawhide's clang segfault is caused by clazy failing to build in Rawhide, can you look into it?

Done. I updated to the latest snapshot and it's building now.

Vogtinator:
@aleixpol @davidedmundson Ping. All other PRs are blocked by this.

Conan-Kudo:
The Fedora buildsystem is down at the moment, but once it's back up, I can get a new build of clazy for CentOS Stream 9 that should fix the CI failure.

Vogtinator:
It's mostly about the qml check which is required. The build check failures aren't urgent, they're not required and will automatically fix themselves once the distros update.

Conan-Kudo:
clazy should be fixed now in CentOS.

Vogtinator:
@aleixpol @davidedmundson Yet another ping.

Vogtinator:
> clazy should be fixed now in CentOS.

Almost. Looks like binutils doesn't match.

davidedmundson (approved):

PR #1860, opened 2024-01-29
Switch the Wayland greeter default to use Weston's Kiosk shell

Conan-Kudo:
Weston introduced the kiosk shell in Weston 9.0 and deprecated the fullscreen shell in Weston 12.0.

### DIFF
--- a/data/man/sddm.conf.rst.in
+++ b/data/man/sddm.conf.rst.in
@@ -170,7 +170,7 @@ The `UserAuthFile=` option was removed, the file is always created as
 
 `CompositorCommand=`
         Path of the compositor to execute when starting the greeter.
-        Default value is "weston --shell=fullscreen-shell.so".
+        Default value is "weston --shell=kiosk".
 
 `SessionDir=`
 	Comma-separated list of directories containing session files.
--- a/src/common/Configuration.h
+++ b/src/common/Configuration.h
@@ -78,7 +78,7 @@ namespace SDDM {
         );
 
         Section(Wayland,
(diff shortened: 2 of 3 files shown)


Conan-Kudo:
I've marked this as draft because I don't know how we want to handle this. My understanding is that this functionality does not work properly in Qt 5 without the KDE patch set. If we're fine with assuming that it will be built against and run with Qt 5 only with the KDE Patch Collection applied, then we can land this now. Otherwise we have to wait until we strip out Qt 5 support and go Qt 6-only.

cc: @fooishbar, @davidedmundson

Vogtinator:
> I've marked this as draft because I don't know how we want to handle this. My understanding is that this functionality does not work properly in Qt 5 without the KDE patch set.

In which way?

Conan-Kudo:
I don't remember exactly where I read it, but apparently xdg-toplevel on xdg-shell doesn't work properly with QtWayland without the KDE Patch Collection?

Conan-Kudo:
Ah, here's where I saw it. It was a comment by @davidedmundson here: https://github.com/sddm/sddm/pull/1393#issuecomment-870502765

Vogtinator:
Ok. IMO checking the Qt version during buildtime is close enough, let's just assume that any distro with Qt >= 5.15.10 uses the patch collection?

Conan-Kudo:
Yeah, I would say so.

fooishbar:
Yeah the Weston patch is great, thanks.

Conan-Kudo:
@Vogtinator can we land this, please? Weston is working to remove fullscreen shell, and this is the only user of it I can find in Fedora.

Vogtinator (approved):

PR #1862, opened 2024-01-31
Align XDG_SESSION_DESKTOP value with other display managers

alebastr:
The only available specification[1] for the variable is not really helpul as nobody else complies with it:
GDM[2] and LightDM[3] are using the same value for `DESKTOP_SESSION`, `XDG_SESSION_DESKTOP` and `GDMSESSION` (sddm doesn't set the last one).

Potentially breaking, as it will change `X_S_D=KDE` to `X_S_D=plasma`, but the check like that would already fail with any other display manager. At the very least, it unbreaks the valid cases where
 - `DesktopNames` key is not set
 - `DesktopNames` is set to a list of values

[1]: https://www.freedesktop.org/software/systemd/man/pam_systemd.html#desktop=
[2]: https://gitlab.gnome.org/GNOME/gdm/-/blob/a76a68c7225fba00de199139a74a8eed28c576b5/daemon/gdm-session.c#L2873
[3]: https://github.com/canonical/lightdm/blob/de3606561b5107aabebca3e624fdf413ae61924f/src/seat.c#L975

### DIFF
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -441,7 +441,7 @@ namespace SDDM {
         env.insert(QStringLiteral("XDG_SEAT"), seat()->name());
         env.insert(QStringLiteral("XDG_VTNR"), QString::number(m_sessionTerminalId));
 #ifdef HAVE_SYSTEMD
-        env.insert(QStringLiteral("XDG_SESSION_DESKTOP"), session.desktopNames());
+        env.insert(QStringLiteral("XDG_SESSION_DESKTOP"), session.desktopSession());
 #endif
 
         if (session.xdgSessionType() == QLatin1String("x11")) {


Vogtinator (changes requested):
We can't do this, it would severely break Plasma with systemd startup.

This is directly used for matching `OnlyShowIn`.

Using the basename of the desktop session file is IMO not stable enough. It would already break for e.g. the `plasmawayland-dev` session.

Vogtinator:
> The only available specification[1](https://www.freedesktop.org/software/systemd/man/pam_systemd.html#desktop=) for the variable is not really helpul as nobody else complies with it:
GDM[2](https://gitlab.gnome.org/GNOME/gdm/-/blob/a76a68c7225fba00de199139a74a8eed28c576b5/daemon/gdm-session.c#L2873) and LightDM[3](https://github.com/canonical/lightdm/blob/de3606561b5107aabebca3e624fdf413ae61924f/src/seat.c#L975) are using the same value for DESKTOP_SESSION, XDG_SESSION_DESKTOP and GDMSESSION (sddm doesn't set the last one).

The specification is fairly clear:

> For example: "GNOME", or "KDE". It is recommended to use the same identifiers and capitalization as for $XDG_CURRENT_DESKTOP, as defined by the [Desktop Entry Specification](https://standards.freedesktop.org/desktop-entry-spec/latest/).

Which says:

> $XDG_CURRENT_DESKTOP should have been set by the login manager, according to the value of the DesktopNames found in the session file. The entry in the session file has multiple values separated in the usual way: with a semicolon.

And I do not see any reason to diverge from that.

alebastr:
> We can't do this, it would severely break Plasma with systemd startup.
> 
> This is directly used for matching `OnlyShowIn`.

Nope, that's `XDG_CURRENT_DESKTOP`. `XDG_SESSION_DESKTOP` is not used anywhere except logind session info AFAIK.

> The specification is fairly clear:
> And I do not see any reason to diverge from that.

1. It clearly says 'recommended'.
2. Nobody implements that recommendation. Please, check the links to GDM and LightDM implementations.

Vogtinator:
> > We can't do this, it would severely break Plasma with systemd startup.
> > This is directly used for matching `OnlyShowIn`.
> 
> Nope, that's `XDG_CURRENT_DESKTOP`. `XDG_SESSION_DESKTOP` is not used anywhere except logind session info AFAIK.

Indeed. I don't really get why both exist. `XDG_CURRENT_DESKTOP` contains multiple values while `XDG_SESSION_DESKTOP` has only one (which is already werid), but "CURRENT" vs "SESSION" is really unclear to me. Apparently it's forwarded to logind to use it as the `Desktop` property of sessions.

> > The specification is fairly clear:
> > And I do not see any reason to diverge from that.
> 
> 1. It clearly says 'recommended'.

Yep.

> 2. Nobody implements that recommendation. Please, check the links to GDM and LightDM implementations.

Please discuss that with systemd upstream. Unless there is some very practical reason to diverge from the existing and recommended behavior, I don't think we should explicitly change it.

alebastr:
> Please discuss that with systemd upstream. Unless there is some very practical reason to diverge from the existing and recommended behavior, I don't think we should explicitly change it.

The whole point of this PR is that existing and commonly accepted behavior does not match either systemd docs or sddm implementation. In this case it's better to stick with what others do.

sddm implementation is additionally flawed as it depends on an optional `DesktopNames` key and will result in an empty value if the property is missing in a session file. Or in multiple values, if `DesktopNames` contains multiple values.

PR #1863, opened 2024-01-31
Allow non-root greeters and sessions to start on kernels without VTs

n3rdopolis:


n3rdopolis:
I am thinking that it appears like it may actually fix the multiseat issue too? It allows a second Wayland session to start on the second seat of this QEMU VM

davidedmundson (approved):

Vogtinator:
I think more than just this is broken for seat != seat0. `src/daemon/Display.cpp` uses `VirtualTerminal` without any consideration that this only makes sense for seat0 and will probably break seat0 in the process.

n3rdopolis:
So in addition to this, how does this look to you?
I am not 100% familiar with SDDM's codebase, and I gotta test this on my VM too...
```
diff --git a/src/daemon/Display.cpp b/src/daemon/Display.cpp
index 4014e0d..0c2e65a 100644
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -116,16 +116,28 @@ namespace SDDM {
         // Create display server
         case X11DisplayServerType:
-            m_terminalId = VirtualTerminal::setUpNewVt();
+            if (seat()->canTty()) {
+                m_terminalId = VirtualTerminal::setUpNewVt();
+            } else {
+                m_terminalId = -1;
+            }
             m_displayServer = new XorgDisplayServer(this);
             break;
         case X11UserDisplayServerType:
-            m_terminalId = fetchAvailableVt();
+            if (seat()->canTty()) {
+                m_terminalId = fetchAvailableVt();
+            } else {
+                m_terminalId = -1;
+            }
             m_displayServer = new XorgUserDisplayServer(this);
             m_greeter->setDisplayServerCommand(XorgUserDisplayServer::command(this));
             break;
         case WaylandDisplayServerType:
-            m_terminalId = fetchAvailableVt();
+            if (seat()->canTty()) {
+                m_terminalId = fetchAvailableVt();
+            } else {
+                m_terminalId = -1;
+            }
             m_displayServer = new WaylandDisplayServer(this);
             m_greeter->setDisplayServerCommand(mainConfig.Wayland.CompositorCommand.get());
             break;
@@ -161,7 +173,9 @@ namespace SDDM {
             }
             // It might be the case that we are trying a tty that has been taken over by a
             // different process. In such a case, switch back to the initial one and try again.
-            VirtualTerminal::jumpToVt(SDDM_INITIAL_VT, true);
+            if (seat()->canTty()) {
+                VirtualTerminal::jumpToVt(SDDM_INITIAL_VT, true);
+            }
             stop();
         });
         connect(m_greeter, &Greeter::displayServerFailed, this, &Display::displayServerFailed);
@@ -421,7 +435,9 @@ namespace SDDM {
         m_sessionTerminalId = m_terminalId;
         if ((session.type() == Session::WaylandSession && m_displayServerType == X11DisplayServerType) || (m_greeter->isRunning() && m_displayServerType != X11DisplayServerType)) {
             // Create a new VT when we need to have another compositor running
-            m_sessionTerminalId = VirtualTerminal::setUpNewVt();
+            if (seat()->canTty()) {
+                m_sessionTerminalId = VirtualTerminal::setUpNewVt();
+            }
         }
 
         // some information
```

n3rdopolis:
OK, more testing (up until now case 4 indeed was not working)

1. if you have just seat0 on a kernel without VTs it works.
2. if you have just seat0 on a kernel with VTs it works.
3. If you have a seat0 on a kernel without VTs, and other seats, it works.
4. If you have a seat0 on a kernel with VTs, and other seats, it works

src/daemon/Display.cpp
```diff
@@ -161,7 +173,9 @@ namespace SDDM {
             }
             // It might be the case that we are trying a tty that has been taken over by a
             // different process. In such a case, switch back to the initial one and try again.
-            VirtualTerminal::jumpToVt(SDDM_INITIAL_VT, true);
+            if (seat()->canTty()) {
```
Vogtinator: This should never be reached

src/daemon/Display.cpp
```diff
@@ -421,7 +435,9 @@ namespace SDDM {
         m_sessionTerminalId = m_terminalId;
         if ((session.type() == Session::WaylandSession && m_displayServerType == X11DisplayServerType) || (m_greeter->isRunning() && m_displayServerType != X11DisplayServerType)) {
             // Create a new VT when we need to have another compositor running
-            m_sessionTerminalId = VirtualTerminal::setUpNewVt();
+            if (seat()->canTty()) {
+                m_sessionTerminalId = VirtualTerminal::setUpNewVt();
```
Vogtinator, line 431: The `env.insert(QStringLiteral("XDG_VTNR"), QString::number(m_sessionTerminalId));` below should probably be conditional depending on `m_sessionTerminalId > 0`

src/daemon/Greeter.cpp
```diff
@@ -197,7 +197,7 @@ namespace SDDM {
             env.insert(QStringLiteral("XDG_SEAT"), m_display->seat()->name());
             env.insert(QStringLiteral("XDG_SEAT_PATH"), daemonApp->displayManager()->seatPath(m_display->seat()->name()));
             env.insert(QStringLiteral("XDG_SESSION_PATH"), daemonApp->displayManager()->sessionPath(QStringLiteral("Session%1").arg(daemonApp->newSessionId())));
-            if (m_display->seat()->name() == QLatin1String("seat0"))
+            if (m_display->seat()->name() == QLatin1String("seat0") && m_display->seat()->canTty())
```
Vogtinator: IMO should check for `m_display->terminalId() > 0`

src/daemon/Seat.cpp
```diff
@@ -141,4 +142,11 @@ namespace SDDM {
             VirtualTerminal::jumpToVt(*nextVt, true);
         }
     }
+
+    bool Seat::canTty() {
+        OrgFreedesktopLogin1ManagerInterface manager(Logind::serviceName(), Logind::managerPath(), QDBusConnection::systemBus());
```
Vogtinator, line 148: This breaks for ConsoleKit2 which does not have CanTTY or similar functionality and also if neither of those are available.

As a fallback, doing `return m_name == QStringLiteral("seat0")` should do or better `return m_name.compare(QStringLiteral("seat0"), Qt::CaseInsensitive) == 0`.

src/daemon/Seat.h
```diff
@@ -35,6 +35,7 @@ namespace SDDM {
 
         const QString &name() const;
         void createDisplay(Display::DisplayServerType serverType);
+        bool canTty();
```
Vogtinator: This looks ugly... `canTTY()` please

src/helper/UserSession.cpp
```diff
@@ -195,6 +195,9 @@ namespace SDDM {
         // the session as stdin so that it stays open without races
         if (hasDisplayServer || waylandUserSession) {
             // open VT and get the fd
+            if (processEnvironment().value(QStringLiteral("XDG_VTNR")).isEmpty())
```
Vogtinator: If unset, it should continue, just without the VT related stuff (`vtFd = -1`)

src/helper/backend/PamBackend.cpp
```diff
@@ -252,8 +252,10 @@ namespace SDDM {
         QProcessEnvironment sessionEnv = m_app->session()->processEnvironment();
         const auto sessionType = sessionEnv.value(QStringLiteral("XDG_SESSION_TYPE"));
         const auto sessionClass = sessionEnv.value(QStringLiteral("XDG_SESSION_CLASS"));
-        QString tty = VirtualTerminal::path(sessionEnv.value(QStringLiteral("XDG_VTNR")).toInt());
-        m_pam->setItem(PAM_TTY, qPrintable(tty));
+        if (!sessionEnv.value(QStringLiteral("XDG_VTNR")).isEmpty()) {
```
Vogtinator: IMO just `sessionEnv.contains(QStringLiteral("XDG_VTNR"))`. Empty is invalid and should be caught

Vogtinator (changes requested):
Breaks without systemd-logind and there are some issues with XDG_VTNR setting

src/daemon/Display.cpp
Vogtinator: To clarify: Currently it ends up with `XDG_VTNR=-1` which causes the helper to open `/dev/tty-1` and then go into the tty-less code path. That does kind of work but is not really how it should be...

src/daemon/Seat.h
n3rdopolis: Sure, that looks better to me too

src/daemon/Seat.cpp
n3rdopolis: How am I detecting the fallback? would an else under a new
`if (QDBusConnection::systemBus().interface()->isServiceRegistered(QStringLiteral("org.freedesktop.login1")))`
work?

n3rdopolis: Oh, and 
`return m_name.compare(QStringLiteral("seat0"), Qt::CaseInsensitive) == 0`
I could also pair it with a `QFile::exists("/dev/tty0")`, not sure if consolekit would work without VTs, but kernels without VTs don't have /dev/tty0 so the check would be more correct if that is OK

Vogtinator: `manager.isValid()` + `seat.property("CanTTY").isValid()` maybe?

src/helper/UserSession.cpp
n3rdopolis: This was needed for multiseat to work on kernels _with_ VTs. Although now that you bring it up, how is it actually working here, I will do more testing

Vogtinator: See https://github.com/sddm/sddm/pull/1863#discussion_r1500446883

n3rdopolis: Ah, OK, I think I can try that

src/daemon/Seat.cpp
n3rdopolis: I can try that, do you agree with also checking for the existence of the /dev/tty0 file in the fallback check too?  

 

EDIT: Actually maybe /sys/class/tty/tty0/active , it's closer to what logind does

Vogtinator: Don't forget about FreeBSD

n3rdopolis: Right. Hmm I wonder if `defaultVtPath` in src/common/VirtualTerminal.cpp  can be used somehow then

src/daemon/Seat.h
n3rdopolis: Done locally

src/daemon/Greeter.cpp
n3rdopolis: Done locally

src/daemon/Display.cpp
n3rdopolis: done locally

src/helper/backend/PamBackend.cpp
n3rdopolis: Done locally

src/daemon/Seat.cpp
n3rdopolis: Done locally

src/daemon/Display.cpp
n3rdopolis: OK, yeah that seems to check out it only gets triggered on ttyFailed which only gets triggered when takeControl is true and it fails, and `vtFd > 0)` right?

I'm seeing that now. Done locally

src/helper/UserSession.cpp
n3rdopolis: done locally, and now it all seems to work, let me try to get these commits in order, and I will push

src/daemon/Seat.cpp
```diff
@@ -141,4 +142,15 @@ namespace SDDM {
...
     }
+
+    bool Seat::canTTY() {
+        OrgFreedesktopLogin1ManagerInterface manager(Logind::serviceName(), Logind::managerPath(), QDBusConnection::systemBus());
+        auto seatPath = manager.GetSeat(m_name);
+        OrgFreedesktopLogin1SeatInterface sessionIface(Logind::serviceName(), seatPath.value().path(), QDBusConnection::systemBus());
+        if (manager.isValid() && sessionIface.property("canTTY").isValid()) {
+            return sessionIface.canTTY();
+        } else {
+            return m_name.compare(QStringLiteral("seat0"), Qt::CaseInsensitive) == 0 && QFile::exists(VirtualTerminal::getDefaultVtPath());
```
Vogtinator: The method wouldn't be necessary, you can use `VirtualTerminal::defaultVtPath` directly. If you want to be fancy even the qstring can be avoided by doing `access(VirtualTerminal::defaultVtPath, F_OK) == 0`

src/daemon/Seat.cpp
```diff
@@ -141,4 +142,15 @@ namespace SDDM {
             VirtualTerminal::jumpToVt(*nextVt, true);
         }
     }
+
+    bool Seat::canTTY() {
+        OrgFreedesktopLogin1ManagerInterface manager(Logind::serviceName(), Logind::managerPath(), QDBusConnection::systemBus());
+        auto seatPath = manager.GetSeat(m_name);
+        OrgFreedesktopLogin1SeatInterface sessionIface(Logind::serviceName(), seatPath.value().path(), QDBusConnection::systemBus());
```
Vogtinator: I'm wondering whether this works as expected if there's no login1. in that case, `manager` is already invalid so `manager.GetSeat` is even more invalid and `seatPath.value().path()` yet more invalid. Does it print a warning at some point or even abort?

src/daemon/Seat.cpp
n3rdopolis: Do I need to include something additional?
Trying
`return m_name.compare(QStringLiteral("seat0"), Qt::CaseInsensitive) == 0 && access(VirtualTerminal::defaultVtPath, F_OK) == 0);`
I get
`src/daemon/Seat.cpp:153:96: No member named 'defaultVtPath' in namespace 'SDDM::VirtualTerminal'`

Vogtinator: Hm, is it missing a declaration in `VirtualTerminal.h`?

src/daemon/Seat.cpp
n3rdopolis: Is there a way I can quickly test that on recent Ubuntu?

Vogtinator: You could probably fake it by forcing it to use the session bus instead of the system bus

src/daemon/Seat.cpp
n3rdopolis: I figured this one out

src/daemon/Seat.cpp
n3rdopolis: I think I will just change it to match what `displayStopped` does

n3rdopolis: I structured it like that, but now I am realizing that logind has had CanTTY since v186 which was released in 2012, so I don't think that that function needs to be separately probed for now

Vogtinator: logind has, but not consolekit. `Logind::serviceName()` returns `org.freedesktop.ConsoleKit` if CK(2) is available.

Vogtinator: You could possibly hardcode `org.freedesktop.login1` here

n3rdopolis: ok, although I wonder about elogind...

n3rdopolis: I feel like hardcoding a check against ConsoleKit might be better than a check for logind.
I can change it if you feel different

Vogtinator: IMO it should look at `CanTTY` - if it exists, use it, if not, use the fallback

n3rdopolis: How's this? if CanTTY doesn't exist, it doesn't hit the early return, and falls back to the 'manual' check

Vogtinator: Looks good! Gave it a try locally as well, no warnings or other complaints.

src/daemon/Display.cpp
```diff
@@ -116,16 +116,28 @@ namespace SDDM {
         // Create display server
         switch (m_displayServerType) {
         case X11DisplayServerType:
-            m_terminalId = VirtualTerminal::setUpNewVt();
+            if (seat()->canTTY()) {
+                m_terminalId = VirtualTerminal::setUpNewVt();
+            } else {
+                m_terminalId = -1;
```
Vogtinator: Can you set the default of `m_terminalId` in the header to `-1` and remove the assignments here?

Vogtinator (approved):
One style comment, LGTM otherwise

src/daemon/Display.cpp
n3rdopolis: Sure!

n3rdopolis: Done

PR #1870, opened 2024-02-21
bump to 0.21.0 and start changelog

hsitter:


Conan-Kudo:
I would really like #1860 landed before we make a release, as Weston is working to rip out the legacy fullscreen shell mode.

Vogtinator:
Unfortunately #1863 isn't quite ready yet for 0.21.0 so let's go ahead with a release.

Here's my suggestion to complete the changelog:

```
    + sddm-greeter can now be built for both Qt5 and Qt6
    + Themes can now declare their Qt version compatibility to control which greeter gets used
    + New CMake option INSTALL_PAM_CONFIGURATION (defaults to ON)
    + Bulgarian translation added
    * DBus policy files are now installed in /usr rather than /etc
    * Do not show keyboard layouts on wayland as switching is not implemented
    * Switch the default commandline for Weston to use the kiosk shell (available in 9.0+)
    - Fix building with Qt 6.7
    - Handle autologin failure by falling back to the greeter (#1775)
    - Fix session names containing ',' (#1745)
    - Don't let tmpfiles delete x11 auth files for very long running sessions
    - Many improvements for the **experimental** wayland support
```

Conan-Kudo:
>     * Do now show keyboard layouts on wayland as switching is not implemented

`    * Do not show keyboard layouts on wayland as switching is not implemented`

Vogtinator (approved):

PR #1876, opened 2024-02-27
Themes: Qt6 compatibiltiy

alebastr:
Disclaimer: not a real QML developer. Please, suggest better ways to fix things if you know how.

See full commit messages for the problem statements, warnings addressed, etc.

Everything's been tested on Fedora 38/Qt5/X11 .. Fedora Rawhide/Qt6/Wayland (Sway) with all 4 included themes.

***
Having serious doubts about `Themes: fix deprecated signal handler declarations`: the new syntax for attached signal handlers is only documented for Qt6 and I can't find when it was introduced in Qt5.

### DIFF
--- a/data/themes/CMakeLists.txt
+++ b/data/themes/CMakeLists.txt
@@ -10,9 +10,15 @@ foreach(THEME ${THEMES})
 
     qt_add_translation(QM_FILES "${TRANSLATION_SOURCES}")
 
-    install(DIRECTORY "${THEME}" DESTINATION "${DATA_INSTALL_DIR}/themes" PATTERN "${THEME}/*.ts"
-            EXCLUDE PATTERN "${THEME}/.gitattributes"
-            EXCLUDE)
+    configure_file("${THEME}/metadata.desktop.in" "${THEME}/metadata.desktop" @ONLY)
+
+    install(DIRECTORY "${THEME}" DESTINATION "${DATA_INSTALL_DIR}/themes"
+            PATTERN "${THEME}/*.in" EXCLUDE
+            PATTERN "${THEME}/*.ts" EXCLUDE
+            PATTERN "${THEME}/.gitattributes" EXCLUDE)
+
+    install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${THEME}/metadata.desktop"
+            DESTINATION "${DATA_INSTALL_DIR}/themes/${THEME}/")
 
     list(APPEND THEMES_QM_FILES ${QM_FILES})
 endforeach(THEME)
--- a/components/2.0/Button.qml
+++ b/components/2.0/Button.qml
@@ -67,7 +67,9 @@ Rectangle {
         }
     ]
 
-    Behavior on color { NumberAnimation { duration: 200 } }
+    transitions: Transition {
+        ColorAnimation { duration: 200 }
+    }
 
     clip: true
     smooth: true
--- a/components/2.0/Button.qml
+++ b/components/2.0/Button.qml
@@ -111,7 +111,7 @@ Rectangle {
         onReleased: { container.focus = true; container.released() }
     }
 
-    Keys.onPressed: {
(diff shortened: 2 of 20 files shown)


Vogtinator:
> Having serious doubts about Themes: fix deprecated signal handler declarations: the new syntax for attached signal handlers is only documented for Qt6 and I can't find when it was introduced in Qt5.

Qt 5.15 warns about this as well, I think it's been introduced in 5.13 so we're safe.

Vogtinator:
Thanks, looks good!

What's missing is to set the `QtVersion` property of the theme desktop files accordingly, so that it actually uses Qt 6 if sddm was built against it.

alebastr:
> Qt 5.15 warns about this as well, I think it's been introduced in 5.13 so we're safe.

Qt5 has one of the two warnings.
```
Parameter "event" is not declared. Injection of parameters into signal handlers is deprecated. Use JavaScript functions with formal parameters instead.
```
was added in 6.2 and refers to the following code
```qml
    Keys.onPressed: {
        if (event.key === Qt.Key_Space) {
    ...
```

The variant with `Keys.onPressed: function (event) { }` does not seem to appear in Qt5.15 documentation.

Vogtinator:
The other warning is in Qt 5 as well: https://github.com/qt/qtdeclarative/commit/efe0bec9468d75b768d1e26d2a8b440ade5ba632

However, it proved to be so annoying that later on this message was changed to be categorized properly and most distros disable logging for that category, as it was recommended by Plasma: https://github.com/qt/qtdeclarative/commit/a2eef6b511988b2435c4e39b6b5551e857ce7775

Vogtinator:
> What's missing is to set the QtVersion property of the theme desktop files accordingly, so that it actually uses Qt 6 if sddm was built against it.

Currently it hardcodes `QtVersion=6`, which breaks the themes if the Qt 6 greeter isn't available. Needs some cmake logic to set it during build, until we drop Qt 5 support.

Vogtinator:
BTW: Apparently I did forget to mention the `QtVersion` metadata.desktop property in the documentation.

I'd make a PR for that but it'd probably conflict with this one anyway. Could you add a sentence like:

```
Themes are run by a Qt 5 built sddm-greeter, unless the `QtVersion` property in metadata.desktop specifies a different version, such as `QtVersion=6` for using `sddm-greeter-qt6`.
```

alebastr:
> The other warning is in Qt 5 as well: https://github.com/qt/qtdeclarative/commit/efe0bec9468d75b768d1e26d2a8b440ade5ba632

Yes, I know that this one is in Qt5. The one that I quoted, `Parameter "event" is not declared.`, is not. So the syntax with functions works on Qt5.15.latest-kde and I haven't tested further.

> Currently it hardcodes `QtVersion=6`, which breaks the themes if the Qt 6 greeter isn't available. Needs some cmake logic to set it during build, until we drop Qt 5 support.

Fixed.

> I'd make a PR for that but it'd probably conflict with this one anyway. Could you add a sentence like:

Done.

data/themes/CMakeLists.txt
```diff
@@ -10,9 +10,15 @@ foreach(THEME ${THEMES})
 
     qt_add_translation(QM_FILES "${TRANSLATION_SOURCES}")
 
-    install(DIRECTORY "${THEME}" DESTINATION "${DATA_INSTALL_DIR}/themes" PATTERN "${THEME}/*.ts"
-            EXCLUDE PATTERN "${THEME}/.gitattributes"
-            EXCLUDE)
+    configure_file("${THEME}/metadata.desktop.in" "${THEME}/metadata.desktop")
```
Vogtinator: `@ONLY` as $ replacements aren't needed

Vogtinator (approved):

data/themes/CMakeLists.txt
alebastr: Thanks! Fixed.

PR #1878, opened 2024-03-04
Notify session manager about idle greeters

### LINKED ISSUE #1302: with no users logged in, sddm does not suspend session as configured in System Settings.
When  kde5 System Settings is set to e.g. put system to sleep after x minutes of  inactivity,
this only works when the user is logged in.

with no users logged in, and the sddm login screen is on  the display, the screen correctly blanks 
the screen after inactivity, but never puts the system to sleep, if that is what is requested
in System Settings.

For consistency sddm should have also  have a feature   
allowing the system to be suspended (sleep/hibernate/hybrid sleep, etc)
if x minute has elapsed with no one logged in.

(obviously this would be off by default, for servers). And would be an "all-user" policy.

Vogtinator:
If a greeter doesn't receive input for 5s, set IdleHint for its session to
true. That way it can participate in session manager based idle handling,
such as automatic suspend.

Fixes https://github.com/sddm/sddm/issues/1302

Draft because:

- [ ] Needs testing
- [ ] Is there a better option than an event filter like this? Is restarting a timer too expensive?
- [ ] Does the idle monitor need to be stopped explicitly during login attempts? If not, it's possible that it triggers while a login is in progress, leaving `IdleHint` true when the DE takes over, which may never reset it.


aviallon:
Thanks for your work @Vogtinator, I was looking to implement it myself since I currently need that.

tadghh:
@Vogtinator CentOS build failed, something about libctf missing.

Vogtinator:
> @Vogtinator CentOS build failed, something about libctf missing.

You can ignore that kind of failures for CentOS and Fedora Rawhide, they're really unreliable (and should probably be removed for that reason)

tadghh:
> > @Vogtinator CentOS build failed, something about libctf missing.
> 
> You can ignore that kind of failures for CentOS and Fedora Rawhide, they're really unreliable (and should probably be removed for that reason)

Oh haha thats a little silly. Just wanted to bump this incase it got left behind, this would some polish for laptop and desktop users. Don't want to end up with burnt in displays. This change brings a lot to the UX

jfroy:
I'm not sure this is the right approach. The `IdleAction` are too limited. I more want to match the energy settings offered by Plasma. For my specific workstation, that's dim the screens after 5 minutes, turn off the screens after 10 minutes. I do not want to suspend, hibernate, reboot, or power off the system.

Vogtinator:
> For my specific workstation, that's dim the screens after 5 minutes, turn off the screens after 10 minutes. I do not want to suspend, hibernate, reboot, or power off the system.

That would require half a Plasma session to handle. Might actually happen in the not too distant future.

jfroy:
> > For my specific workstation, that's dim the screens after 5 minutes, turn off the screens after 10 minutes. I do not want to suspend, hibernate, reboot, or power off the system.
> 
> That would require half a Plasma session to handle. Might actually happen in the not too distant future.

I am really not familiar with Plasma in general, but would it not “just” require powerdevil and kwin? I’m already using kwin as the sddm Wayland compositor.

Vogtinator:
> > > For my specific workstation, that's dim the screens after 5 minutes, turn off the screens after 10 minutes. I do not want to suspend, hibernate, reboot, or power off the system.
> > 
> > 
> > That would require half a Plasma session to handle. Might actually happen in the not too distant future.
> 
> I am really not familiar with Plasma in general, but would it not “just” require powerdevil and kwin? I’m already using kwin as the sddm Wayland compositor.

Basically, which is already a lot

gillesgagniard:
> I'm not sure this is the right approach. The `IdleAction` are too limited. I more want to match the energy settings offered by Plasma. For my specific workstation, that's dim the screens after 5 minutes, turn off the screens after 10 minutes. I do not want to suspend, hibernate, reboot, or power off the system.

Hi there ! While I agree we could imagine better solutions in the future (this PR is already 1 year old, and alternatives have not materialized so far), this PR, which I've tested and works perfectly, already brings a lot of value, namely preventing monitor burn-in after unsupervised reboots.

Anything preventing its integration ?

Vogtinator:
> Anything preventing its integration ?

Justs the three items in the description. If you were able to test this successfully, 2/3 are completed though I'd say.

aviallon:
Thank you, will cherry-pick this for my system.

PR #1883, opened 2024-03-06
Add info about required PAM services to INSTALL.md

Vogtinator:
I hope this is clear enough without being too specific... Examples might help, but not sure how they should look like.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -25,6 +25,20 @@ For documentation on standard CMake variables, see:
 By default, SDDM runs as its own user. An `sddm` user needs to be created, with
 its home set to `/var/lib/sddm` by default.
 
+Several PAM services need to be created to set up authentication. As distributions
+have different methods for PAM service configuration, SDDM does not provide them
+itself anymore.
+
+* `sddm`: Used for user sessions with password-based authentication.
+  Authentication must be password-based (e.g. `pam_unix`), the session needs
+  to be set up and managed if applicable (e.g. `pam_systemd`).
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -25,6 +25,20 @@ For documentation on standard CMake variables, see:
...
+* `sddm`: Used for user sessions with password-based authentication.
+  Authentication must be password-based (e.g. `pam_unix`), the session needs
+  to be set up and managed if applicable (e.g. `pam_systemd`).
+* `sddm-autologin`: Used for user sessions without authentication. Like `sddm`,
+  without modules asking for passwords.
+* `sddm-greeter`: Used for the greeter session. Authentication must always
+  succeed (`pam_permit`). Session should use modules appropriate for the
+  greeter environment, which can be identical to user sessions.
+
+
```
evelikov, line 41: As you mentioned examples, IMHO it's fine to refer people to their distribution (vaguely as below).

The example might be useful for people creating their own distro and knowing little about PAM modules. Then again, chances are it will miss-behave in one way or another. 

```
If you are interested in specifics or examples, please check your distribution of choice.
```

evelikov (approved):

INSTALL.md
Vogtinator: Could be one, but arguably this is targeted at distro packagers anyway so that line would not be of much help. Users shouldn't care about this, they'll just end up using the PAM services provided by the distro's sddm.

Maybe that could be mentioned, hm.

PR #1895, opened 2024-03-16
Enable build without libxau and libxcb dependencies

### LINKED ISSUE #1420: Wayland: Remove all remaining X11 dependencies
First of all, congratulations and a big Thanks on finally landing #1393!

I went over the code very briefly, and it looks like there are still a few X11/Xorg dependencies remaining in the Wayland code. For example, `WaylandKeyboardBackend` doesn't look implemented, and [still uses X11 symbols](https://github.com/sddm/sddm/pull/1393/files#diff-aaa62d813d24c4933a2edc519a9fd6e5658e78086c4f9658f3398002310cfa8eR42):
```c++
void WaylandKeyboardBackend::init()
{
    d->layouts.clear();

    QDir dir(QStringLiteral("/usr/share/X11/xkb/symbols"));
    auto entries = dir.entryList(QDir::Files);
    for (const auto &entry : qAsConst(entries))
        d->layouts << new KeyboardLayout(entry, entry);
}
```
This should be using `libinput`, I think?

Likewise, `WaylandHelper` is [still using `XAuth`](https://github.com/sddm/sddm/pull/1393/files#diff-ff9e48305e347a6c73ce70a26a106adf1d187f110e94e0ed3bacf3cdb80c5570R25). Not sure what the alternative here is. Probably `logind`?

Here is the full list of X dependencies for version `0.19`, according to [Ubuntu packages](https://packages.ubuntu.com/impish/sddm):
> - `libxcb-xkb1`: X C Binding, XKEYBOARD extension 
> - `libxcb1`: X C Binding 
> - `x11-common`: X Window System (X.Org) infrastructure 
> - `xauth`: X authentication utility 
> - `xserver-xorg`: X.Org X server 
> or `xserver` (virtual package)

## Goal
*All* of these dependencies should be optional when running with Wayland. It should be possible to install and run SDDM/Wayland without *any* X dependencies.

LittleHuba:
This PR enables building SDDM without dependencies on libxcb and libxau by setting the CMake flag `BUILD_WITH_X11 =OFF`. This way, it is possible to build and use SDDM on Wayland-only systems. By default this flag is set to `ON`.

I kept preprocessor flags in the code to a minimum by using runtime polymorphism wherever feasible. In a few places runtime polymorphism would have caused a high overhead. That is where I fell back on the preprocessor flag. I see those as future places for refactoring.

Fixes: https://github.com/sddm/sddm/issues/1420

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,7 @@ option(ENABLE_JOURNALD "Enable logging to journald" ON)
 option(NO_SYSTEMD "Disable systemd support" OFF)
 option(USE_ELOGIND "Use elogind instead of logind" OFF)
 option(BUILD_WITH_QT6 "Build with Qt 6" OFF)
+option(BUILD_WITH_X11 "Build with X11 support" ON)
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
@@ -65,10 +66,14 @@ find_package(PkgConfig)
 find_package(PAM REQUIRED)
 
 # XAU
-pkg_check_modules(LIBXAU REQUIRED "xau")
+if(BUILD_WITH_X11)
+    pkg_check_modules(LIBXAU REQUIRED "xau")
+endif ()
 
 # XCB
-find_package(XCB REQUIRED)
+if(BUILD_WITH_X11)
+    find_package(XCB REQUIRED)
+endif ()
 
 # XKB
 find_package(XKB REQUIRED)
@@ -171,6 +176,13 @@ endif()
 # utmps
 find_package(utmps)
 add_feature_info("utmps" UTMPS_FOUND "utmps support")
+
+# x11
+if(BUILD_WITH_X11)
+    add_definitions(-DBUILD_WITH_X11=1)
+else ()
+    add_definitions(-DBUILD_WITH_X11=0)
+endif ()
 
 # Set constants
 set(DATA_INSTALL_DIR            "${CMAKE_INSTALL_FULL_DATADIR}/sddm"                CACHE PATH      "System application data install directory")
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -24,8 +24,10 @@
 #include "Configuration.h"
 #include "DaemonApp.h"
 #include "DisplayManager.h"
+#if BUILD_WITH_X11 == 1
 #include "XorgDisplayServer.h"
 #include "XorgUserDisplayServer.h"
+#endif
 #include "Seat.h"
 #include "SocketServer.h"
 #include "Greeter.h"
@@ -113,8 +115,10 @@ namespace SDDM {
         m_socketServer(new SocketServer(this)),
         m_greeter(new Greeter(this))
     {
+
         // Create display server
         switch (m_displayServerType) {
+#if BUILD_WITH_X11 == 1
         case X11DisplayServerType:
             if (seat()->canTTY()) {
                 m_terminalId = VirtualTerminal::setUpNewVt();
@@ -126,8 +130,16 @@ namespace SDDM {
                 m_terminalId = fetchAvailableVt();
             }
             m_displayServer = new XorgUserDisplayServer(this);
-            m_greeter->setDisplayServerCommand(XorgUserDisplayServer::command(this));
+            m_greeter->setDisplayServerCommand(m_displayServer->getCommand(this));
             break;
+#else
+        case X11DisplayServerType:
+        case X11UserDisplayServerType:
+            // Reset type to Wayland and fall through to initialization
+            qWarning() << "Built without X11 support. Trying Wayland display server.";
+            m_displayServerType = WaylandDisplayServerType;
+            [[fallthrough]];
+#endif
         case WaylandDisplayServerType:
             if (seat()->canTTY()) {
                 m_terminalId = fetchAvailableVt();
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -491,8 +491,10 @@ namespace SDDM {
                 manager.UnlockSession(m_reuseSessionId);
                 manager.ActivateSession(m_reuseSessionId);
             } else {
-                if (qobject_cast<XorgDisplayServer *>(m_displayServer))
-                    m_auth->setCookie(qobject_cast<XorgDisplayServer *>(m_displayServer)->cookie());
+                if (m_displayServerType == X11DisplayServerType) {
+                    const QByteArray cookie = m_displayServer->getCookie();
+                    m_auth->setCookie(cookie);
+                }
             }
 
             // save last user and last session
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -462,14 +462,11 @@ namespace SDDM {
         env.insert(QStringLiteral("XDG_SESSION_DESKTOP"), session.desktopNames());
 #endif
 
-        if (session.xdgSessionType() == QLatin1String("x11")) {
-          if (m_displayServerType == X11DisplayServerType)
+        if (session.xdgSessionType() == QLatin1String("x11") && m_displayServerType == X11DisplayServerType) {
             env.insert(QStringLiteral("DISPLAY"), name());
-          else
-            m_auth->setDisplayServerCommand(XorgUserDisplayServer::command(this));
         } else {
-            m_auth->setDisplayServerCommand(QStringLiteral());
-	}
+            m_auth->setDisplayServerCommand(m_displayServer->getCommand(this));
+        }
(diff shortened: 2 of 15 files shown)


src/daemon/Display.cpp
```diff
@@ -116,17 +118,27 @@ namespace SDDM {
         // Create display server
         switch (m_displayServerType) {
         case X11DisplayServerType:
+#ifdef BUILD_WITH_X11
             if (seat()->canTTY()) {
                 m_terminalId = VirtualTerminal::setUpNewVt();
             }
             m_displayServer = new XorgDisplayServer(this);
+#else
+            qCritical() << "Build without X11 support";
```
Vogtinator:
```suggestion
            qCritical() << "Built without X11 support";
```

src/daemon/Display.cpp
```diff
@@ -116,17 +118,27 @@ namespace SDDM {
...
+            QCoreApplication::exit(24);
```
Vogtinator: IMO this should be unreachable by overriding the type instead of exiting.

src/greeter/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ message(STATUS "Building greeter for Qt ${QT_MAJOR_VERSION} as ${GREETER_TARGET}
 include_directories(
     "${CMAKE_SOURCE_DIR}/src/common"
     "${CMAKE_BINARY_DIR}/src/common"
-    "${LIBXCB_INCLUDE_DIR}"
+    ${LIBXCB_INCLUDE_DIR}
```
Vogtinator, line 15: Is this correct? If it can't include multiple dirs it should probably remain quoted

src/greeter/KeyboardModel.cpp
```diff
@@ -32,9 +34,14 @@ namespace SDDM {
 
     KeyboardModel::KeyboardModel() : d(new KeyboardModelPrivate) {
         if (QGuiApplication::platformName() == QLatin1String("xcb")) {
+#ifdef BUILD_WITH_X11
             m_backend = new XcbKeyboardBackend(d);
             m_backend->init();
             m_backend->connectEventsDispatcher(this);
+#else
+            qCritical("Xcb backend is not supported in this build");
+            std::terminate();
```
Vogtinator: Instead of terminating, it should warn and continue anyway.

src/helper/CMakeLists.txt
```diff
@@ -34,7 +37,8 @@ target_link_libraries(sddm-helper
                       Qt${QT_MAJOR_VERSION}::Network
                       Qt${QT_MAJOR_VERSION}::DBus
                       Qt${QT_MAJOR_VERSION}::Qml
-                      ${LIBXAU_LINK_LIBRARIES})
+                      ${LIBXAU_LINK_LIBRARIES}
+                      ${JOURNALD_LIBRARIES})
```
Vogtinator, line 41: Unconditional?

Vogtinator (changes requested):

Vogtinator:
I'm not sure how valuable it is to avoid libxau and libxcb considering they will stay for the foreseeable future for Xwayland anyway.

src/greeter/CMakeLists.txt
LittleHuba: The empty string fails when building with `BUILD_WITH_X11=OFF`. The variable itself should be either a string or a list. In both cases, it requires no quoting. But I'm not too deep into CMake...

src/helper/CMakeLists.txt
LittleHuba: As far as I know, by not doing the `find_package` those variables should be empty. Hence, including them unconditionally would have no effect.

src/daemon/Display.cpp
LittleHuba: Do I understand you right? You would like to forcefully set `m_displayServerType` to `WaylandDisplayServerType`?
That way I could drop the else-path.

src/daemon/Display.cpp
LittleHuba: Fixed

src/greeter/KeyboardModel.cpp
LittleHuba: Fixed.

LittleHuba:
> I'm not sure how valuable it is to avoid libxau and libxcb considering they will stay for the foreseeable future for Xwayland anyway.

Yes, I agree that those libraries will stay around for some time. But to get rid of them, people must stop using them. So with this, it is one less usage and a little bit more pressure on other applications to do the same.

src/daemon/Display.cpp
Vogtinator: Yes. That way users with unsuitable config still get a working sddm.

LittleHuba: Fixed. Issuing a warning now, modifying `m_displayServerType` to `WaylandDisplayServerType`, and falling through to initialization.

CMakeLists.txt
```diff
@@ -168,6 +173,11 @@ else()
     set(RUNTIME_DIR_DEFAULT "${CMAKE_INSTALL_FULL_LOCALSTATEDIR}/run/sddm")
 endif()
 
+# x11
+if(BUILD_WITH_X11)
+    add_definitions(-DBUILD_WITH_X11)
```
Vogtinator: IMO definitions should always be 0 or 1, to avoid that a typo breaks stuff silently.

src/daemon/Display.cpp
```diff
@@ -113,21 +115,30 @@ namespace SDDM {
...
         // Create display server
         switch (m_displayServerType) {
         case X11DisplayServerType:
+#ifdef BUILD_WITH_X11
             if (seat()->canTTY()) {
                 m_terminalId = VirtualTerminal::setUpNewVt();
             }
             m_displayServer = new XorgDisplayServer(this);
             break;
+#endif
```
Vogtinator: IMO this should be merged for easier readability:

```
#if BUILD_WITH_X11
        case X11DisplayServerType:
            if (seat()->canTTY()) {
                m_terminalId = VirtualTerminal::setUpNewVt();
            }
            m_displayServer = new XorgDisplayServer(this);
            break;
        case X11UserDisplayServerType:
        ....
#else
        case X11DisplayServerType:
        case X11UserDisplayServerType:
        ...
        [[fallthrough]];
#endif
        case WaylandDisplayServerType:
```

CMakeLists.txt
LittleHuba: Fixed.

src/daemon/Display.cpp
LittleHuba: Fixed.

src/daemon/Display.cpp
```diff
@@ -24,8 +24,10 @@
 #include "Configuration.h"
 #include "DaemonApp.h"
 #include "DisplayManager.h"
+#if BUILD_WITH_X11 == 1
```
Vogtinator, line 27:
```suggestion
#if BUILD_WITH_X11
```

src/daemon/Display.cpp
```diff
@@ -126,8 +130,16 @@ namespace SDDM {
                 m_terminalId = fetchAvailableVt();
             }
             m_displayServer = new XorgUserDisplayServer(this);
-            m_greeter->setDisplayServerCommand(XorgUserDisplayServer::command(this));
+            m_greeter->setDisplayServerCommand(m_displayServer->getCommand(this));
             break;
+#else
+		case X11DisplayServerType:
```
Vogtinator: Indentation looks wrong

src/daemon/Display.cpp
```diff
@@ -126,8 +130,16 @@ namespace SDDM {
...
             }
             m_displayServer = new XorgUserDisplayServer(this);
-            m_greeter->setDisplayServerCommand(XorgUserDisplayServer::command(this));
+            m_greeter->setDisplayServerCommand(m_displayServer->getCommand(this));
             break;
+#else
+        case X11DisplayServerType:
+        case X11UserDisplayServerType:
+			// Reset type to Wayland and fall through to initialization
+			qWarning("Built without X11 support. Trying Wayland display server.");
```
Vogtinator:
```suggestion
			qWarning() << "Built without X11 support. Trying Wayland display server.";
```

Vogtinator (commented):
Looks ok from a quick view, needs a closer look at some point later

LittleHuba:
Fixed some formatting issues. All your comments should be fixed.

src/daemon/Display.cpp
LittleHuba: Also fixed now.

LittleHuba:
Missed one. Also fixed now.

Conan-Kudo:
Could we split out the automatic failover stuff to a separate pull request? That's useful even in the current setup.

PR #1904, opened 2024-04-04
Use xrdb to set Xcursor.theme

### LINKED ISSUE #1894: SDDM ignores the cursor theme and uses Adwaita
Desktop - KDE Plasma
OS - Arch
SDDM 0.21

My config file `/etc/sddm.conf.d/kde_settings.conf`

```
[Autologin]
Relogin=false
Session=
User=

[General]
HaltCommand=/usr/bin/systemctl poweroff
RebootCommand=/usr/bin/systemctl reboot

[Theme]
Current=breeze
CursorSize=24
CursorTheme=breeze_cursors
Font=Noto Sans,10,-1,0,400,0,0,0,0,0,0,0,0,0,0,1

[Users]
MaximumUid=60513
MinimumUid=1000

[X11]
ServerArguments=-dpi 0
```
cursor

Vogtinator:
XCURSOR_THEME is no longer enough.

Fixes #1894

### DIFF
--- a/src/daemon/XorgDisplayServer.cpp
+++ b/src/daemon/XorgDisplayServer.cpp
@@ -261,6 +261,9 @@ namespace SDDM {
         // create display setup script process
         QProcess *displayScript = new QProcess();
 
+        const QString xcursorTheme = mainConfig.Theme.CursorTheme.get(),
+                      xcursorSize = mainConfig.Theme.CursorSize.get();
+
         // set process environment
         QProcessEnvironment env;
         env.insert(QStringLiteral("DISPLAY"), m_display);
@@ -268,8 +271,8 @@ namespace SDDM {
         env.insert(QStringLiteral("PATH"), mainConfig.Users.DefaultPath.get());
         env.insert(QStringLiteral("XAUTHORITY"), m_xauth.authPath());
         env.insert(QStringLiteral("SHELL"), QStringLiteral("/bin/sh"));
-        env.insert(QStringLiteral("XCURSOR_THEME"), mainConfig.Theme.CursorTheme.get());
-        QString xcursorSize = mainConfig.Theme.CursorSize.get();
+        if (!xcursorTheme.isEmpty())
+            env.insert(QStringLiteral("XCURSOR_THEME"), xcursorTheme);
         if (!xcursorSize.isEmpty())
             env.insert(QStringLiteral("XCURSOR_SIZE"), xcursorSize);
         setCursor->setProcessEnvironment(env);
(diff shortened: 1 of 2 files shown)


Firestar-Reimu:
How is this PR's situation now?

Vogtinator:
Waiting for someone to review it so it can be merged.

@aleixpol @davidedmundson

Vogtinator:
> Waiting for someone to review it so it can be merged.
> 
> @aleixpol @davidedmundson

Ping.

src/daemon/XorgDisplayServer.cpp
```diff
@@ -261,15 +261,18 @@ namespace SDDM {
...
+
         // set process environment
         QProcessEnvironment env;
         env.insert(QStringLiteral("DISPLAY"), m_display);
         env.insert(QStringLiteral("HOME"), QStringLiteral("/"));
         env.insert(QStringLiteral("PATH"), mainConfig.Users.DefaultPath.get());
         env.insert(QStringLiteral("XAUTHORITY"), m_xauth.authPath());
         env.insert(QStringLiteral("SHELL"), QStringLiteral("/bin/sh"));
-        env.insert(QStringLiteral("XCURSOR_THEME"), mainConfig.Theme.CursorTheme.get());
-        QString xcursorSize = mainConfig.Theme.CursorSize.get();
```
davidedmundson, line 272: Isn't this two solutions to the same problem?

src/daemon/XorgDisplayServer.cpp
```diff
@@ -287,6 +290,16 @@ namespace SDDM {
             setCursor->kill();
         }
 
+        // Unlike libXcursor, xcb-util-cursor no longer looks at XCURSOR_THEME. Set the resource.
+        if (!xcursorTheme.isEmpty()) {
+            QProcess xrdbProcess;
+            xrdbProcess.setProcessEnvironment(env);
+            xrdbProcess.start(QStringLiteral("xrdb"), QStringList{QStringLiteral("-nocpp"), QStringLiteral("-merge")});
+            xrdbProcess.write(QStringLiteral("Xcursor.theme: %1").arg(xcursorTheme).toUtf8());
```
davidedmundson: Xcursor.size too?

src/daemon/XorgDisplayServer.cpp
Vogtinator: IIRC, both are needed. The environment variable one for `xsetroot` and other users of libXcursor, and the resource for Qt applications and maybe other frameworks.

src/daemon/XorgDisplayServer.cpp
Vogtinator: IIRC Qt actually looks at `XCURSOR_SIZE` still, but for consistency this might be better. I'll add it.

src/daemon/XorgDisplayServer.cpp
davidedmundson: Whilst that's absolutely true on the desktop, we know what toolkit the greeter uses. 
Anything outside the greeter seems like the desktop's problem

src/daemon/XorgDisplayServer.cpp
davidedmundson: src/gui/platform/unix/qgenericunixthemes.cpp:    static const int xCursorSize = qEnvironmentVariableIntValue("XCURSOR_SIZE");

You are right, and it doesn't read XCursor.size, don't bother.

src/daemon/XorgDisplayServer.cpp
Vogtinator: I can try to remove the environment variables and call `xsetroot` after `xrdb`. Will need a bit of testing though.

src/daemon/XorgDisplayServer.cpp
Vogtinator: For consistency I added it anyway. With rootful X11, the X server state is somewhat kept until the DE, so if we leak some settings it should be consistent at least.

Vogtinator:
Qt 6.5 added code that exposes $XCURSOR_THEME and $XCURSOR_SIZE as platform theme hint, but IIRC it doesn't actually do anything.

davidedmundson (approved):

src/daemon/XorgDisplayServer.cpp
Vogtinator: Technically works fine.

Qt 6's own platform code (which isn't used anywhere for cursor stuff?) only looks at the environment variables though, so let's just keep it around for now? Also makes it less likely to break something...

Vogtinator:
In `src/daemon/Greeter.cpp` I found that it's actually possible for theme.conf to set cursor themes as well - that is completely ignored by this fix and would need a bit of a redesign to take into account. Ok for now to ignore?

Vogtinator:
I noticed that xorguserhelper uses its own `startProcess` function for starting processes which wasn't used for `xrdb`. I switched to that, but needs testing.

PR #1907, opened 2024-04-06
Default to Qt6 and always use versioned greeters

Conan-Kudo:
Distributors can handle creating an unversioned name if necessary.

### DIFF
--- a/src/common/ThemeMetadata.cpp
+++ b/src/common/ThemeMetadata.cpp
@@ -28,7 +28,7 @@ namespace SDDM {
         QString mainScript { QStringLiteral("Main.qml") };
         QString configFile;
         QString translationsDirectory { QStringLiteral(".") };
-        int qtVersion = 5;
+        int qtVersion = 6;
     };
 
     ThemeMetadata::ThemeMetadata(const QString &path, QObject *parent) : QObject(parent), d(new ThemeMetadataPrivate()) {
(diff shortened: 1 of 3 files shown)


Conan-Kudo:
This patch has been in used in Fedora for some time now, and I'm reasonably confident it's fine for upstream.

Vogtinator (changes requested):
Changing the default value of QtVersion is wrong, Qt 5 themes are Qt 5 themes.

If you don't want to build the Qt 5 greeter, just not building it should be fully sufficient.

Conan-Kudo:
> Changing the default value of QtVersion is wrong, Qt 5 themes are Qt 5 themes.
> 
> If you don't want to build the Qt 5 greeter, just not building it should be fully sufficient.

It is not. As you are very well aware, if the version is unset, it uses the default no matter what. And if the Qt5 greeter doesn't exist, it fails.

Vogtinator:
> > Changing the default value of QtVersion is wrong, Qt 5 themes are Qt 5 themes.
> > If you don't want to build the Qt 5 greeter, just not building it should be fully sufficient.
> 
> It is not. As you are very well aware, if the version is unset, it uses the default no matter what. And if the Qt5 greeter doesn't exist, it fails.

Which is exactly the desired and expected behaviour: If you don't build support for Qt 5 themes, you cannot use Qt 5 themes. All Qt6 themes **must** set QtVersion.

With your default change you introduce an incompatibility: Qt 6 themes developed on Fedora might only work there by accident if they forget to set QtVersion. Please don't do that.

PR #1910, opened 2024-04-08
Set HideShells to "/usr/sbin/nologin,/sbin/nologin" by default

Vogtinator:


C0rn3j:
I am not sure if that commit takes care of `/usr/bin/nologin` which is what Arch Linux uses.

Vogtinator:
> I am not sure if that commit takes care of `/usr/bin/nologin` which is what Arch Linux uses.

It doesn't. Is there `/bin/nologin` as well?

Maybe we should just hardcode a check for `.endswith("/nologin")`? Wouldn't be configurable, but :shrug:

C0rn3j:
/bin is symlinked to usr/bin, so yeah.

> Maybe we should just hardcode a check for .endswith("/nologin")? Wouldn't be configurable, but 🤷

I think that's a good solution. I don't think people would want to login to nologin users, and if they do they're probably doing something wrong.

Could also look what GDM does in regards to this.

PR #1922, opened 2024-04-30
Respond to SecureAttentionKey events from logind

n3rdopolis:
This depends on https://github.com/systemd/systemd/pull/29542 not merged yet, subject to change

The first commit fixes an issue where the greeter hangs if SwitchToGreeter double fires.
Something that if the systemd MR is merged will be easier for users to accidentally trigger, at least with logind, as it is possible to hit Ctrl+Alt+Shift+Esc or whatever the final sequence is, when the greeter is active

dbus-send --system --type=method_call --print-reply --dest=org.freedesktop.DisplayManager $XDG_SEAT_PATH org.freedesktop.DisplayManager.Seat.SwitchToGreeter
sleep 5
dbus-send --system --type=method_call --print-reply --dest=org.freedesktop.DisplayManager $XDG_SEAT_PATH org.freedesktop.DisplayManager.Seat.SwitchToGreeter

### DIFF
--- a/src/daemon/SeatManager.cpp
+++ b/src/daemon/SeatManager.cpp
@@ -27,6 +27,8 @@
 #include <QDBusPendingReply>
 #include <QDBusContext>
 
+#include <Login1Manager.h>
+#include <Login1Session.h>
 #include "LogindDBusTypes.h"
 
 namespace SDDM {
@@ -147,6 +149,24 @@ namespace SDDM {
         // check if seat exists
         if (!m_seats.contains(name))
             return;
+
+        // Switch to existing greeter session if available
+        if (Logind::isAvailable()) {
+            OrgFreedesktopLogin1ManagerInterface manager(Logind::serviceName(), Logind::managerPath(), QDBusConnection::systemBus());
+            auto reply = manager.ListSessions();
+            reply.waitForFinished();
+
+            const auto info = reply.value();
+            for(const SessionInfo &s : reply.value()) {
+                if (s.userName == QLatin1String("sddm")) {
+                    OrgFreedesktopLogin1SessionInterface session(Logind::serviceName(), s.sessionPath.path(), QDBusConnection::systemBus());
+                    if (session.service() == QLatin1String("sddm-greeter") && session.seat().name == name) {
+                        session.Activate();
+                        return;
+                    }
+                }
+            }
+        }
 
         // switch to greeter
         m_seats.value(name)->createDisplay(Display::defaultDisplayServerType());
--- a/src/daemon/SeatManager.cpp
+++ b/src/daemon/SeatManager.cpp
@@ -115,6 +115,7 @@ namespace SDDM {
             }
         });
 
+        QDBusConnection::systemBus().connect(Logind::serviceName(), Logind::managerPath(), Logind::managerIfaceName(), QStringLiteral("SecureAttentionKey"), this, SLOT(logindSecureAttentionKey(QString,QDBusObjectPath)));
         QDBusConnection::systemBus().connect(Logind::serviceName(), Logind::managerPath(), Logind::managerIfaceName(), QStringLiteral("SeatNew"), this, SLOT(logindSeatAdded(QString,QDBusObjectPath)));
         QDBusConnection::systemBus().connect(Logind::serviceName(), Logind::managerPath(), Logind::managerIfaceName(), QStringLiteral("SeatRemoved"), this, SLOT(logindSeatRemoved(QString,QDBusObjectPath)));
     }
@@ -170,6 +171,12 @@ namespace SDDM {
 
         // switch to greeter
         m_seats.value(name)->createDisplay(Display::defaultDisplayServerType());
+    }
+
+    void SDDM::SeatManager::logindSecureAttentionKey(const QString& name, const QDBusObjectPath& objectPath)
+    {
+        Q_UNUSED(objectPath);
+        daemonApp->seatManager()->switchToGreeter(name);
     }
 
     void SDDM::SeatManager::logindSeatAdded(const QString& name, const QDBusObjectPath& objectPath)
--- a/data/interfaces/org.freedesktop.login1.Manager.xml
+++ b/data/interfaces/org.freedesktop.login1.Manager.xml
@@ -185,6 +185,10 @@
    <arg type="s" direction="in"/>
    <arg type="h" direction="out"/>
   </method>
+  <signal name="SecureAttentionKey">
+   <arg type="s"/>
+   <arg type="o"/>
+  </signal>
   <signal name="SessionNew">
    <arg type="s"/>
    <arg type="o"/>
--- a/src/daemon/SeatManager.h
+++ b/src/daemon/SeatManager.h
@@ -43,6 +43,7 @@ namespace SDDM {
         void seatRemoved(const QString &name);
 
     private Q_SLOTS:
+        void logindSecureAttentionKey(const QString &name, const QDBusObjectPath &objectPath);
         void logindSeatAdded(const QString &name, const QDBusObjectPath &objectPath);
         void logindSeatRemoved(const QString &name, const QDBusObjectPath &objectPath);
 


src/daemon/SeatManager.cpp
```diff
@@ -152,6 +153,13 @@ namespace SDDM {
         m_seats.value(name)->createDisplay(Display::defaultDisplayServerType());
     }
 
+    void SDDM::SeatManager::logindSecureAttentionKey(const QString& name, const QDBusObjectPath& objectPath)
+    {
+        Q_UNUSED(objectPath);
+        auto logindSeat = m_systemSeats.take(name);
```
Vogtinator: `take` removes it from the list, is that intentional? I don't think the key implies that the seat got removed.

src/daemon/Seat.cpp
```diff
@@ -47,6 +47,23 @@ namespace SDDM {
     }
 
     void Seat::createDisplay(Display::DisplayServerType serverType) {
+        if (Logind::isAvailable()) {
```
Vogtinator: Should IMO be done internally in `SeatManager::switchToGreeter`, iterate all displays and check whether it's a greeter.

Vogtinator (changes requested):

Vogtinator:
> The first commit fixes an issue where the greeter hangs if SwitchToGreeter double fires.

Why does it hang? Technically it should work to have multiple greeters on the same seat simultaneously, even though it does not make sense.

src/daemon/SeatManager.cpp
n3rdopolis: Woah, that is not intentional. Will fix

Thanks

n3rdopolis:
> > The first commit fixes an issue where the greeter hangs if SwitchToGreeter double fires.
> 
> Why does it hang? Technically it should work to have multiple greeters on the same seat simultaneously, even though it does not make sense.

If it helps, it's when I am using xorg-user as the greeter. I have not tested on my multiseat vm yet so it's only seat0 that I saw this

I am not sure why it hangs, it only hangs if you try to activate the greeter _again_ when the greeter is active. 
If the greeter is _running_, but not the active session it's OK

It seems to be something to do with the xserver itself or something? Because it's seems to be the display server that acts up, it doesn't respond to Ctrl+Alt+Fx either...

Vogtinator:
> > > The first commit fixes an issue where the greeter hangs if SwitchToGreeter double fires.
> > 
> > 
> > Why does it hang? Technically it should work to have multiple greeters on the same seat simultaneously, even though it does not make sense.
> 
> If it helps, it's when I am using xorg-user as the greeter. I have not tested on my multiseat vm yet so it's only seat0 that I saw this
> 
> I am not sure why it hangs, it only hangs if you try to activate the greeter _again_ when the greeter is active. If the greeter is _running_, but not the active session it's OK
> 
> It seems to be something to do with the xserver itself or something? Because it's seems to be the display server that acts up, it doesn't respond to Ctrl+Alt+Fx either...

Maybe a deadlock during VT switch?

n3rdopolis:
> > > > The first commit fixes an issue where the greeter hangs if SwitchToGreeter double fires.
> > > 
> > > 
> > > Why does it hang? Technically it should work to have multiple greeters on the same seat simultaneously, even though it does not make sense.
> > 
> > 
> > If it helps, it's when I am using xorg-user as the greeter. I have not tested on my multiseat vm yet so it's only seat0 that I saw this
> > I am not sure why it hangs, it only hangs if you try to activate the greeter _again_ when the greeter is active. If the greeter is _running_, but not the active session it's OK
> > It seems to be something to do with the xserver itself or something? Because it's seems to be the display server that acts up, it doesn't respond to Ctrl+Alt+Fx either...
> 
> Maybe a deadlock during VT switch?

Could be, or it could be it trying to restart the greeter/server incompletely or something

If I try to vt switch away from the greeter vt and back to it, it still seems to deadlock (This is with QEMU's virtual serial console that I am able to run the chvt commands)

It could also be an older xserver.... or just this VM

Are you able to replicate with x11-user and these sets of commands

```
dbus-send --system --type=method_call --print-reply --dest=org.freedesktop.DisplayManager $XDG_SEAT_PATH org.freedesktop.DisplayManager.Seat.SwitchToGreeter
sleep 5
dbus-send --system --type=method_call --print-reply --dest=org.freedesktop.DisplayManager $XDG_SEAT_PATH org.freedesktop.DisplayManager.Seat.SwitchToGreeter
```
by chance?

n3rdopolis:
Testing on a clean Kubuntu 24.04 ISO, with that set of commands to double-fire SwitchToGreeter, and well it seems that instead of hanging, it is starting the greeter twice, first on tty1, then on tty3,

It defaults to running Xorg as root, and not x11-user, it seems like that this is the issue, it is trying another instance, it just was starting multiple greeters before, and maybe x11-user is different, and is stepping on the first instance, maybe that explains the hang

n3rdopolis:
https://github.com/systemd/systemd/pull/29542 is now merged

src/daemon/SeatManager.cpp
```diff
@@ -148,10 +151,33 @@ namespace SDDM {
         if (!m_seats.contains(name))
             return;
 
+        if (Logind::isAvailable()) {
```
Vogtinator, line 155: Add a comment, like

```
// Switch to existing greeter session if available
```

Vogtinator:
I wonder whether this should be handled by `Seat.cpp` instead, it could iterate all of `m_display` and check whether it is an available greeter. Might need some refactoring though...

Vogtinator:
In the case of using x11user or wayland as display server, the greeter is handled separately from the session and is only stopped 5s after successful login. I imagine that in some cases (e.g. apparent hang) the SAK may get pressed in between and it might switch to an effectively zombie greeter.

n3rdopolis:
What would need to be moved? I did a lot of this by grepping though, so I am not 100% familiar...

Vogtinator:
> What would need to be moved? I did a lot of this by grepping though, so I am not 100% familiar...

Not sure. The main issue I see that might happen in practice is https://github.com/sddm/sddm/blob/ae072f901671b68861da9577e3e12e350a9053d5/src/daemon/Display.cpp#L557

Would be great if you could confirm that this is the case.

Doing the switch internally instead of through logind makes it independent of logind and also more reliable IMO.

n3rdopolis:
Just to make sure I understand,, this is that that 5 second timer that quits the greeter might cause the greeter to quit, but the display server to still be running, correct?

Vogtinator:
> Just to make sure I understand,, this is that that 5 second timer that quits the greeter might cause the greeter to quit, but the display server to still be running, correct?

I think so, but I'm not fully sure.

n3rdopolis:
OK, I can try to test that. If you want, I also have a diff, and backported SecureAttentionKey against logind v252 in Debian Stable, and I can make a .deb with it, that does not interfere with installed files (installs under /opt) from the main systemd package if you would also like to test, and don't want to deal with getting a system with v257

n3rdopolis:
@Vogtinator I am trying to replicate what you want me to replicate, I have x11-user enabled, I log in, and the greeter and the display server both quit

I try the Ctrl+Alt+Shift+Esc with the patched logind, and I can't seem to get a dormant display server with no greeter running, unless I am following the directions wrong

Vogtinator:
Maybe related to #1960 the greeter quits internally quickly enough that the 5s delay doesn't even trigger - but why does the code exist then? :confused:

n3rdopolis:
Is there anything you want me to try?

Vogtinator:
I'd like to know how exactly the greeter session (sddm-greeter + possibly display server) lifetime is handled first, to make sure that there's no weird case lingering around. Currently it's not that obvious.

n3rdopolis:
OK, I can try, but I am not sure if I am creative enough to figure out all the hypotheticals

I do see that it seems that at least in xorg-user, if I `pkill sddm-greeter` the display server also quits

n3rdopolis:
Any experiments you need me to try?

n3rdopolis:
I am sorry, I am not sure what needs to be done from here. I do know that killing the greeter seems to kill the xorg-user server

Vogtinator:
I'm not sure either. I only have hypothetical issues and don't see anything obviously wrong with the code, so let's just be optimistic and fix it later if necessary.

Vogtinator (approved):

n3rdopolis:
Thanks!

PR #1896, opened 2024-03-17
Always use VT_OPENQRY

Vogtinator:
There is no other way to reliably grab a VT for use.

\+ bonus tty fix.

Draft because not tested yet.

PR #1932, opened 2024-05-16
Display: add plymouth integration

snowkat:
This provides plymouth smooth transition support in SDDM::Display by telling Plymouth to deactivate and quit as the greeter is brought online.

This is similar in goal to #1013, although following the suggestion in https://github.com/sddm/sddm/pull/1013#issuecomment-1493269798, the plan is to handle the transition from within the sddm daemon.

I wouldn't consider this PR complete, primarily due to the following lines in src/daemon/Display.cpp:

```cpp
        case X11DisplayServerType:
            if (seat()->canTTY()) {
#if HAVE_PLYMOUTH
                // TODO: why can't rootful X11 use fetchAvailableVt?
                if (m_plymouthIsRunning) {
                    quitPlymouth(false);
                    m_plymouthIsRunning = false;
                }
#endif
                m_terminalId = VirtualTerminal::setUpNewVt();
            }
```

More specifically, the `x11` display server type always creates a new VT, whereas `x11-user` and `wayland` will use `fetchAvailableVt()`. Is there a requirement for this to be the case?

My understanding of Plymouth is that to support a smooth transition, the DM needs to load on the active VT (i.e., same as the splash screen), so if there is a requirement, `x11` display server type won't be able to support smooth transition support with this method.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,7 @@ option(ENABLE_JOURNALD "Enable logging to journald" ON)
 option(NO_SYSTEMD "Disable systemd support" OFF)
 option(USE_ELOGIND "Use elogind instead of logind" OFF)
 option(BUILD_WITH_QT6 "Build with Qt 6" OFF)
+option(WITH_PLYMOUTH "Integrate with plymouth at startup" ON)
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
@@ -168,6 +169,15 @@ else()
     set(RUNTIME_DIR_DEFAULT "${CMAKE_INSTALL_FULL_LOCALSTATEDIR}/run/sddm")
 endif()
 
+# plymouth
+if (WITH_PLYMOUTH)
+    pkg_check_modules(PLYMOUTH "ply-boot-client")
+endif()
+
+if (PLYMOUTH_FOUND)
+    add_definitions(-DHAVE_PLYMOUTH)
+endif()
+add_feature_info("plymouth" PLYMOUTH_FOUND "plymouth smooth transition support")
 
 # Set constants
 set(DATA_INSTALL_DIR            "${CMAKE_INSTALL_FULL_DATADIR}/sddm"                CACHE PATH      "System application data install directory")
--- a/services/sddm.service.in
+++ b/services/sddm.service.in
@@ -2,7 +2,7 @@
 Description=Simple Desktop Display Manager
 Documentation=man:sddm(1) man:sddm.conf(5)
 Conflicts=getty@tty${SDDM_INITIAL_VT}.service
-After=systemd-user-sessions.service getty@tty${SDDM_INITIAL_VT}.service plymouth-quit.service systemd-logind.service
+After=systemd-user-sessions.service getty@tty${SDDM_INITIAL_VT}.service systemd-logind.service
 PartOf=graphical.target
 StartLimitIntervalSec=30
 StartLimitBurst=2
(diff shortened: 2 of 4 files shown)


Vogtinator:
> More specifically, the x11 display server type always creates a new VT, whereas x11-user and wayland will use fetchAvailableVt(). Is there a requirement for this to be the case?

No. The VT handling code is a mess and needs to be redone, see #1896.

> My understanding of Plymouth is that to support a smooth transition, the DM needs to load on the active VT (i.e., same as the splash screen), so if there is a requirement, x11 display server type won't be able to support smooth transition support with this method.

AFAICT it's the opposite: The greeter needs to start on another VT, to be able to switch in an instant.

snowkat:
> AFAICT it's the opposite: The greeter needs to start on another VT, to be able to switch in an instant.

That would still cause a flicker, wouldn't it? Admittedly, the best documentation I've found is a 15-year-old mailing list post (https://lists.freedesktop.org/archives/plymouth/2009-June/000126.html), but it describes:

> The way it works is instead of running "plymouth quit" we run "plymouth quit --retain-splash" which forces plymouth to keep the splash contents on screen and keep the vt in KD_GRAPHICS mode after plymouth exits.  When in this mode the user won't be able to switch to other vts.
>
> We then have a patch fo GDM which forces X to start on the active vt.

From what I can tell of the Plymouth and GDM sources, this is still the case.

Vogtinator:
> > AFAICT it's the opposite: The greeter needs to start on another VT, to be able to switch in an instant.
> 
> That would still cause a flicker, wouldn't it?

If the mode is the same, it should not. I don't get any flicker here using manual VT switching.

> Admittedly, the best documentation I've found is a 15-year-old mailing list post (https://lists.freedesktop.org/archives/plymouth/2009-June/000126.html), but it describes:
> 
> > The way it works is instead of running "plymouth quit" we run "plymouth quit --retain-splash" which forces plymouth to keep the splash contents on screen and keep the vt in KD_GRAPHICS mode after plymouth exits.  When in this mode the user won't be able to switch to other vts.
> > We then have a patch fo GDM which forces X to start on the active vt.
> 
> From what I can tell of the Plymouth and GDM sources, this is still the case.

Ouch. Can we try a better way?

snowkat:
> > That would still cause a flicker, wouldn't it?
> 
> If the mode is the same, it should not. I don't get any flicker here using manual VT switching.

Heh, it would have been a good idea for me to test before saying that. Can confirm there's no flicker on my system, either.

> Ouch. Can we try a better way?

Depends on your definition of a "smooth transition." If we're fine with not supporting the fade-in effect done by GDM and LightDM, the VT switch would be fine.

Arguably it'd be possible to have that effect even when switching the VT, but you'd already need the active VT open to get a "screenshot" of the splash.

Vogtinator:
> Arguably it'd be possible to have that effect even when switching the VT, but you'd already need the active VT open to get a "screenshot" of the splash.

Plymouth could fade out, then after the VT switch sddm or the DE could fade in.

snowkat:
> Plymouth could fade out, then after the VT switch sddm or the DE could fade in.

AFAICT Plymouth doesn't have a fade-out mechanism of its own. The DM seems to be expected to handle all parts of the transition, including animation.

Technically the `two-step` splash type could do this, but it's more meant to add a little flair when loading is complete, and none of the default themes using `two-step` use it for a fade-out effect.

Zamundaaa:
> Ouch. Can we try a better way?

With https://gitlab.freedesktop.org/plymouth/plymouth/-/merge_requests/313 the display server could read back the framebuffer and do a smooth transition that way

PR #1951, opened 2024-06-13
Translations: fix bad translation for password change to German

libewa:


data/translations/de.ts
```diff
@@ -107,11 +107,11 @@
     </message>
     <message>
         <source>Password change aborted because maximum tries reached</source>
-        <translation>Wegen zu vieler Versuche, Passwort wechsel abgebrochen</translation>
+        <translation>Zu viele Fehlversuche, Passwortänderung abgebrochen</translation>
     </message>
     <message>
         <source>New password change round! Please input current password again!</source>
-        <translation>Bitte Passwort Eingabe wiederholen</translation>
+        <translation>Bitte neues Passwort erneut eingeben.</translation>
```
Vogtinator, line 114: It says `current password`, so:

```suggestion
        <translation>Bitte aktuelles Passwort erneut eingeben.</translation>
```

libewa: I haven't looked at the sddm source, so correct me if I'm wrong, but this seems to be the "typo check" prompt when changing your password, so the translation would actually be better, right?

Vogtinator: Actually, I don't think that string is actually used anywhere... I don't see any reference to `pamMaxtriesInfo` in the source.

PR #1961, opened 2024-07-08
Move X11 display setup into separate function

Vogtinator:
Alternate version of #1904 which has some refactoring ~~to let themes specify the cursor theme again.~~

Draft because:

- [ ] Only tested in daemon test mode.
- [x] ~It uses QProcess instead of xorguserhelper's custom `startProcess`. Not sure whether that's needed though.~ Not needed, it even breaks `xrdb` because of stdin forwarding.

### DIFF
--- a/src/common/XSetup.cpp
+++ b/src/common/XSetup.cpp
@@ -1,0 +1,81 @@
+/***************************************************************************
+* Copyright (c) 2024 Fabian Vogt <fabian@ritter-vogt.de>
+*
+* This program is free software; you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation; either version 2 of the License, or
+* (at your option) any later version.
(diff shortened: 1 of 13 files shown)


Vogtinator:
I noticed that there's another issue with the ability to set the cursor theme from the sddm theme's config: In the case of rootful X, the end result of the `Xcursor` resources depends on whether a greeter was started or not, i.e. whether autologin was used. If not, the theme config isn't used so it's never set to the theme's config value.

That wasn't an issue before #1904 as it was only set for the greeter's process, but that's no longer possible. To restore that behavior, it needs to be restored or unset explicitly before the user session is started in the same rootful X.

For now it's probably ok to keep the status quo, which is to just ignore the theme config's cursor theme with rootful X.

That makes this PR just a refactor which somehow ended up adding more code than it removes :-/

PR #1962, opened 2024-07-09
Introduce utmps support

petershh:
utmps library can be used to provide utmp functionality on systems whose libcs do not implement it, e.g. musl.

For more details, see https://skarnet.org/software/utmps/.


Vogtinator:
What's the source of `Findutmps.cmake`?

petershh:
> What's the source of `Findutmps.cmake`?

I have written this file by myself, since utmps doesn't provide any CMake modules or pkg-config files.

cmake/Findutmps.cmake
```diff
@@ -0,0 +1,38 @@
...
+# UTMPS_FOUND - system has utmps and dependencies
+# UTMPS_INCLUDE_DIR - the utmps include dir
+# UTMPS_LIBRARIES - libutmps, libskarnet library and sysdeps libraries
+
+if (UTMPS_INCLUDE_DIR AND UTMPS_LIBRARIES)
+    # Already cached
+    set(UTMPS_FIND_SILENTLY TRUE)
+endif (UTMPS_INCLUDE_DIR AND UTMPS_LIBRARIES)
+
+find_path(UTMPS_INCLUDE_DIR NAMES utmps/utmpx.h)
```
Vogtinator, line 7: I think only

```
find_library(UTMPS_LIBRARY utmps)
find_path(UTMPS_INCLUDE_DIR NAMES utmps/utmpx.h)
```

are actually necessary. The include doesn't have any dependencies and as long as the library is linked dynamically it's enough to mention just that.

CMakeLists.txt
```diff
@@ -168,6 +169,10 @@ else()
     set(RUNTIME_DIR_DEFAULT "${CMAKE_INSTALL_FULL_LOCALSTATEDIR}/run/sddm")
 endif()
 
+# utmps
+if (USE_UTMPS)
```
Vogtinator: IMO this should be set as optional and get used automatically if found.

petershh: Do I understand correctly that utmps should be used if found even if user didn't request it?

Vogtinator: If it's available, it's on the user's system and it should probably be used. Do you see any reason not to?

petershh: I concur. I've just asked to make sure I understand you correctly.

cmake/Findutmps.cmake
petershh: I've changed this part too. Or should I remove UTMPS_FIND_SILENTLY too? I've looked at FindPAM.cmake while writing this so I've borrowed this part.

Vogtinator: IMO everything except lines 12-18 can be removed, though I'm not familiar enough with cmake whether there are some edge cases or expectations that need to be met by find modules.

src/helper/CMakeLists.txt
```diff
@@ -53,6 +53,11 @@ else()
     target_link_libraries(sddm-helper crypt)
 endif()
 
+if (UTMPS_FOUND)
+    target_include_directories(sddm-helper SYSTEM PUBLIC ${UTMPS_INCLUDE_DIR})
```
Vogtinator: Not used in public headers (which sddm-helper doesn't have any of anyway), so:

```suggestion
    target_include_directories(sddm-helper SYSTEM PRIVATE ${UTMPS_INCLUDE_DIR})
```

cmake/Findutmps.cmake
petershh: Should I move what have remained in the primary CMakeLists.txt ? Separate Find file looks cleaner to me though.

Vogtinator: Separate find module file is fine IMO

Vogtinator (approved):

Issue #1803, opened 2023-09-27
Removing a Display causes a VT switch


arrowd:
I bumped into a following problem:

1. SDDM starts and creates the first Display (vt9 in my case).
2. Greeter runs on this Display.
3. User logs in and the session replaces the greeter. The Display stays the same.
4. User performs "Switch user" action.
5. New Display is created (vt10) and Greeter is shown there.
6. Log into different user. Like in a step 3, Greeter exits and a user session occupies the Display.
7. We now have 2 Displays both running a user session.
8. User performs "Switch user" action again.
9. The third Display is created and Greeter runs in it.
10. Login into the first user again.

Step 10 triggers the "reuse" code path and causes SDDM to switch into vt9. After a small delay SDDM shuts down itself. The `removeDisplay(display)` call in `Seat::displayStopped()` causes the third Display to be destroyed and the third `Xorg` process to exit. When it exits the VT switch happens which always brings me to the **last occupied VT**. Effectively, after a few seconds of being in the user session on vt9 I get switched to the other session (vt10) which is in a locked state.

That Xorg shutdown behavior might be FreeBSD specific, so I'm not sure if it is visible on Linux.

Vogtinator:
IIRC I hit this as well while trying to improve session switching. There are a LOT of race conditions.

Back then I tried to start X with `-novtswitch` to make this more explicit and controllable, but it only ended up breaking stuff even more.

arrowd:
Do you have an idea how to move this forward? I want to fix this, so any input would be appreciated.

> Back then I tried to start X with -novtswitch to make this more explicit and controllable, but it only ended up breaking stuff even more.

How exactly did it break? Does it also disable VT switch on Xorg start?

arrowd:
What if we don't remove a Display if we switch into an existing session? This is suboptimal, as it would cause SDDM greeter to work in the background, but at least would get us rid of switching problems.

Vogtinator:
> Do you have an idea how to move this forward? I want to fix this, so any input would be appreciated.
> 
> > Back then I tried to start X with -novtswitch to make this more explicit and controllable, but it only ended up breaking stuff even more.
> 
> How exactly did it break? Does it also disable VT switch on Xorg start?

I don't think so, but this actually shouldn't be much of an issue as SDDM handles VT switching already.

> What if we don't remove a Display if we switch into an existing session? This is suboptimal, as it would cause SDDM greeter to work in the background, but at least would get us rid of switching problems.

[...]

PR #1968, opened 2024-07-16
Stop rootful Xorg after starting a Wayland session

Vogtinator:
With rootful Xorg, the display manager is kept running and used for the user
X session. If a Wayland session is started though keeping Xorg running just
wastes resources (VT, RAM, GPU) and looks weird (black screen + cursor).

To avoid that stopping Xorg causes a VT switch, https://github.com/sddm/sddm/issues/1803 needs to be fixed by adding `-novtswitch`.

Draft because:
- [x] Test logging into X11 and Wayland sessions
- [ ] Need to figure out what to do on SDDM exit. Switch to whatever VT was active during start or hardcode VT 1?
- [ ] Implement and test ^

CC @arrowd

### DIFF
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -555,6 +555,12 @@ namespace SDDM {
         qDebug() << "Session started" << success;
         if (success) {
             QTimer::singleShot(5000, m_greeter, &Greeter::stop);
+
+            // If no longer needed, stop the display server.
+            if(m_sessionTerminalId != m_terminalId) {
+                disconnect(m_displayServer, &DisplayServer::stopped, this, &Display::stop);
+                QTimer::singleShot(5000, m_displayServer, &DisplayServer::stop);
+            }
         }
     }
 }
--- a/src/daemon/XorgDisplayServer.cpp
+++ b/src/daemon/XorgDisplayServer.cpp
@@ -24,6 +24,7 @@
 #include "DaemonApp.h"
 #include "Display.h"
 #include "Seat.h"
+#include "VirtualTerminal.h"
 
 #include <QDebug>
 #include <QFile>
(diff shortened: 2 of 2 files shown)


arrowd:
We're running sddm with `-novtswitch` change for a year now and it cause no problems for us.

Regarding shutting down sddm after Wayland session started - what would happen if user clicks "Switch user" in Plasma? Currently it starts a new SDDM Greeter, would that still be the case?

Vogtinator:
> We're running sddm with `-novtswitch` change for a year now and it cause no problems for us.

Using `[X11] ServerArguments=`? If so, perfect - already some field testing.

> Regarding shutting down sddm after Wayland session started - what would happen if user clicks "Switch user" in Plasma? Currently it starts a new SDDM Greeter, would that still be the case?

Yep. Didn't test that case, but the code path touched here isn't actually involved.

Without this PR it ended up with three "Display"s on three VTs:
1. Display server originally for the greeter, no content
2. Wayland user session
3. Xorg with greeter inside

This gets rid of 1.

arrowd:
> Using [X11] ServerArguments=? If so, perfect - already some field testing.

Nope, just like you did in XorgDisplayServer.cpp

> This gets rid of 1.

Hum, in the X11-only case this 1 display gets reused for the user session. Did you check that this display is not then killed by your change?

> Switch to whatever VT was active during start or hardcode VT 1?

Hardcoding is certainly wrong, because at least on FreeBSD the X server defaults to VT 9.

Vogtinator:
> > Using [X11] ServerArguments=? If so, perfect - already some field testing.
> 
> Nope, just like you did in XorgDisplayServer.cpp

Ok.

> > This gets rid of 1.
> 
> Hum, in the X11-only case this 1 display gets reused for the user session. Did you check that this display is not then killed by your change?

Yes, of course.

> > Switch to whatever VT was active during start or hardcode VT 1?
> 
> Hardcoding is certainly wrong, because at least on FreeBSD the X server defaults to VT 9.

VT 9 for starting X? Or does X default to switching to VT 9 on exit?

arrowd:
> VT 9 for starting X? Or does X default to switching to VT 9 on exit?

The former. When exiting and if no `-novtswitch` passed, it just switches to the VT where it was started from, IIRC.

Issue #1844, opened 2023-12-30
SDDM on Wayland loops because it needs tty1 to start - but there's getty there.


NyaomiDEV:
On Artix Linux, tty1 to 6 are `getty` sessions. SDDM (Wayland + KWin) could start on tty7 which is free to use, but there's no configuration file to make it do so.

Also, it will keep trying to get VT 1 regardless! (Of course, `getty` sessions are still on `login` so logind doesn't know about them)

Relevant log (copied from `bat`):
```
12675   │ [12:11:28.113] (II) DAEMON: Loading theme configuration from "/usr/share/sddm/themes/sugar-candy/theme.conf"
12676   │ [12:11:28.114] (II) DAEMON: Greeter starting...
12677   │ [12:11:28.123] (II) HELPER: [PAM] Starting...
12678   │ [12:11:28.123] (II) HELPER: [PAM] Authenticating...
12679   │ [12:11:28.123] (II) HELPER: [PAM] returning.
12680   │ [12:11:28.154] (EE) HELPER: Failed to take control of "/dev/tty1" ("root"): Operation not permitted
12681   │ [12:11:28.154] (II) DAEMON: Greeter session started successfully
12682   │ [12:11:28.155] (II) HELPER: [PAM] Closing session
12683   │ [12:11:28.155] (II) HELPER: [PAM] Ended.
12684   │ [12:11:28.156] (WW) DAEMON: Auth: sddm-helper exited with 5
12685   │ [12:11:28.156] (II) DAEMON: Greeter stopped. SDDM::Auth::HELPER_TTY_ERROR
12686   │ [12:11:28.156] (II) DAEMON: Jumping to VT 1
12687   │ [12:11:28.185] (II) DAEMON: VT mode didn't need to be fixed
12688   │ [12:11:28.185] (II) DAEMON: Socket server stopping...
12689   │ [12:11:28.186] (II) DAEMON: Socket server stopped.
12690   │ [12:11:28.186] (II) DAEMON: Removing display SDDM::Display(0x55e813be6a10) ...
12691   │ [12:11:28.186] (II) DAEMON: Adding new display...
12692   │ [12:11:28.186] (II) DAEMON: Loaded empty theme configuration
12693   │ [12:11:28.186] (II) DAEMON: Using VT 1
12694   │ [12:11:28.186] (II) DAEMON: Display server started.
12695   │ [12:11:28.186] (II) DAEMON: Socket server starting...
[...]

PR #1969, opened 2024-07-16
If autologin is used, avoid starting a display server for the greeter

Vogtinator:
Introduce a new "early autologin" path that starts the session without
starting the display server first. This avoids starting a rootful X just
to then ignore it and start a Wayland session in a separate VT.

This should avoid the race condition causing tty opening failures with wayland autologin, which is actually separate from #1844. In the autologin case, SDDM first starts the display server, which switches to e.g. tty2, then starts the user session which then picks e.g. tty1 which is about to be used by getty.

Needed some refactoring of Display.cpp. It's split into various small commits, should be simple to read individually.

I think I tested all possible combinations:
- [x] `DisplayServer=x11` + autologin into `plasmawayland`
- [x] `DisplayServer=x11` + autologin into `plasma6` (x11)
- [x] `DisplayServer=x11` + failed autologin (invalid user) into `plasma6` (x11)
- [x] `DisplayServer=x11` + failed autologin (invalid user) into `plasmawayland`
- [x] `DisplayServer=x11-user` + autologin into `plasmawayland`
- [x] `DisplayServer=wayland` + autologin into `plasma6`

All of those cases behaved as expected.

### DIFF
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -175,9 +175,6 @@ namespace SDDM {
         // Load autologin configuration (whether to autologin, user, session, session type)
         if ((daemonApp->first || mainConfig.Autologin.Relogin.get()) &&
             !mainConfig.Autologin.User.get().isEmpty()) {
-            // reset first flag
-            daemonApp->first = false;
-
             // determine session type
             QString autologinSession = mainConfig.Autologin.Session.get();
             // not configured: try last successful logged in
@@ -192,6 +189,9 @@ namespace SDDM {
                 qCritical() << "Unable to find autologin session entry" << autologinSession;
             }
         }
+
+        // reset first flag
+        daemonApp->first = false;
     }
 
     Display::~Display() {
@@ -254,9 +254,6 @@ namespace SDDM {
 
         // start greeter
         m_greeter->start();
-
-        // reset first flag
-        daemonApp->first = false;
     }
 
     void Display::handleAutologinFailure() {
@@ -273,9 +270,6 @@ namespace SDDM {
         qDebug() << "Display server started.";
 
         if (m_autologinSession.isValid()) {
-            // reset first flag
-            daemonApp->first = false;
-
             m_auth->setAutologin(true);
             if (!startAuth(mainConfig.Autologin.User.get(), QString(), m_autologinSession))
                 handleAutologinFailure();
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
Does this affect when you do logout from an autologin session?

Vogtinator:
> Does this affect when you do logout from an autologin session?

Not meaningfully. With this PR it no longer has to stop the zombie X for the greeter anymore before starting the next greeter. (Same as with #1968 for the non-autologin case.)

Conan-Kudo (approved):

Conan-Kudo:
The code looks good to me, but this needs to be rebased to fix the CI.

Vogtinator:
Meh, now TW failed because https://build.opensuse.org/request/show/1191331 didn't get published yet.

buzzmandt:
What is the status on this please? I think this might fix my autologin landing in a tty1 issue.

Vogtinator:
Ready to merge, waiting for @aleixpol and/or @davidedmundson

davidedmundson:
If we switch the default to x11-user, which we should definitely do, do we not get all of this effectively for free without any new code?

Given we want to throw rootful mode away, we should avoid adding any other complexity unless there's a user-facing benefit.

Vogtinator:
> Given we want to throw rootful mode away, we should avoid adding any other complexity unless there's a user-facing benefit.

Yes, this fixes autologin into wayland sessions.

davidedmundson:
> Yes, this fixes autologin into wayland sessions.

Which bit? I can see we avoid calling displayServer->start() but that basically does nothing for x11-user or wayland setups.

Vogtinator:
> > Yes, this fixes autologin into wayland sessions.
> 
> Which bit? I can see we avoid calling displayServer->start() but that basically does nothing for x11-user or wayland setups.

Correct, in case SDDM is configured to use rootful X11 still which is still the default.

Unless you want to drop rootful X11 as a bugfix, we need to address this somehow.

Ideally we'd address the root cause of the VT allocation issues somehow, but I don't know how (see #1896)

davidedmundson:
>Unless you want to drop rootful X11 

I 1000% do. This code is already way too unmaintainable and it's unsafe for users. 

That said, I've done a code review. The other 4 cleanup commits are things we definitely want.

I'm at risk of spending longer discussing this than it would take to clean it up when the rootful X is dropped. 
If you think it's worth it, I'll press the button :)

>Ideally we'd address the root cause of the VT allocation issues somehow, but I don't know how (see https://github.com/sddm/sddm/pull/1896)

Ack.

davidedmundson (approved):

PR #1989, opened 2024-09-24
Don't quit when there are no screens

davidedmundson:
Default Qt behaviour is to quit when all windows are closed.

Windows can get a close event when the output is being removed, potentially slightly before we handle the output removed and delete the window.

This ordering means we can have no screens for enough time for this hook to take effect.

### DIFF
--- a/src/greeter/GreeterApp.cpp
+++ b/src/greeter/GreeterApp.cpp
@@ -362,6 +362,8 @@ int main(int argc, char **argv)
 
     QGuiApplication app(argc, argv);
     SDDM::SignalHandler s;
+
+    app.setQuitOnLastWindowClosed(false);
     QObject::connect(&s, &SDDM::SignalHandler::sigtermReceived, &app, [] {
         QCoreApplication::instance()->exit(-1);
     });


Vogtinator (approved):

Vogtinator:
I guess without this the greeter quit and sddm had to restart the display?

i.e. just annoying but no harm done?

davidedmundson:
>I guess without this the greeter quit and sddm had to restart the display?

Apparently not. https://bugs.kde.org/show_bug.cgi?id=479359

Vogtinator:
Hm, that's arguably a more severe bug that needs investigation. I'd expect that `Greeter stopped. SDDM::Auth::HELPER_SUCCESS` ends up in `Display::stopped` eventually which should realize that there's no display left.

davidedmundson:
We have status = HELPER_SUCCESS

```
   void Greeter::onHelperFinished(Auth::HelperExitStatus status) {
...
        if (status == Auth::HELPER_DISPLAYSERVER_ERROR) {
            Q_EMIT displayServerFailed();
        } else if (status == Auth::HELPER_TTY_ERROR) {
            Q_EMIT ttyFailed();
        } else if (status == Auth::HELPER_SESSION_ERROR) {
            Q_EMIT failed();
        }
    }
```
we no-op

In the normal case (for some reason) when we start login successfully we tell the greeter to quit after 5 seconds. 
(Display::slotSessionStarted)
For this reason we want a successful exit to do nothing.

We could make Greeter::stop set a flag and then check that in onHelperFinished

Vogtinator:
> In the normal case (for some reason) when we start login successfully we tell the greeter to quit after 5 seconds.
(Display::slotSessionStarted)
> For this reason we want a successful exit to do nothing.

Right, that can actually be an issue: The session must only start after the greeter quit successfully, otherwise you can technically have a window where the greeter and user session are running in the same rootful X display.

Maybe `slotSessionStarted` should be replaced by `onHelperFinished(success)`?

davidedmundson:
>otherwise you can technically have a window where the greeter and user session are running in the same rootful X display.

I think that's intentional to make it smooth?

Vogtinator:
> > otherwise you can technically have a window where the greeter and user session are running in the same rootful X display.
> 
> I think that's intentional to make it smooth?

Only in theory if the greeter would actually wait after successful authentication. Currently it just exits and the screen remains black until the session takes over.

pboling:
I don't know if it is related to this, but I have the dual screen Asus Zenbook running Fedora 42, aurora-dx-hwe and if I disable a screen and reboot I get a black screen on both screens after boot, and Wayland can't be started from run level 4, as it complains that it can't find $DISPLAY.

I haven't resolved this yet.

pboling:
It is dying with this coredump, and `sddm-helper-start-wayland` says something about an invalid pointer:

image

pboling:
`ujust update` fixed it for me!

a17r:
Same fix was committed in https://github.com/sddm/sddm/commit/da0b64d8e2378c4bce16cbea6cac2ffc89ad1a4a

PR #2004, opened 2024-10-25
xcb: use xcb_connection_has_error to check for failue

lbwtw:
xcb_connect() always returns a non-NULL pointer to a xcb_connection_t, even on failure.

### DIFF
--- a/src/greeter/XcbKeyboardBackend.cpp
+++ b/src/greeter/XcbKeyboardBackend.cpp
@@ -80,7 +80,7 @@ namespace SDDM {
         xcb_generic_error_t *error = nullptr;
 
         m_conn = xcb_connect(nullptr, nullptr);
-        if (m_conn == nullptr) {
+        if (xcb_connection_has_error(m_conn)) {
             qCritical() << "xcb_connect failed, keyboard extension disabled";
             d->enabled = false;
             return;


src/greeter/XcbKeyboardBackend.cpp
```diff
@@ -80,7 +80,7 @@ namespace SDDM {
         xcb_generic_error_t *error = nullptr;
 
         m_conn = xcb_connect(nullptr, nullptr);
-        if (m_conn == nullptr) {
+        if (xcb_connection_has_error(m_conn)) {
             qCritical() << "xcb_connect failed, keyboard extension disabled";
```
Vogtinator, line 84: In that case it would also have to `xcb_disconnect(m_conn); m_conn = nullptr;` for cleanup.

Edit: Well, doesn't have to but maybe should for completeness.

Vogtinator:
Did you run into this situation?

eszlari:
> Did you run into this situation?

I guess he searched for the function name in multiple projects:

https://invent.kde.org/plasma/kwin/-/merge_requests/6693
https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/31812

Vogtinator (approved):

src/greeter/XcbKeyboardBackend.cpp
lbwtw: your suggestion makes perfect sense, but I'm on vacation for a few days. I was going to work on it today, but you've already merged it.

PR #2028, opened 2025-01-04
Use a fully locked system account with systemd>=257

antonio-rojas:
As per upstream recommendation https://github.com/systemd/systemd/blob/25a306f6c4d9b8596e25d7771802f005ebae91ba/NEWS#L822-L832


antonio-rojas:
This has unintended side effects since it creates expired accounts, which pam doesn't like.

PR #2031, opened 2025-01-18
Fixes #2021

### LINKED ISSUE #2021: sddm-helper clears all environent variables and reset `LANG` to the wrong value for a multi-locale installation
## Summary

On a multi-locale system, which uses a different locale for system daemons and user sessions, sddm-helper falls back to the wrong locale after it has cleared all environment variables. This leads to a faulty value for `LANG` on a system with a multi-locale setup, i.e. a system which uses a different locale for system processes and user processes (i.e. those started by `systemd --user`).

### Prerequisites

This bug only becomes apparent if one uses different locales as the system's default and for user processes. The system's default locale is set as `LANG=C.UTF-8` in `/etc/locale.conf`. This ensures to get unlocalized messages in the system log.

The file `/etc/locale-users.conf` sets `LANG=de_DE.UTF-8` as the default for ordinary user accounts. This file is used used in various places:

 * The file `/etc/environment.d/20-locale.conf` is a symbolic link to `/etc/locale-users.conf`.
    * The PAM module `pam_env.so` parses the directory `/etc/environment.d/` upon session creation.
    * Likewise the user's systemd process `systemd --user`  parses `/etc/environment.d/` during initialization.
 * Moreover, certain user-facing system services uses the file `/etc/locale-users.conf` to setup a localized environment via systemd service drop-ins:
    * `user@.service` and
    * `sddm.service`
   
   both have a drop-in `/etc/systemd/system/{user@.service.d,sddm.service.d}/override.conf` with
   
   ```
   [Service]
   EnvironmentFile=-/etc/locale-users.conf
   ```
   
   This way, SDDM greets user with a localized UI and `systemd --user` runs in a localized environment.

### The Actual Bug

sddm-helper clears the environment variables and then re-constructs a “cleansed” environment. However, in doing so, it resets `LANG` to `C.UTF-8` instead of keeping it at `de_DE.UTF-8`. As a consequence, the spawned desktop environment picks up a wrong locae.

A typical pstree looks like

```
systemd,1 --switched-root --system --deserialize=50 splash
  ├─sddm,711
  │   └─sddm-helper,886 --socket /tmp/sddm-auth-65c418cf-f7c6-4130-8213-9d60d6c82829 --id 1 --start /usr/libexec/plasma-dbus-run-session-if-needed /usr/bin/startplasma-wayland --user matthias
  │       └─startplasma-wayland,935
  └─systemd,908 --user
      └─plasmashell,1221 --no-respawn
```

`sddm` (PID 711) creates the initial login screen. Although this process runs as a child of the system's `systemd` (PID 1), the environment is fine:

```
/proc # cat 711/environ | tr '\0' '\n' | egrep LANG
LANG=de_DE.UTF-8
GDM_LANG=de_DE.UTF-8
LANGUAGE=de_DE:en_US
```

(Because of that service drop-in for `sddm.service`.) When the user attempts to authenticate, `sddm` spawns `sddm-helper`. This `sddm-helper` clears the environment and creates a “cleansed” one:

```
/proc # cat 886/environ | tr '\0' '\n'
LANG="C.UTF-8"
```

(Note: I did not even use egrep here to filter for LANG. That is everything which is set.) `sddm-helper` spawns `startplasma-wayland` which inherits the wrong value for `LANG`

Btw: The environment for the user’s systemd (PID 908) is fine again:

```
/proc # cat 908/environ | tr '\0' '\n' | egrep LANG
LANG=de_DE.UTF-8
GDM_LANG=de_DE.UTF-8
LANGUAGE=de_DE:en_US
```

## Possible solutions for that bug

Maybe `sddm-helper` should not clear all environment variables, but stash away the values of harmless environment variables (such as `LANG`) and restore them to their original value as inherited from `sddm`.

OBSERVED RESULT

On a multi-locale system, which uses a different locales for system daemons and user sessions, `sddm-helper` sets the wrong value for `LANG`. `sddm-helper` clears all environment variables and seemingly re-initializes `LANG` from `/etc/locale.conf`.

EXPECTED RESULT

`sddm-helper` should not clear all environment variables, but keep harmless environment variables at the same value as `sddm`, in particular `LANG`, `GDM_LANG`, `LANGUAGE` and `LC_*`.

nagmat84:
This PR fixes #2021.

First, all non-empty environment variables related to localization are copied from the running process into the environment of the future child process. Only if `LANG` is empty after that, the code reads `/etc/locale.conf` and inserts new environment variables for all non-empty (name,value)-pairs unless the variable is already set. Only as a final rescue, `LANG` is initialized with `C`.

### DIFF
--- a/src/auth/Auth.cpp
+++ b/src/auth/Auth.cpp
@@ -52,6 +52,7 @@ namespace SDDM {
     public:
         Private(Auth *parent);
         ~Private();
+        void initLocale();
         void setSocket(QLocalSocket *socket);
     public slots:
         void dataPending();
@@ -71,6 +72,8 @@ namespace SDDM {
         QProcessEnvironment environment { };
         qint64 id { 0 };
         static qint64 lastId;
+    public:
+        static const QString LOCALE_ENV_VARS[];
     };
 
     qint64 Auth::Private::lastId = 1;
@@ -114,24 +117,7 @@ namespace SDDM {
             , child(new QProcess(this))
             , id(lastId++) {
         SocketServer::instance()->helpers[id] = this;
-        QProcessEnvironment env = child->processEnvironment();
-        bool langEmpty = true;
-        QFile localeFile(QStringLiteral("/etc/locale.conf"));
-        if (localeFile.open(QIODevice::ReadOnly | QIODevice::Text)) {
-            QTextStream in(&localeFile);
-            while (!in.atEnd()) {
-                QStringList parts = in.readLine().split(QLatin1Char('='));
-                if (parts.size() >= 2) {
-                    env.insert(parts[0], parts[1]);
-                    if (parts[0] == QLatin1String("LANG"))
-                        langEmpty = false;
-                }
-            }
-            localeFile.close();
-        }
-        if (langEmpty)
-            env.insert(QStringLiteral("LANG"), QStringLiteral("C"));
-        child->setProcessEnvironment(env);
+        initLocale();
         connect(child, QOverload<int,QProcess::ExitStatus>::of(&QProcess::finished), this, &Auth::Private::childExited);
         connect(child, &QProcess::errorOccurred, this, &Auth::Private::childError);
         connect(request, &AuthRequest::finished, this, &Auth::Private::requestFinished);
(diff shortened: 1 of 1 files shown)


lae:
https://github.com/sddm/sddm/blob/5ef1dba2eb89da7c1bced46915b5758cab2ff8a8/src/auth/Auth.cpp#L147-L149

I believe the (existing) behavior here inserts `"blah.UTF-8"` into `LANG` which fails to identify as UTF-8 because of the quotes when `/etc/locale.conf` wraps values in quotes. If my understanding is correct, could you also strip them here?

Not a maintainer, but from what I can tell several people have run into this locale config issue/warning in `sddm-helper`, me included when I had set `LANG="ja_JP.UTF-8"`. Removing the quotes in `/etc/locale.conf` made the warning go away.

Image

(that error message is a red herring for many of these folk running into SDDM issues, so hopefully this PR gets merged sooner than later...)

lae:
happened to see this PR again

@nagmat84 any thoughts on the above?

roman-zilka:
I hit this issue as well in Gentoo (downstream bug https://bugs.gentoo.org/970226). `sddm-helper` spams the journal with the following:
> Detected locale "C" with character encoding "ANSI_X3.4-1968", which is not UTF-8.
> Qt depends on a UTF-8 locale, and has switched to "C.UTF-8" instead.
> If this causes problems, reconfigure your locale. See the locale(1) manual
> for more information.
```
# cat /etc/locale.conf
LANG="en_US.utf8"
LC_COLLATE="C.utf8"
```
This patch fixes it. Applying 5ef1dba and removing all quotation marks from `/etc/locale.conf` made sddm produce no warning and use the locale set in `LANG`. Let me know if you want me to open a separate issue. This is not a #2021 situation, there is no multi-locales setup here.

By the way, this line in that commit appears like it'll trigger the non-UTF8 locale situation anew.
```cpp
env.insert(QStringLiteral("LANG"), QStringLiteral("C"));
```

PR #2036, opened 2025-01-24
i18n: Update Japanese translation

kisaragi-hiu:
Trying my best to match Japanese KDE.

- Update new strings with lupdate. ...except the (Wayland) string is left as-is without changing its context.
- Try my best to translate the new strings.
- Use "Caps Lock" for "Caps Lock". This is existing convention, already used in eg. kdeplasma-addons, plasma-desktop, and plasma5support.
- ~~Use ユーザー名 for username. This is, for instance, recorded in goo辞書 while ユーザ名 isn't.~~ ~~On (personal) review Japanese KDE convention is ユーザ名 ([like ユーザ会](https://jp.kde.org/)); this PR now matches that.~~ This PR now uses ユーザー again per feedback.

### DIFF
--- a/data/translations/ja.ts
+++ b/data/translations/ja.ts
@@ -5,7 +5,7 @@
     <name>PictureBox</name>
     <message>
         <source>Press to login</source>
-        <translation>押してログイン</translation>
+        <translation>ここを押してログイン</translation>
     </message>
 </context>
 <context>
@@ -104,14 +104,6 @@
     <message>
         <source>Hibernate</source>
         <translation>ハイバネート</translation>
-    </message>
-    <message>
-        <source>Password change aborted because maximum tries reached</source>
-        <translation>最大試行回数に達したため、パスワード変更は中止しました</translation>
-    </message>
-    <message>
-        <source>New password change round! Please input current password again!</source>
-        <translation>新しいパスワード変更ラウンドです! 今のパスワードをもう一度入力してください!</translation>
     </message>
 </context>
 </TS>
--- a/data/translations/ja.ts
+++ b/data/translations/ja.ts
@@ -23,7 +23,7 @@
     </message>
     <message>
         <source>Warning, Caps Lock is ON!</source>
-        <translation>注意: Caps Lock が有効です!</translation>
+        <translation>注意: Caps Lock がオンになっています</translation>
     </message>
     <message>
         <source>Layout</source>
@@ -71,7 +71,7 @@
     </message>
     <message>
         <source>Please enter a password!</source>
-        <translation>パスワードを入力してください!</translation>
+        <translation>パスワードを入力してください</translation>
     </message>
     <message>
         <source>Change password</source>
@@ -79,7 +79,7 @@
     </message>
     <message>
         <source>Enter your username</source>
-        <translation>ユーザ名を入力</translation>
+        <translation>ユーザー名を入力</translation>
     </message>
     <message>
         <source>Enter your password</source>
@@ -91,11 +91,11 @@
     </message>
     <message>
         <source>Show password</source>
-        <translation>パスワードを表示する</translation>
+        <translation>パスワードを表示</translation>
     </message>
     <message>
         <source>Hide password</source>
-        <translation>パスワードを隠す</translation>
+        <translation>パスワードを非表示</translation>
     </message>
     <message>
         <source>Suspend</source>
@@ -107,7 +107,7 @@
     </message>
     <message>
         <source>Password change aborted because maximum tries reached</source>
-        <translation>最大試行回数に達しましたため、パスワード変更は中止しました</translation>
+        <translation>最大試行回数に達したため、パスワード変更は中止しました</translation>
     </message>
     <message>
         <source>New password change round! Please input current password again!</source>
--- a/data/translations/ja.ts
+++ b/data/translations/ja.ts
@@ -5,14 +5,14 @@
     <name>PictureBox</name>
     <message>
         <source>Press to login</source>
-        <translation type="unfinished"></translation>
+        <translation>押してログイン</translation>
     </message>
 </context>
 <context>
     <name>QObject</name>
     <message>
         <source>%1 (Wayland)</source>
-        <translation type="unfinished"></translation>
+        <translation>%1 (Wayland)</translation>
     </message>
 </context>
 <context>
@@ -23,7 +23,7 @@
     </message>
     <message>
         <source>Warning, Caps Lock is ON!</source>
-        <translation>注意: Caps ロックがオンになっています!</translation>
+        <translation>注意: Caps Lock が有効です!</translation>
     </message>
     <message>
         <source>Layout</source>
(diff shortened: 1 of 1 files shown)


data/translations/ja.ts
```diff
@@ -69,5 +69,49 @@
...
+        <source>Enter your password</source>
+        <translation>パスワードを入力</translation>
+    </message>
+    <message>
+        <source>Password:</source>
+        <translation>パスワード:</translation>
+    </message>
+    <message>
+        <source>Show password</source>
+        <translation>パスワードを表示する</translation>
```
phanect: While 表示する is still OK, I think 表示 is a bit more simple and natural.

```suggestion
        <translation>パスワードを表示</translation>
```

data/translations/ja.ts
```diff
@@ -69,5 +69,49 @@
...
+    </message>
+    <message>
+        <source>Hide password</source>
+        <translation>パスワードを隠す</translation>
```
phanect: While 隠す makes sense, I think 非表示 is better to express contrast to 表示.

```suggestion
        <translation>パスワードを非表示</translation>
```

data/translations/ja.ts
```diff
@@ -69,5 +69,49 @@
...
+        <source>Suspend</source>
+        <translation>サスペンド</translation>
+    </message>
+    <message>
+        <source>Hibernate</source>
+        <translation>ハイバネート</translation>
+    </message>
+    <message>
+        <source>Password change aborted because maximum tries reached</source>
+        <translation>最大試行回数に達しましたため、パスワード変更は中止しました</translation>
```
phanect: For messages on the software UI, 「達しましたため」 is a bit too formal.

```suggestion
        <translation>最大試行回数に達したため、パスワード変更は中止しました</translation>
```

data/translations/ja.ts
```diff
@@ -69,5 +69,49 @@
...
+        <source>Please enter a password!</source>
+        <translation>パスワードを入力してください!</translation>
+    </message>
+    <message>
+        <source>Change password</source>
+        <translation>パスワードを変更</translation>
+    </message>
+    <message>
+        <source>Enter your username</source>
+        <translation>ユーザ名を入力</translation>
```
phanect: While we, Japan KDE Users Group, have not renamed our organization name from 日本 KDE **ユーザ**会 to **ユーザー会** in Japanese yet, the recent trend is **ユーザー** (with 長音符: "ー") because expression without 長音符 (ー) is (or was) only used in software and some other specific industries.

```suggestion
        <translation>ユーザー名を入力</translation>
```

data/translations/ja.ts
```diff
@@ -5,14 +5,14 @@
     <name>PictureBox</name>
     <message>
         <source>Press to login</source>
-        <translation type="unfinished"></translation>
+        <translation>押してログイン</translation>
```
phanect: @kisaragi-hiu
It sounds a bit unnatural, but I'm not sure how to make it natural because I don't know the context...
I found it is used [here](https://github.com/sddm/sddm/blob/develop/components/2.0/PictureBox.qml#L125), but not sure how it is displayed. (Screenshot is much appreciated if you have)

data/translations/ja.ts
```diff
@@ -69,5 +69,49 @@
...
+        <source>Hibernate</source>
+        <translation>ハイバネート</translation>
+    </message>
+    <message>
+        <source>Password change aborted because maximum tries reached</source>
+        <translation>最大試行回数に達しましたため、パスワード変更は中止しました</translation>
+    </message>
+    <message>
+        <source>New password change round! Please input current password again!</source>
+        <translation>新しいパスワード変更ラウンドです! 今のパスワードをもう一度入力してください!</translation>
```
phanect: @kisaragi-hiu
"パスワード変更ラウンド" sounds a bit unnatural, but I'm not sure how to make it natural because I don't know the context.
Do you know where this text is used?
(Screenshot is much appreciated if you have)

data/translations/ja.ts
```diff
@@ -69,5 +69,49 @@
         <source>Select your user and enter password</source>
         <translation>ユーザ名を選び、パスワードを入力してください</translation>
     </message>
+    <message>
+        <source>Please enter a password!</source>
+        <translation>パスワードを入力してください!</translation>
```
phanect: In the Japanese context, adding an exclamation mark (!) here sounds like a bit too strong.
To avoid letting users feel offended or blamed, I suggest removing the exclamation mark.

If possible, I want to know why the developers added the original English text to the exclamation mark for better translation.

```suggestion
        <translation>パスワードを入力してください</translation>
```

phanect (changes requested):
I'm a native Japanese speaker who got a review request for this PR via the KDE Japanese community mailing list.
I proofread the translations by @kisaragi-hiu, who is not a native Japanese speaker.

data/translations/ja.ts
phanect: Is this text used anywhere?
I quickly searched this repository and suspect this text is no longer used.

data/translations/ja.ts
phanect: Is this text used anywhere?
I quickly searched this repository and suspect this text is no longer used.

data/translations/ja.ts
```diff
@@ -23,7 +23,7 @@
     </message>
     <message>
         <source>Warning, Caps Lock is ON!</source>
-        <translation>注意: Caps ロックがオンになっています!</translation>
+        <translation>注意: Caps Lock が有効です!</translation>
```
phanect: I want to revert this translation except for ロック → Lock.
The reason why I reverted the most part:

- I quickly googled and *ON*/*OFF* (*オン*/*オフ*) looks like a bit more popular expression than *有効*/*無効* in Japanese. In addition, the original text is not *enabled* but *ON*, I think *オン* is a better translation than *有効*.
- I think になっています is better to express the current *status* of Caps Lock than です.

I also removed the exclamation mark (!) because it sounds a bit too emphasized in Japanese.

```suggestion
        <translation>注意: Caps Lock がオンになっています</translation>
```

data/translations/ja.ts
kisaragi-hiu: It's used in the default Maui theme, shown for users that do not have passwords.

螢幕截圖_20250126_010258

data/translations/ja.ts
kisaragi-hiu: These are from [TextConstants.qml](https://github.com/sddm/sddm/blob/2d27436bab48eb49e4607ab073ab7ad1a98b9350/components/common/TextConstants.qml#L4), and themes use it for common strings. For instance, [sddm-astronaut-theme](https://github.com/search?q=repo%3AKeyitdev%2Fsddm-astronaut-theme%20textconstants&type=code); [other search results](https://github.com/search?q=sddm+theme+textconstants+NOT+language%3Axml&type=code)

data/translations/ja.ts
kisaragi-hiu: I'm not sure this is used anywhere either. Searches for the text or [the key](https://github.com/sddm/sddm/blob/2d27436bab48eb49e4607ab073ab7ad1a98b9350/components/common/TextConstants.qml#L50) on GitHub just land back at [TextConstants.qml](https://github.com/sddm/sddm/blob/2d27436bab48eb49e4607ab073ab7ad1a98b9350/components/common/TextConstants.qml) in this repository.

The name of the key suggests it might've been meant for [PAM lockout (max tries)](https://wiki.archlinux.org/title/Security#Lock_out_user_after_three_failed_login_attempts) at some point, but then the text just doesn't make sense.

data/translations/ja.ts
phanect: Thank you.
I could not find the best translation, but I suggest adding ここを.
The new Japanese translation literally means "Press **here** to login."

```suggestion
        <translation>ここを押してログイン</translation>
```

data/translations/ja.ts
phanect: Then I think we can keep your current translation for now.
The current translation is not meaningful enough, but it is much better than showing the wrong information by trying a more natural translation.

phanect: Or we have another option to remove this Japanese translation for now.

data/translations/ja.ts
phanect: Got it. The text is used in another repositories. Thanks!

data/translations/ja.ts
Vogtinator: FWICT it's used in the stale `expired-password` branch and was added such that themes can use some SDDM provided translations. I don't think it's used at all ATM.

data/translations/ja.ts
kisaragi-hiu: Applied in [`2aaa80d` (#2036)](https://github.com/sddm/sddm/pull/2036/commits/2aaa80da6995ce3f57d286d4fde92851264779f6)

data/translations/ja.ts
kisaragi-hiu: I've removed these two strings from the PR.

phanect (approved):
Thanks, looks good to me.

Vogtinator (approved):

PR #2047, opened 2025-02-24
Allow themes to disable delays on login failure in the PAM backend

akasure:
- Define "requiresZeroDelayOnFailAuth" as a theme property, that can be set to true to disable login failure delays from PAM
-  This property is exposed by the config to the Greeter, so if not set the greeter can fall back to its old behaviour
- Required for https://bugs.kde.org/show_bug.cgi?id=407473

### DIFF
--- a/src/helper/backend/PamHandle.cpp
+++ b/src/helper/backend/PamHandle.cpp
@@ -23,6 +23,11 @@
 #include <QtCore/QDebug>
 
 namespace SDDM {
+
+    static void fail_delay(int, unsigned, void*) {
+
+    }
+
     bool PamHandle::putEnv(const QProcessEnvironment& env) {
         const auto envs = env.toStringList();
         for (const QString& s : envs) {
@@ -84,6 +89,7 @@ namespace SDDM {
 
     bool PamHandle::authenticate(int flags) {
         qDebug() << "[PAM] Authenticating...";
+        pam_fail_delay(m_handle, 0);
         m_result = pam_authenticate(m_handle, flags | m_silent);
         if (m_result != PAM_SUCCESS) {
             qWarning() << "[PAM] authenticate:" << pam_strerror(m_handle, m_result);
@@ -144,7 +150,7 @@ namespace SDDM {
         return c->converse(n, msg, resp);
     }
 
-    bool PamHandle::start(const QString &service, const QString &user) {
+    bool PamHandle::start(const QString &service, const QString &user, const bool requiresZeroDelayOnFailAuth) {
         if (user.isEmpty())
             m_result = pam_start(qPrintable(service), NULL, &m_conv, &m_handle);
         else
@@ -156,6 +162,15 @@ namespace SDDM {
         else {
             qDebug() << "[PAM] Starting...";
         }
+#ifdef PAM_FAIL_DELAY
+        qDebug() << "[PAM] requiresZeroDelayOnFailAuth:" << requiresZeroDelayOnFailAuth;
+        if (requiresZeroDelayOnFailAuth) {
+            setItem(PAM_FAIL_DELAY, reinterpret_cast< void* >(fail_delay));
+        }
+#else
+        Q_UNUSED(requiresZeroDelayOnFailAuth);
+        Q_UNUSED(fail_delay);
+#endif
         return true;
     }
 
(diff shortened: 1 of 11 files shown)


Vogtinator:
Is there a way to catch the start of the delay and pass it to the greeter instead? I'd still like the delay to be handled and configured by PAM in a single place.

akasure:
> Is there a way to catch the start of the delay and pass it to the greeter instead? I'd still like the delay to be handled and configured by PAM in a single place.

I think it isn't possible. The Helper and the Daemon seem to be responding immediately when control returns from pam_authenticate and pam_acct_mgmt. Also according to https://linux.die.net/man/3/pam_fail_delay, it looks like the delay is generated by PAM itself, and control is only returned after the delay times out. I can definitely take a look again just to make sure.

davidedmundson:
>it looks like the delay is generated by PAM itself, and control is only returned after the delay times out. I can definitely take a look again just to make sure.

you can't log in again till it times out but you can pass this information to the greeter. One of the arguments on the fail delay callback is a timeout. I would happily approve forwarding that information to the greeter.

src/helper/backend/PamHandle.cpp
```diff
@@ -84,6 +89,7 @@ namespace SDDM {
 
     bool PamHandle::authenticate(int flags) {
         qDebug() << "[PAM] Authenticating...";
+        pam_fail_delay(m_handle, 0);
```
akasure: This change snuck in and wasn't intended

akasure:
> > it looks like the delay is generated by PAM itself, and control is only returned after the delay times out. I can definitely take a look again just to make sure.
> 
> you can't log in again till it times out but you can pass this information to the greeter. One of the arguments on the fail delay callback is a timeout. I would happily approve forwarding that information to the greeter.

I can add in the forwarding changes, but I don't think that would help too much with the intention of https://bugs.kde.org/show_bug.cgi?id=407473.

Sorry if I misunderstood your comment, but using pam_fail_delay wouldn't do much for this case, as it is just a request made to PAM, and doesn't guarantee that the requested timeout would be adhered to.

Also I just took a glance at the logging from the PAM backend, and it really doesn't look like we can inform the greeter about the login failed event before the fail delay times out. The control lies with pam_unix till the delay times out. I added in a pam_fail_delay of 10 seconds to  make the logs clearer.
image

src/helper/backend/PamHandle.cpp
akasure: Removed now

davidedmundson:
What I meant was: 

```
pam_set_item(m_handle, PAM_FAIL_DELAY, (void *)fail_delay);


static void fail_delay(int retval, unsigned usec_delay, void *appdata_ptr)
{
   forward the used_delay here, then the UI can become enabled again the moment the backend becomes available again
}
```

this would allow a user to change things in pam and have the UI work accordingly.

----

Given I know you hacked on kscreenlocker too, it might be easier to do any testing mods there as that's not multi-process. 

It's been on my TODO list for a while, so ping if you get stuck and I can take a look.

Vogtinator:
What @davidedmundson describes would also be my preferred option, if it works as expected. Should be easier to implement and actually use safely in themes.

akasure:
I overlooked that approach. It definitely makes more sense, and would keep PAM behaviour consistent. I'll make the changes and update this MR

akasure:
@davidedmundson do you think we should get rid of the grace lock timer in the LockscreenUi from plasma-desktop, and replace it with the same event that's being discussed here (but in KScreenLocker )? That way the lockscreen and sddm theme would be consistent with the PAM config

davidedmundson:
>@davidedmundson do you think we should get rid of the grace lock timer in the LockscreenUi 

We need to be careful, it seems some past dev got confused.

There's something called a gracelock in the lock screen UI. This happens after an idle timeout, where the lockscreen loads, but if you wiggle the mouse in the first N seconds it just closes immediately before we've done a proper lock. This is authenticator.graceLocked 

don't remove this.

There's also:

```
        Timer {
            id: graceLockTimer
            interval: 3000
            onTriggered: {
                root.clearPassword();
                authenticator.startAuthenticating();
            }
        }
        ```
        
which we should remove, to be some new property on the authenticator, based on what we're discussing above.
Don't call it grace-anything this time please!

akasure:
Hi @davidedmundson ,
Made an initial set of changes in the following MRs:
https://github.com/sddm/sddm/pull/2051
https://invent.kde.org/plasma/plasma-workspace/-/merge_requests/5253
https://invent.kde.org/plasma/kscreenlocker/-/merge_requests/272
https://invent.kde.org/plasma/plasma-desktop/-/merge_requests/2851

Need to perform another round of sanity myself, but it would help if you could take a look. I wanted to make sure I'm going in the right direction.

PR #2034, opened 2025-01-24
CI: Replace ubuntu 23.04 with 24.04

Vogtinator:
23.04 went EOL in 2024-01. Switch to 24.04 which is LTS.

PR #2051, opened 2025-03-02
Add loginFailedDelayStarted event

akasure:
- Prevent duplicate loginFailed event
- On authentication failure, inform the greeter about the start of the PAM delay by triggering the loginFailedDelayStarted event.
- Required for https://bugs.kde.org/show_bug.cgi?id=407473

### DIFF
--- a/src/auth/Auth.cpp
+++ b/src/auth/Auth.cpp
@@ -210,6 +210,13 @@ namespace SDDM {
                     str.send();
                     break;
                 }
+                case LOGIN_FAILED_DELAY_STARTED: {
+                    uint uSecDelay {0};
+                    str >> uSecDelay;
+                    qDebug() << "Auth: Login failed delay started notification received; delay duration:" << uSecDelay;
+                    Q_EMIT auth->loginFailedDelayStarted(uSecDelay);
+                    break;
+                }
                 default: {
                     Q_EMIT auth->error(QStringLiteral("Auth: Unexpected value received: %1").arg(m), ERROR_INTERNAL);
                 }
--- a/src/auth/Auth.h
+++ b/src/auth/Auth.h
@@ -230,6 +230,13 @@ namespace SDDM {
         */
         void info(QString message, Auth::Info type);
 
+        /**
+        * Event to indicate that a wrong password lockout has been started, with a certain duration.
+        *
+        * @param uSecDelay: The duration of the lockout delay (in MicroSeconds)
+        */
+        void loginFailedDelayStarted(const uint uSecDelay);
+
     private:
         class Private;
         class SocketServer;
--- a/src/auth/AuthMessages.h
+++ b/src/auth/AuthMessages.h
@@ -98,6 +98,7 @@ namespace SDDM {
         AUTHENTICATED,
         SESSION_STATUS,
         DISPLAY_SERVER_STARTED,
+        LOGIN_FAILED_DELAY_STARTED,
         MSG_LAST,
     };
 
--- a/src/common/Messages.h
+++ b/src/common/Messages.h
@@ -39,6 +39,7 @@ namespace SDDM {
         LoginSucceeded,
         LoginFailed,
         InformationMessage,
+        LoginFailedDelayStarted // Emitted after we know that the login attempt has failed, but PAM still hasn't returned control. This is the start of the failure delay.
     };
 
     enum Capability {
--- a/src/daemon/SocketServer.cpp
+++ b/src/daemon/SocketServer.cpp
@@ -200,6 +200,11 @@ namespace SDDM {
         SocketWriter(socket) << quint32(DaemonMessages::LoginSucceeded);
     }
 
+    void SocketServer::loginFailedDelayStarted(QLocalSocket *socket, const uint uSecDelay)
+    {
+        SocketWriter(socket) << static_cast< quint32 >(DaemonMessages::LoginFailedDelayStarted) << uSecDelay;
+    }
+
     void SocketServer::informationMessage(QLocalSocket *socket, const QString &message) {
         SocketWriter(socket) << quint32(DaemonMessages::InformationMessage) << message;
     }
--- a/src/helper/HelperApp.cpp
+++ b/src/helper/HelperApp.cpp
@@ -152,8 +152,6 @@ namespace SDDM {
 
         Q_ASSERT(getuid() == 0);
         if (!m_backend->authenticate()) {
-            authenticated(QString());
-
             // write failed login to btmp
             const QProcessEnvironment env = m_session->processEnvironment();
             const QString displayId = env.value(QStringLiteral("DISPLAY"));
@@ -207,6 +205,14 @@ namespace SDDM {
     void HelperApp::error(const QString& message, Auth::Error type) {
         SafeDataStream str(m_socket);
         str << Msg::ERROR << message << type;
+        str.send();
+        m_socket->waitForBytesWritten();
+    }
+
+    void HelperApp::loginFailedDelayStarted(const uint uSecDelay)
+    {
+        SafeDataStream str(m_socket);
+        str << Msg::LOGIN_FAILED_DELAY_STARTED << uSecDelay;
         str.send();
         m_socket->waitForBytesWritten();
     }
--- a/src/helper/backend/PamHandle.cpp
+++ b/src/helper/backend/PamHandle.cpp
@@ -21,8 +21,25 @@
 #include "PamBackend.h"
 
 #include <QtCore/QDebug>
+#include <QThread>
 
 namespace SDDM {
+    static void fail_delay(int retval, uint usec_delay, void* appdata_ptr) {
+        auto* backend = reinterpret_cast< PamBackend* >(appdata_ptr); // Refer the pam_conv (@sa m_conv) structure for info on appdata_ptr
+        if (!backend) {
+            qFatal() << "[PAM] appdata_ptr not convertible to a valid PamBackend! Cannot apply fail delay";
+            return;
+        }
+        if (retval == PAM_SUCCESS) {
+            qDebug() << "[PAM] Fail delay function was called, but authentication result was a success!";
+            return;
+        }
+        Q_EMIT backend->loginFailedDelayStarted(usec_delay);
+        if (usec_delay > 0u) {
+            QThread::usleep(usec_delay); // This calls nanosleep as of Qt 6.8. It also handles EINTR (restarts the sleep with the remainder duration when interrupted), but not EFAULT or EINVAL.
+        }
+    }
+
     bool PamHandle::putEnv(const QProcessEnvironment& env) {
         const auto envs = env.toStringList();
         for (const QString& s : envs) {
@@ -149,6 +166,13 @@ namespace SDDM {
             m_result = pam_start(qPrintable(service), NULL, &m_conv, &m_handle);
         else
             m_result = pam_start(qPrintable(service), qPrintable(user), &m_conv, &m_handle);
+
+#if defined(HAVE_PAM_FAIL_DELAY)
+        setItem(PAM_FAIL_DELAY, reinterpret_cast< void* >(fail_delay));
+#else
+        Q_UNUSED(fail_delay);
+#endif
+
         if (m_result != PAM_SUCCESS) {
             qWarning() << "[PAM] start" << pam_strerror(m_handle, m_result);
             return false;
--- a/src/daemon/Display.cpp
+++ b/src/daemon/Display.cpp
@@ -147,6 +147,7 @@ namespace SDDM {
         connect(m_auth, &Auth::finished, this, &Display::slotHelperFinished);
         connect(m_auth, &Auth::info, this, &Display::slotAuthInfo);
         connect(m_auth, &Auth::error, this, &Display::slotAuthError);
+        connect(m_auth, &Auth::loginFailedDelayStarted, this, &Display::slotAuthLoginFailedDelayStarted);
 
         // restart display after display server ended
         connect(m_displayServer, &DisplayServer::started, this, &Display::displayServerStarted);
@@ -534,6 +535,14 @@ namespace SDDM {
         m_socketServer->informationMessage(m_socket, message);
         if (error == Auth::ERROR_AUTHENTICATION)
             emit loginFailed(m_socket);
+    }
+
+    void Display::slotAuthLoginFailedDelayStarted(const uint uSecDuration)
+    {
+        if (!m_socket)
+            return;
+
+        m_socketServer->loginFailedDelayStarted(m_socket, uSecDuration);
     }
 
     void Display::slotHelperFinished(Auth::HelperExitStatus status) {
--- a/src/daemon/Display.h
+++ b/src/daemon/Display.h
@@ -118,6 +118,7 @@ namespace SDDM {
         void slotHelperFinished(Auth::HelperExitStatus status);
         void slotAuthInfo(const QString &message, Auth::Info info);
         void slotAuthError(const QString &message, Auth::Error error);
+        void slotAuthLoginFailedDelayStarted(const uint uSecDuration);
     };
 }
 
--- a/src/daemon/SocketServer.h
+++ b/src/daemon/SocketServer.h
@@ -49,6 +49,7 @@ namespace SDDM {
         void informationMessage(QLocalSocket *socket, const QString &message);
         void loginFailed(QLocalSocket *socket);
         void loginSucceeded(QLocalSocket *socket);
+        void loginFailedDelayStarted(QLocalSocket *socket, const uint uSecDelay);
 
     signals:
         void login(QLocalSocket *socket,
--- a/src/greeter/GreeterProxy.cpp
+++ b/src/greeter/GreeterProxy.cpp
@@ -214,6 +214,13 @@ namespace SDDM {
                     emit informationMessage(message);
                 }
                 break;
+                case DaemonMessages::LoginFailedDelayStarted: {
+                    uint delay {0};
+                    input >> delay;
+                    qDebug() << "Login failed delay started notification received from daemon; delay duration:" << delay;
+                    Q_EMIT loginFailedDelayStarted(delay);
+                    break;
+                }
                 default: {
                     // log message
                     qWarning() << "Unknown message received from daemon.";
--- a/src/greeter/GreeterProxy.h
+++ b/src/greeter/GreeterProxy.h
@@ -81,6 +81,7 @@ namespace SDDM {
 
         void socketDisconnected();
         void loginFailed();
+        void loginFailedDelayStarted(const uint uSecDelay);
         void loginSucceeded();
 
     private:
--- a/src/helper/Backend.cpp
+++ b/src/helper/Backend.cpp
@@ -38,6 +38,15 @@ namespace SDDM {
     Backend::Backend(HelperApp* parent)
             : QObject(parent)
             , m_app(parent) {
+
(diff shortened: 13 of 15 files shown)


src/auth/AuthMessages.h
```diff
@@ -94,6 +94,7 @@ namespace SDDM {
         HELLO = 1,
         ERROR,
         INFO,
+        NOTIFICATION, // Notify listeners of an event.
```
Vogtinator: To not break running sddm daemons, only append at the end

src/common/Messages.h
```diff
@@ -38,6 +38,7 @@ namespace SDDM {
         Capabilities,
         LoginSucceeded,
         LoginFailed,
+        LoginFailedDelayStarted, // Emitted after we know that the login attempt has failed, but PAM still hasn't returned control. This is the start of the failure delay.
```
Vogtinator: Only append at the end

src/auth/AuthMessages.h
akasure: Done. Noted for future MRs

src/common/Messages.h
akasure: Done

src/helper/backend/PamHandle.cpp
```diff
@@ -17,12 +17,36 @@
...
+        }
+        if (PAM_SUCCESS == retval) {
+            qDebug() << "[PAM] Fail delay function was called, but authentication result was a success!";
+            return;
+        }
+        Q_EMIT backend->loginFailedDelayStarted(usec_delay);
+        if (0 < usec_delay) {
+            // Refer pam_delay.c (linux-pam) for the default implementation of this delay
+            timeval delay {.tv_sec = usec_delay / uSecMultiplier, .tv_usec = usec_delay % uSecMultiplier};
+            select(0, nullptr, nullptr, nullptr, &delay);
```
Vogtinator: I see that it's exactly the same way in pam_delay.c, but I wonder about early returns of select which might only wait for a fraction of the passed delay... I guess only signals would trigger this.

akasure: Thought about switching to pselect for this. It has a signal mask (https://linux.die.net/man/2/pselect). But I wasn't too sure if masking was a good idea.

akasure:
Hi @Vogtinator ,
I'm new to the SDDM pipelines, and I'm not sure if there's anything I can do to fix these failing checks. Can you please help out?
Example: https://github.com/sddm/sddm/actions/runs/13628744617/job/38094384765?pr=2051

Vogtinator:
#2034 should fix most of them, waiting for reviews for a while now without progress.

akasure:
Hi @davidedmundson,
Please take a look too

src/helper/backend/PamHandle.cpp
mkow: You can fix that without switching to `pselect()`, like that:
```C++
while (select(0, nullptr, nullptr, nullptr, &delay) == -1 && errno == EINTR) {
}
```

But... this works only on Linux, unfortunately. See "The timeout" section in https://man7.org/linux/man-pages/man2/select.2.html - POSIX doesn't require the `timeout` argument to be updated after an interrupted call, so it seems that outside of Linux you can't use `select()` nor `pselect()` to sleep reliably. And this could be important here, as this is intended to be a security measure.
So, I'd say you should use a looped `nanosleep()` here instead?

src/auth/AuthMessages.h
akasure: Adding NOTIFICATION after MSG_LAST does not work out, due to checks in SafeDataStream

src/helper/backend/PamHandle.cpp
akasure: I couldn't figure out if linux-pam was allowing signals to interrupt the sleep/delay intentionally or not. Trusted them to have had a reason. If we are enforcing this delay in SDDM/KScreenLocker regardless of signals, then I guess we have two POSIX spec options.
-> nanosleep
-> clock_nanosleep (https://man7.org/linux/man-pages/man2/clock_nanosleep.2.html)
Using clock_nanosleep would add a little more arithmetic, but would also increase accuracy. However I think the slight entropy on resuming sleeps with nanosleep might actually be beneficial. So I went with nanosleep for now.
@mkow can you please take a look again?

PS Would've been nice and clean if usleep (https://linux.die.net/man/3/usleep) wasn't deprecated, since it has the same scale as PAM delays.

src/helper/HelperApp.cpp
```diff
@@ -152,8 +152,6 @@ namespace SDDM {
 
         Q_ASSERT(getuid() == 0);
         if (!m_backend->authenticate()) {
-            authenticated(QString());
```
akasure, line 155: Removed this line, as it seemed to be triggering a duplicate loginFailed event. I couldn't find the utility behind this. Hope I didn't miss anything.

m_backend->authenticate() already notifies of failures.

src/helper/backend/PamHandle.cpp
```diff
@@ -17,12 +17,38 @@
...
+        if (PAM_SUCCESS == retval) {
+            qDebug() << "[PAM] Fail delay function was called, but authentication result was a success!";
+            return;
+        }
+        Q_EMIT backend->loginFailedDelayStarted(usec_delay);
+        if (0u < usec_delay) {
+            const uint delaySecs {usec_delay / uSecMultiplier};
+            const uint delayNSecs {(usec_delay - (delaySecs * uSecMultiplier)) * uSecToNanoSecMultiplier}; // Reuse delaySecs, instead of using "%".
+            timespec req {.tv_sec = delaySecs, .tv_nsec = delayNSecs}, rem {};
+            while (-1 == nanosleep(&req, &rem) && EINTR == errno) {
```
mkow: Why using Yoda Conditionals? It's quite unintuitive to read / not idiomatic, and nowadays it doesn't provide any advantages in terms of being error-prone, because compilers detect accidental `=` in IFs automatically.
Also, the code in this file doesn't seem to use this convention.

src/helper/backend/PamHandle.cpp
```diff
@@ -17,12 +17,38 @@
...
+#include <ctime>
+
 #include "PamHandle.h"
 #include "PamBackend.h"
 
 #include <QtCore/QDebug>
 
 namespace SDDM {
+    static void fail_delay(int retval, uint usec_delay, void* appdata_ptr) {
+        constexpr uint uSecMultiplier {1000000}, uSecToNanoSecMultiplier {1000};
```
mkow: Stylistic nitpick: (feel free to ignore)
I'd just remove these constants and use them directly in the computations, it's not too helpful. Or declare them right before the time computations, so the reader doesn't need to look up their definitions when reading the delay code.

src/helper/backend/PamHandle.cpp
mkow: Probably not important, but note that you're skipping error checks here. I'm not sure if it's worth handling though, the only other possible errors are `EFAULT` and `EINVAL` and both can't really happen with this invocation.
So, it's more like a note to the maintainers whether they want to be super safe here.

src/helper/backend/PamHandle.cpp
mkow: The current code looks correct, thanks for the change!

> I couldn't figure out if linux-pam was allowing signals to interrupt the sleep/delay intentionally or not.

To me that seems like a bad coding practice from their side, you shouldn't really assume that there are no signal handlers in your program. It may be true currently, but then someone makes a change at the other end of the codebase in a completely unrelated place, or a library you use adds a handler and suddenly you have a security issue in this code.

src/helper/backend/PamHandle.cpp
akasure: Aah. It's almost inveterate at this point. It's enforced in my workplace, and it comes to me without a second thought. But I agree that it's against the style of the codebase. Made the changes.
"When in SDDM, be like SDDM"

akasure: I also don't think these errors are possible given how the function is being invoked. Not sure how we would handle these errors (if ever) either. A retry probably doesn't make sense. Do we maybe add logs?

src/helper/backend/PamHandle.cpp
akasure: Would like to keep the constants personally. I feel it brings out the intention better. But I will move their init to right before their usage, so that it's easier to find.

src/helper/backend/PamHandle.cpp
mkow: No idea, panic the whole process maybe? This should rather have semantics of an assertion failure, because it _should_ be impossible to reach. Anyways, I guess up to the maintainers whether they are ok with the current code.

mkow: > It's enforced in my workplace

Sounds like a rule which was introduced 30 years ago and never thought about again :) Nowadays it's IMO strictly worse than writing it normally.

akasure: Maybe this is a fruit of habit, but I actually find yoda notations to be more readable for equality comparisons against constants (and enums). For other operators though, it definitely makes the code look more like a riddle.

akasure: @davidedmundson any opinions on how we should handle EFAULT and EINVAL?

mkow: Yeah, I guess it's just your habit, because of your company policies ;) I've never worked in a company enforcing this and it reads really strange and unnatural to me.
Fortunately nowadays almost no one writes code this way :)

akasure: @mkow switched to https://doc.qt.io/qt-6/qthread.html#usleep. It does exactly what the current impl was doing, but with a neat Qt wrapper. Handling EINVAL, and EFAULT is still not covered though.

It also takes micro seconds. So no more multipliers in our code

src/auth/Auth.cpp
```diff
@@ -210,6 +210,23 @@ namespace SDDM {
                     str.send();
                     break;
                 }
+                case NOTIFICATION: {
+                    uint uSecDuration {0};
+                    Notification type {NOTIFICATION_NONE};
+                    str >> uSecDuration >> type;
```
Vogtinator: This doesn't quite match. If `NOTIFICATION` is meant to be a generically used event, which is implied by the existence of the expandable `type`, then `uSecDuration` needs to be type specific

src/auth/Auth.h
```diff
@@ -80,6 +80,12 @@ namespace SDDM {
         };
         Q_ENUM(Error)
 
+        enum Notification {
+            NOTIFICATION_NONE = 0,
```
Vogtinator: What is `NONE` useful for?

src/daemon/SocketServer.cpp
```diff
@@ -200,6 +200,11 @@ namespace SDDM {
         SocketWriter(socket) << quint32(DaemonMessages::LoginSucceeded);
     }
 
+    void SocketServer::loginFailedDelayStarted(QLocalSocket *socket, const uint uSecDelay)
+    {
+        SocketWriter(socket) << static_cast< quint32 >(DaemonMessages::LoginFailedDelayStarted) << uSecDelay;
```
Vogtinator, line 205: For consistency:

```suggestion
        SocketWriter(socket) << quint32(DaemonMessages::LoginFailedDelayStarted) << uSecDelay;
```

src/helper/backend/PamHandle.cpp
```diff
@@ -17,12 +17,30 @@
...
 #include "PamBackend.h"
 
 #include <QtCore/QDebug>
+#include <QThread>
 
 namespace SDDM {
+    static void fail_delay(int retval, uint usec_delay, void* appdata_ptr) {
+        auto* backend = reinterpret_cast< PamBackend* >(appdata_ptr); // Refer the pam_conv (@sa m_conv) structure for info on appdata_ptr
+        if (!backend) {
+            qWarning() << "[PAM] appdata_ptr is invalid; Cannot apply delay";
```
Vogtinator: Why not?

IMO it needs to make sure that the delay is performed whenever possible

src/helper/backend/PamHandle.cpp
```diff
@@ -17,12 +17,30 @@
...
+            return;
+        }
+        if (retval == PAM_SUCCESS) {
```
Vogtinator, line 33: Can this actually happen?

Vogtinator (approved):
(Misclicked - meant to request changes)

Vogtinator (changes requested):
I like the design! Even though it's a bit annoying that the event has to be passed around that much...

Does it still work to log out and ideally also back in if SDDM is updated while a user is logged in?

src/helper/backend/PamHandle.cpp
mkow: Looks good!

src/helper/backend/PamHandle.cpp
davidedmundson: We're in sync :D

To save OP replying to the same thing twice.
 https://invent.kde.org/plasma/kscreenlocker/-/merge_requests/272#note_1159510

src/helper/backend/PamHandle.cpp
davidedmundson: >IMO it needs to make sure that the delay is performed whenever possible

If we're ever in this state, things have gone seriously seriously wrong. I don't think we need to be delaying. 

We could bump it to qFatal so we don't have to worry about handling it.

src/helper/backend/PamHandle.cpp
Vogtinator: Ok, hm.

gdb/ltrace would be useful IMO or maybe just ignore this altogether given that delay == 0 anyway?

src/auth/Auth.cpp
akasure: I'd initially kept uSecDuration specific to NOTIFICATION_LOGIN_FAILED_DELAY_STARTED, but then for the sake of simplicity/utility, I added this uSecDuration to the notification trigger (`void notification(const uint uSecDuration, Auth::Notification type`)
`@param uSecDuration: Duration for which the notification can be treated as valid (in MicroSeconds)`.
Since most notifications do have a validUntil field, I though we could reuse the same for all notification types. Maybe we could also pass in 0 if we don't want this behavior for a certain notification type?

src/auth/Auth.h
akasure: I guess it's similar to INFO_NONE and ERROR_NONE. For the scope of this diff, I'm using it to initialize the notification type in the notification clients. Ofcourse I could `static_cast< Notification >(-1)` to signify a no-op, but that looks messier to me. (also undefined behavior if not careful)

src/daemon/SocketServer.cpp
akasure: I'll add in the changes if you really think it's necessary, but I feel we're better off with a static_cast than a C style cast. It brings out the intention better according to me. I know it isn't fully consistent with the above lines, but I don't think it comes at a cost to readability either, which is usually the argument for these changes.
@Vogtinator please let me know if we can keep the static_cast

src/helper/backend/PamHandle.cpp
akasure: > If we're ever in this state, things have gone seriously seriously wrong. I don't think we need to be delaying.

Yep. I think if we ever got here, then a lot would've failed before, and we'd definitely not proceed with authentication attempts.

> We could bump it to qFatal so we don't have to worry about handling it.

Will do. I'll also make the log a little more explicit

src/helper/backend/PamHandle.cpp
akasure: Hi @Vogtinator @davidedmundson ,
I went through the logs again and I was wrong. The function is being called with PAM_SUCCESS, but not with a zero delay.
image
Didn't get a chance to take a look at kwallet-pam yet, but do you know if it triggers this call?
If it's not expected, I can also annotate the branch as "unlikely"

akasure: > maybe just ignore this altogether given that delay == 0 anyway?

It would be better to treat them separately even if the delay was 0 imo. That way we can differentiate between a success, and a failure with PAM configured to "no delay"

akasure:
> Does it still work to log out and ideally also back in if SDDM is updated while a user is logged in?

I don't think it would, especially since there are daemon changes. I usually just restart sddm

src/helper/backend/PamHandle.cpp
akasure: > We could bump it to qFatal so we don't have to worry about handling it.

Done

src/helper/backend/PamHandle.cpp
akasure: `2025-03-12T19:14:03.594629+05:30 akasure-alienwarex15r2 (systemd): pam_unix(systemd-user:session): session opened for user sddm(uid=115) by sddm(uid=0)
2025-03-12T19:14:06.737124+05:30 akasure-alienwarex15r2 sddm-helper: pam_kwallet5(sddm:auth): pam_kwallet5: pam_sm_authenticate
2025-03-12T19:14:06.766876+05:30 akasure-alienwarex15r2 kscreenlocker_greet: pam_unix(kde:auth): authentication failure; logname=akasure uid=1000 euid=1000 tty= ruser= rhost=  user=akasure
2025-03-12T19:14:06.996392+05:30 akasure-alienwarex15r2 sddm-helper: pam_unix(sddm-greeter:session): session closed for user sddm
2025-03-12T19:14:06.999003+05:30 akasure-alienwarex15r2 systemd-logind[1043]: Removed session 5.
2025-03-12T19:14:17.257908+05:30 akasure-alienwarex15r2 (sd-pam): pam_unix(systemd-user:session): session closed for user sddm`

From the above PAM logs, the fail_delay might be getting called by kscreenlocker_greet: pam_unix(kde:auth).
@davidedmundson @Vogtinator would it be possible to check if this is happening for you all too? It could be only my machine too, since I have messed with it in the past (And lost track).

Can we proceed with this branch in the code (maybe with Q_UNLIKELY), or would you prefer to debug this PAM_SUCCESS fail_delay call first? I can take a look at this bit only over the weekend unfortunately.

src/auth/Auth.cpp
arrowd: To me this addition feels too general. I'd rather have a specific `Msg::LOGIN_FAILED_DELAY` and only generalize it to `NOTIFICATION` if we eventually get too many message types with common properties.

akasure: Done. I think this also addresses concerns from @Vogtinator that the duration should be notification type specific. Maybe it makes more sense to keep this notification impl for later. Moving `Msg::AUTHENTICATED`, `Msg::SESSION_STATUS`, and `Msg::DISPLAY_SERVER_STARTED` to this notification type might make sense in a separate PR.

src/auth/Auth.cpp
```diff
@@ -210,6 +210,13 @@ namespace SDDM {
                     str.send();
                     break;
                 }
+                case LOGIN_FAILED_DELAY_STARTED: {
+                    uint uSecDelay {0};
+                    str >> uSecDelay;
+                    qDebug() << "Auth: Login failed delay started notification received; delay duration:" << uSecDelay;
+                    Q_EMIT auth->loginFailedDelayStarted(uSecDelay);
```
arrowd, line 217: I see that other cases of this switch not only emit an event, but also send some data back. Shouldn't we do that there too?

src/helper/HelperApp.cpp
arrowd: Looks like it, nice catch.

src/auth/Auth.cpp
akasure: I'm not sure if there is any information worth passing back here. Looks like Info and Error don't relay any data back either.

PR #2063, opened 2025-03-26
Fix index check for UserModel

wineee:
when 'index.row() == d->users.count()', it's an out-of-bounds access to an array


davidedmundson (approved):

davidedmundson:
there's also a handy checkIndex() function in newer Qt

PR #2080, opened 2025-04-23
Use QProcessEnvironment::systemEnvironment() to set initial process env

### LINKED ISSUE #2079: XorgDisplayServer::setupDisplay() ignores system environment variables
sddm manually configures the process environment variables in XorgDisplayServer::setupDisplay() which can cause mesa shader caches to be created in the root of the filesystem /.cache. 

This occurs because mesa added $HOME as one of the environment variables to search when looking for the cache location. 
[https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/13614/diffs?commit_id=04f232ed99bb68ef7c5169173253d4cdfa840bce](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/13614/diffs?commit_id=04f232ed99bb68ef7c5169173253d4cdfa840bce)

Since HOME is set to "/" in setupDisplay() manually the host OS cannot set the environment when launching sddm, resulting in the errant .cache folder. The following patch makes sure sddm obeys any relevant vars set by the host and allows OS maintainers to prevent cache folders from being created in /. It also removes setting of SHELL and PATH since these are already configured by the system.
```
diff --git a/src/daemon/XorgDisplayServer.cpp b/src/daemon/XorgDisplayServer.cpp
index 5101cfb..3987cfe 100644
--- a/src/daemon/XorgDisplayServer.cpp
+++ b/src/daemon/XorgDisplayServer.cpp
@@ -262,12 +262,9 @@ namespace SDDM {
         QProcess *displayScript = new QProcess();
 
         // set process environment
-        QProcessEnvironment env;
+        QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
         env.insert(QStringLiteral("DISPLAY"), m_display);
-        env.insert(QStringLiteral("HOME"), QStringLiteral("/"));
-        env.insert(QStringLiteral("PATH"), mainConfig.Users.DefaultPath.get());
         env.insert(QStringLiteral("XAUTHORITY"), m_xauth.authPath());
-        env.insert(QStringLiteral("SHELL"), QStringLiteral("/bin/sh"));
         env.insert(QStringLiteral("XCURSOR_THEME"), mainConfig.Theme.CursorTheme.get());
         QString xcursorSize = mainConfig.Theme.CursorSize.get();
         if (!xcursorSize.isEmpty())
```

This also matches how the environment is set in XorgDisplayServer::start(). Tested on Slackware-current. With this patch we are able to define HOME=/root and make sure the .cache folder is created in an appropriate location. 

Is it possible to get this patch included in the next release of sddm? I'm not sure what your dev pipeline is so let me know if there is anything else I can do to get this included. Thanks! :)

pithium:
This fixes #2079 

XorgDisplayServer::setupDisplay()  is ignoring environment variables set by the host system. This results in sddm creating a /.cache folder on systems with versions of mesa that look for HOME. 

This changes sddm to respect the HOME, PATH, and SHELL variables set by the platform.

### DIFF
--- a/src/daemon/XorgDisplayServer.cpp
+++ b/src/daemon/XorgDisplayServer.cpp
@@ -265,12 +265,9 @@ namespace SDDM {
                       xcursorSize = mainConfig.Theme.CursorSize.get();
 
         // set process environment
-        QProcessEnvironment env;
+        QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
         env.insert(QStringLiteral("DISPLAY"), m_display);
-        env.insert(QStringLiteral("HOME"), QStringLiteral("/"));
-        env.insert(QStringLiteral("PATH"), mainConfig.Users.DefaultPath.get());
         env.insert(QStringLiteral("XAUTHORITY"), m_xauth.authPath());
-        env.insert(QStringLiteral("SHELL"), QStringLiteral("/bin/sh"));
         if (!xcursorTheme.isEmpty())
             env.insert(QStringLiteral("XCURSOR_THEME"), xcursorTheme);
         if (!xcursorSize.isEmpty())


src/daemon/XorgDisplayServer.cpp
```diff
@@ -265,12 +265,9 @@ namespace SDDM {
                       xcursorSize = mainConfig.Theme.CursorSize.get();
 
         // set process environment
-        QProcessEnvironment env;
+        QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
         env.insert(QStringLiteral("DISPLAY"), m_display);
-        env.insert(QStringLiteral("HOME"), QStringLiteral("/"));
-        env.insert(QStringLiteral("PATH"), mainConfig.Users.DefaultPath.get());
```
Vogtinator, line 271: Might still need this

pithium: The only problem I have with that is HOME, PATH, and SHELL are all variables that should be defined by the system to maintain consistency among all applications. Hardcoding these in sddm makes it problematic for distributions that need to be more flexible. If for any reason /bin/sh does not exist, sddm will break and there is no way for the package maintainer to fix it without patching the source. 

That said, this change could break things for downstream package maintainers who are used to relying on sddm to set this. The fix for them would be simple since this patch imports the environment before adding those needed by sddm. 

HOME is the only one causing pain right now. I removed PATH and SHELL since I considered it best practice to import the environment and assume those have already been set.

Vogtinator: sddm.conf has an option to set `PATH`, so if this is removed that will break some configurations.

PR #2082, opened 2025-04-29
Fix sddm stuck on boot with black screen and allow reducing Xorg verbosity level in rootless mode

nyx191:
It appears that my system boots slower and slower over years without additional software :( So that bug happened for me rarely at first. But after recent minor qt6 libs upgrade I finally got repro in 100% cases (system won't boot into GUI).

### DIFF
--- a/src/daemon/XorgDisplayServer.cpp
+++ b/src/daemon/XorgDisplayServer.cpp
@@ -285,7 +285,7 @@ namespace SDDM {
         connect(setCursor, QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished), setCursor, &QProcess::deleteLater);
 
         // wait for finished
-        if (!setCursor->waitForFinished(1000)) {
+        if (!setCursor->waitForFinished(5000)) {
             qWarning() << "Could not setup default cursor";
             setCursor->kill();
         }
@@ -302,7 +302,7 @@ namespace SDDM {
                 xrdbProcess.write(QStringLiteral("Xcursor.size: %1\n").arg(xcursorSize).toUtf8());
 
             xrdbProcess.closeWriteChannel();
-            if (!xrdbProcess.waitForFinished(1000)) {
+            if (!xrdbProcess.waitForFinished(5000)) {
                 qDebug() << "Could not set Xcursor resources" << xrdbProcess.error();
                 xrdbProcess.kill();
             }
--- a/src/daemon/XorgDisplayServer.cpp
+++ b/src/daemon/XorgDisplayServer.cpp
@@ -243,13 +243,12 @@ namespace SDDM {
         const auto program = displayStopCommand.takeFirst();
         displayStopScript->start(program, displayStopCommand);
 
+        // delete displayStopScript on finish
+        connect(displayStopScript, QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished), displayStopScript, &QProcess::deleteLater);
+
         // wait for finished
         if (!displayStopScript->waitForFinished(5000))
             displayStopScript->kill();
-
-        // clean up the script process
-        displayStopScript->deleteLater();
-        displayStopScript = nullptr;
 
         // emit signal
         emit stopped();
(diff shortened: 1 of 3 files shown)


Vogtinator:
`xsetroot` is really just there to set the default cursor, it won't have any other negative effects if it fails.

I'm afraid your change is a placebo at most.

nyx191:
> `xsetroot` is really just there to set the default cursor, it won't have any other negative effects if it fails.

On my machine boot always fail without this commit and always succeed with. Tried for about ~10 times in both cases.
Failed case looks like:
```
22:56:17 sddm-helper-start-x11user[581]: Setting default cursor...
22:56:17 sddm-greeter-qt6[596]: High-DPI autoscaling Enabled
22:56:18 sddm-helper-start-x11user[581]: Could not setup default cursor
22:56:18 sddm-helper-start-x11user[581]: Running display setup script: /usr/share/sddm/scripts/Xsetup
22:56:18 sddm-helper-start-x11user[581]: QProcess: Destroyed while process ("xsetroot") is still running.
22:56:18 sddm-helper-start-x11user[581]: quitting helper-start-x11
22:56:18 sddm-helper-start-x11user[581]: Stopping server...
```

> I'm afraid your change is a placebo at most.

Interesting, maybe there is another problem. Will try to remove xsetroot call whatsoever.

Vogtinator:
> 22:56:18 sddm-helper-start-x11user[581]: QProcess: Destroyed while process ("xsetroot") is still running.

That's a genuine bug, it's missing a `waitForFinished()` after `kill()`. Also harmless though.

nyx191:
Removed the following code block
https://github.com/sddm/sddm/blob/42e88b70c3e558495d07d29d346664301da6e974/src/helper/xorguserhelper.cpp#L216-L223
and yep, it works flawlessly.

Built the original source using fresh dependencies as specified in debian/control and it still fails with exact same log lines as above. _Somehow_ xcursor subprocess timeout causes helper-start-x11 to quit and xserver to stop.

Will try to handle timeout gracefully as you suggest, thanks!

nyx191:
@Vogtinator waiting for xsetroot to terminate and then delete QProcess indeed helps. Weird, I guess its qt6 library bug, but we should not get those warnings anyway. I still think that raising timeout for xsetroot to 5 s is nice since this eliminates yet another warning in log on slow systems and does not affect fast cases. Also it should not make slow systems much slower :)

Added patch to wait for processes correctly in situations I was able to find.

Vogtinator:
> @Vogtinator waiting for xsetroot to terminate and then delete QProcess indeed helps

I wonder why. Maybe it's some race condition that makes the event loop think it's no longer needed. Not sure how to debug this further.

src/daemon/XorgDisplayServer.cpp
```diff
@@ -302,9 +301,10 @@ namespace SDDM {
                 xrdbProcess.write(QStringLiteral("Xcursor.size: %1\n").arg(xcursorSize).toUtf8());
 
             xrdbProcess.closeWriteChannel();
-            if (!xrdbProcess.waitForFinished(1000)) {
+            if (!xrdbProcess.waitForFinished(5000)) {
                 qDebug() << "Could not set Xcursor resources" << xrdbProcess.error();
                 xrdbProcess.kill();
+                xrdbProcess.waitForFinished(25000);
```
Vogtinator: No timeout here please

Vogtinator:
Actually, I see why. This is using the Xorg user helper which for some reason quits on any failure, even timeouts: https://github.com/sddm/sddm/blob/42e88b70c3e558495d07d29d346664301da6e974/src/helper/xorguserhelper.cpp#L104

Can you port the `xsetroot` call to plain `QProcess` like used for the `xrdb` invocation below?

Vogtinator (changes requested):

nyx191:
My bad, I thought `startProcess()` was standard Qt function. Clearly `quit()` is the root cause, also checked with logs. Btw, waiting for setCursor after kill
```
setCursor->kill();
setCursor->waitForFinished(-1);
```
prevents `QCoreApplication::instance()->quit()` from doing something, this is vaguely conformant to `quit()` documentation. That is why above I was reporting that wait helps.

Vogtinator (approved):

Issue #2003, opened 2024-10-22
A notebook with dual graphics cards connected to an external monitor cannot log in to the system through sddm


lingsixuan:
On a dual-GPU laptop with multiple monitors connected, it is impossible to log into the system using SDDM after booting. Specifically, after entering the correct password and pressing Enter, SDDM remains stuck in the state of checking the password, without showing the welcome screen. Additionally, the sleep and other buttons at the bottom of the login page are unclickable. However, if the external monitor is disconnected before booting the system and reconnected after the system starts, logging in works normally.

## Steps to reproduce:

A laptop with an AMD integrated GPU and an NVIDIA dedicated GPU.
Use sudo envycontrol -s nvidia to switch to using the dedicated GPU for output (because the laptop's AMD integrated GPU cannot output to the external monitor).
Connect the external monitor.
Reboot the system. At this point, the login screen will fail to log in properly.
## Expected behavior:
The login screen should allow logging into the system normally.

## my device
The operating system is arch linux
Using kde plasma 6.2.1,
sddm version 0.21.0-4
With amd integrated graphics and nvidia discrete graphics
Has a type-c external screen

packages are kept up to date

## Below is the SDDM log information:

~~~
-- Boot 3f5d30a38c7844a0a35537e21c4f8d1c --
10月 23 01:19:32 ZHSS systemd[1]: Started Simple Desktop Display Manager.
10月 23 01:19:32 ZHSS sddm[838]: Initializing...
10月 23 01:19:32 ZHSS sddm[838]: Starting...
10月 23 01:19:32 ZHSS sddm[838]: Logind interface found
10月 23 01:19:32 ZHSS sddm[838]: Adding new display...
10月 23 01:19:32 ZHSS sddm[838]: Loaded empty theme configuration
10月 23 01:19:32 ZHSS sddm[838]: Xauthority path: "/run/sddm/xauth_cGFBHp"
10月 23 01:19:32 ZHSS sddm[838]: Using VT 2
10月 23 01:19:32 ZHSS sddm[838]: Display server starting...
10月 23 01:19:32 ZHSS sddm[838]: Writing cookie to "/run/sddm/xauth_cGFBHp"
10月 23 01:19:32 ZHSS sddm[838]: Running: /usr/bin/X -nolisten tcp -background none -seat seat0 vt2 -auth /run/sddm/xauth_cGFBHp -noreset -displayfd 16
10月 23 01:19:33 ZHSS sddm[838]: Setting default cursor
[...]

PR #2103, opened 2025-07-29
fix: restart greeter when helper is in a wrong state

### LINKED ISSUE #1806: SDDM crashes on boot
When my system boots the SDDM screen briefly appears for a second or so and then I get dumped to a terminal login instead.

Running journalctl to look at the SDDM logs I see:
```
Sep 26 19:30:17 opensuse-1 sddm[13655]: Session started true

Sep 26 19:30:18 opensuse-1 sddm[13655]: Auth: sddm-helper exited successfully

Sep 26 19:30:18 opensuse-1 sddm[13655]: Greeter stopped. SDDM::Auth::HELPER_SUCCESS

Sep 30 14:02:49 opensuse-1 sddm[13655]: Authentication error: SDDM::Auth::ERROR_INTERNAL "Process crashed"

Sep 30 14:02:49 opensuse-1 sddm[13655]: Auth: sddm-helper (--socket /tmp/sddm-auth-4805ecf9-4fe3-48e8-b08a-b96cf0012418 --id 1 --start /usr/bin/startplasma-x11 --user testuser --display-server /usr/bin/X -nolisten tcp -background none -seat seat0 -noreset -keeptty -novtswitch -verbose 3) crashed (exit code 1)

Sep 30 14:02:49 opensuse-1 sddm[13655]: Authentication error: SDDM::Auth::ERROR_INTERNAL "Process crashed"

Sep 30 14:02:49 opensuse-1 sddm[13655]: Auth: sddm-helper exited with 1

Sep 30 14:02:50 opensuse-1 systemd[1]: Stopping Simple Desktop Display Manager...

Sep 30 14:02:50 opensuse-1 sddm[13655]: Signal received: SIGTERM

Sep 30 14:02:50 opensuse-1 sddm[13655]: Socket server stopping...

Sep 30 14:02:50 opensuse-1 sddm[13655]: Socket server stopped.
```

If I log in from the terminal and do a systemctl restart sddm then SDDM fires up and this time does not crash.

I'm running OpenSUSE tumbleweed on an HP OMEN Laptop 17-ck2xxx (Intel and NVidia cards with nvidia proprietary drivers installed).

Kernel version is 6.5.4-1-default

### LINKED ISSUE #1908: SDDM Crashes on loginctl terminate-user $USER
Currently whenever a users session is ended via `loginctl terminate-user` SDDM crashes to black screen and fails to respond
Logs from journalctl:
`Apr 06 21:57:13 leviathan sddm[727]: Authentication error: SDDM::Auth::ERROR_INTERNAL "Process crashed"`
`Apr 06 21:57:13 leviathan sddm[727]: Auth: sddm-helper (--socket /tmp/sddm-auth-a66c3ad0-1b7a-4dff-a3fc-10be7eb8c41a --id 9 --start /usr/bin/zsh -l -c sway --user haden) crashed (exit code 1)`
`Apr 06 21:57:13 leviathan sddm[727]: Authentication error: SDDM::Auth::ERROR_INTERNAL "Process crashed"`
`Apr 06 21:57:13 leviathan sddm[727]: Auth: sddm-helper exited with 1`
Currently running `sddm 0.21.0-4` on Arch Linux

### LINKED ISSUE #2027: sddm-helper crashes with system freeze when ~/.profile exits with error
So i was having issue with sddm not working and i can't find out what was the issue.
After using Ly, i saw error from ~/.profile appears on start. And when i tested it, ~/.profile had removed script command and after removing it got fixed

# Expected behaviour
 when sddm-helper crashes, it should restart sddm instead of freezing system and display error. such as error occurred when initialising .profile file

# Related issue
https://github.com/sddm/sddm/issues/1806 - i think this is same, issuer claims that it's fixed but i think there should be sddm-helper error reporting at sddm display manager

https://github.com/sddm/sddm/issues/2003 - same logs like mine, tries to start session and then exit 
https://github.com/sddm/sddm/issues/1908

ItsShamed:
This PR fixes a logic bug that prevents SDDM from restarting the greeter when either the session exits with code 1, or the SDDM helper gets signal-killed.

### Helper side

If the helper gets killed by a signal (currently handled are `SIGTERM` and `SIGHUP`), exit with `Auth::HELPER_OTHER_ERROR`. This will prevent the helper from exiting with code `Auth::HELPER_AUTH_ERROR`, which is wrong.
If the session exits with a bad code (!= 0), exit with `Auth::HELPER_SESSION_ERROR`. This will prevent the helper from exiting `Auth::HELPER_AUTH_ERROR` if the session deliberately ends with code 1.

### Daemon side

**DO NOT** cast the `exitCode` passed by `QProcess::finished` as a `Auth::HelperStatus` if the `exitStatus` is not `QProcess::NormalExit`. On Linux `QProcess::finished` is like a wrapper around `waitid(2)` which will put the signal number in `exitCode` if it got killed by one.
The [Qt Docs] even advises against doing it:

> exitCode is the exit code of the process **(only valid for normal exits)**

[Qt Docs]: https://doc.qt.io/qt-6/qprocess.html#finished

- Closes #1806 
- Closes #1908
- Closes #2027

For issues like #2003, this will result in a "restart loop" (which is essentially the same original symptom if the greeter fails to start in the first place).

Yes for the other session related issues, these are not SDDM related issues per se, but I think it's better UX to restart the greeter rather than leaving the user on a blank tty if a session fails.

### DIFF
--- a/src/helper/HelperApp.cpp
+++ b/src/helper/HelperApp.cpp
@@ -53,8 +53,13 @@ namespace SDDM {
             , m_socket(new QLocalSocket(this)) {
         qInstallMessageHandler(HelperMessageHandler);
         SignalHandler *s = new SignalHandler(this);
+        s->addCustomSignal(SIGHUP);
         QObject::connect(s, &SignalHandler::sigtermReceived, m_session, [] {
-            QCoreApplication::instance()->exit(-1);
+            QCoreApplication::instance()->exit(Auth::HELPER_OTHER_ERROR);
+        });
+
+        QObject::connect(s, &SignalHandler::customSignalReceived, m_session, [](int) {
+            QCoreApplication::instance()->exit(Auth::HELPER_OTHER_ERROR);
         });
 
         QTimer::singleShot(0, this, SLOT(setUp()));
--- a/src/helper/HelperApp.cpp
+++ b/src/helper/HelperApp.cpp
@@ -206,7 +206,11 @@ namespace SDDM {
             }
             else
                 exit(Auth::HELPER_SUCCESS);
+            return;
         }
+
+        Q_ASSERT_X(exitCode != 0, "HelperApp::sessionFinished", "Crashing with 0 is impossible");
+
         qWarning("Session crashed (killed by signal %d).", exitCode);
         exit(Auth::HELPER_SESSION_ERROR);
     }
@@ -298,7 +302,9 @@ namespace SDDM {
     HelperApp::~HelperApp() {
         Q_ASSERT(getuid() == 0);
 
-        m_session->stop();
+        // Avoid re-emitting QProcess::finished if session is not running
+        if (m_session->state() != QProcess::NotRunning)
+            m_session->stop();
         m_backend->closeSession();
 
         // write logout to utmp/wtmp
--- a/src/helper/HelperApp.cpp
+++ b/src/helper/HelperApp.cpp
@@ -302,7 +302,8 @@ namespace SDDM {
     HelperApp::~HelperApp() {
         Q_ASSERT(getuid() == 0);
 
-        // Avoid re-emitting QProcess::finished if session is not running
+        // Avoid calls to sessionFinished when exiting
+        disconnect(m_session, QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished), this, &HelperApp::sessionFinished);
         if (m_session->state() != QProcess::NotRunning)
             m_session->stop();
         m_backend->closeSession();
--- a/src/helper/HelperApp.cpp
+++ b/src/helper/HelperApp.cpp
@@ -73,7 +73,7 @@ namespace SDDM {
         if ((pos = args.indexOf(QStringLiteral("--socket"))) >= 0) {
             if (pos >= args.length() - 1) {
                 qCritical() << "This application is not supposed to be executed manually";
-                exit(Auth::HELPER_OTHER_ERROR);
+                QCoreApplication::exit(Auth::HELPER_OTHER_ERROR);
                 return;
             }
             server = args[pos + 1];
@@ -82,7 +82,7 @@ namespace SDDM {
         if ((pos = args.indexOf(QStringLiteral("--id"))) >= 0) {
             if (pos >= args.length() - 1) {
                 qCritical() << "This application is not supposed to be executed manually";
-                exit(Auth::HELPER_OTHER_ERROR);
+                QCoreApplication::exit(Auth::HELPER_OTHER_ERROR);
                 return;
             }
             m_id = QString(args[pos + 1]).toLongLong();
@@ -91,7 +91,7 @@ namespace SDDM {
         if ((pos = args.indexOf(QStringLiteral("--start"))) >= 0) {
             if (pos >= args.length() - 1) {
                 qCritical() << "This application is not supposed to be executed manually";
-                exit(Auth::HELPER_OTHER_ERROR);
+                QCoreApplication::exit(Auth::HELPER_OTHER_ERROR);
                 return;
             }
             m_session->setPath(args[pos + 1]);
@@ -100,7 +100,7 @@ namespace SDDM {
         if ((pos = args.indexOf(QStringLiteral("--user"))) >= 0) {
             if (pos >= args.length() - 1) {
                 qCritical() << "This application is not supposed to be executed manually";
-                exit(Auth::HELPER_OTHER_ERROR);
+                QCoreApplication::exit(Auth::HELPER_OTHER_ERROR);
                 return;
             }
             m_user = args[pos + 1];
@@ -109,7 +109,7 @@ namespace SDDM {
         if ((pos = args.indexOf(QStringLiteral("--display-server"))) >= 0) {
             if (pos >= args.length() - 1) {
                 qCritical() << "This application is not supposed to be executed manually";
-                exit(Auth::HELPER_OTHER_ERROR);
+                QCoreApplication::exit(Auth::HELPER_OTHER_ERROR);
                 return;
             }
             m_session->setDisplayServerCommand(args[pos + 1]);
@@ -126,7 +126,7 @@ namespace SDDM {
 
         if (server.isEmpty() || m_id <= 0) {
             qCritical() << "This application is not supposed to be executed manually";
-            exit(Auth::HELPER_OTHER_ERROR);
+            QCoreApplication::exit(Auth::HELPER_OTHER_ERROR);
             return;
         }
 
@@ -151,7 +151,7 @@ namespace SDDM {
             const QString vt = env.value(QStringLiteral("XDG_VTNR"));
             utmpLogin(vt, displayId, m_user, 0, false);
 
-            exit(Auth::HELPER_AUTH_ERROR);
+            QCoreApplication::exit(Auth::HELPER_AUTH_ERROR);
             return;
         }
 
@@ -165,7 +165,7 @@ namespace SDDM {
             const QString vt = env.value(QStringLiteral("XDG_VTNR"));
             utmpLogin(vt, displayId, m_user, 0, false);
 
-            exit(Auth::HELPER_AUTH_ERROR);
+            QCoreApplication::exit(Auth::HELPER_AUTH_ERROR);
             return;
         }
 
@@ -178,7 +178,7 @@ namespace SDDM {
 
             if (!m_backend->openSession()) {
                 sessionOpened(false);
-                exit(Auth::HELPER_SESSION_ERROR);
+                QCoreApplication::exit(Auth::HELPER_SESSION_ERROR);
                 return;
             }
 
@@ -194,7 +194,7 @@ namespace SDDM {
             }
         }
         else
-            exit(Auth::HELPER_SUCCESS);
+            QCoreApplication::exit(Auth::HELPER_SUCCESS);
         return;
     }
 
@@ -202,17 +202,17 @@ namespace SDDM {
         if (exitStatus == QProcess::NormalExit) {
             if (exitCode != 0) {
                 qWarning("Session crashed (exit code %d).", exitCode);
-                exit(Auth::HELPER_SESSION_ERROR);
+                QCoreApplication::exit(Auth::HELPER_SESSION_ERROR);
             }
             else
-                exit(Auth::HELPER_SUCCESS);
+                QCoreApplication::exit(Auth::HELPER_SUCCESS);
             return;
         }
 
         Q_ASSERT_X(exitCode != 0, "HelperApp::sessionFinished", "Crashing with 0 is impossible");
 
         qWarning("Session crashed (killed by signal %d).", exitCode);
-        exit(Auth::HELPER_SESSION_ERROR);
+        QCoreApplication::exit(Auth::HELPER_SESSION_ERROR);
     }
 
     void HelperApp::info(const QString& message, Auth::Info type) {
--- a/src/helper/HelperApp.cpp
+++ b/src/helper/HelperApp.cpp
@@ -199,16 +199,16 @@ namespace SDDM {
     }
 
     void HelperApp::sessionFinished(int exitCode, QProcess::ExitStatus exitStatus) {
-        if (exitStatus == QProcess::CrashExit) {
-            qWarning("Session crashed (killed by signal %d).", exitCode);
-            exit(Auth::HELPER_SESSION_ERROR);
+        if (exitStatus == QProcess::NormalExit) {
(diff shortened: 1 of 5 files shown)


Vogtinator:
There's actually one more case of this: https://github.com/sddm/sddm/blob/08a4a77073ffc5566baa54fa5345f20eea999648/src/helper/UserSession.cpp#L50

That connects `QProcess::finished(int exitCode, QProcess::ExitStatus exitStatus)` to `UserSession::finished(int)`, losing `exitStatus` and making `exitCode` potentially invalid. Could you fix that up as well?

ItsShamed:
Should be good now

https://github.com/sddm/sddm/blob/3fd079f1e9bec2b124b61ad2a0d3bd82963d3a91/src/helper/UserSession.cpp#L159

For this line, I think this is fine, as we explicitly specify an enum, and `HelperApp` filters exit codes anyway.
If this is still bothering, I can change to a straight `exit(2)` instead

```diff
diff --git a/src/helper/UserSession.cpp b/src/helper/UserSession.cpp
index 5fd378a..eb6b598 100644
--- a/src/helper/UserSession.cpp
+++ b/src/helper/UserSession.cpp
@@ -156,7 +156,7 @@ namespace SDDM {
                 }
             }
         } else {
-            Q_EMIT finished(Auth::HELPER_OTHER_ERROR, QProcess::NormalExit);
+            exit(Auth::HELPER_OTHER_ERROR);
         }
     }
 
```

src/helper/HelperApp.cpp
```diff
@@ -193,8 +199,17 @@ namespace SDDM {
...
     }
 
-    void HelperApp::sessionFinished(int status) {
-        exit(status);
+    void HelperApp::sessionFinished(int exitCode, QProcess::ExitStatus exitStatus) {
+        if (exitStatus == QProcess::CrashExit) {
+            qWarning("Session crashed (killed by signal %d).", exitCode);
+            exit(Auth::HELPER_SESSION_ERROR);
+        }
+        else if (exitCode != 0) {
```
Vogtinator: Shouldn't this check for `NormalExit`?

ItsShamed: The first `if` branch already checks for `CrashExit` instead, if we run into this branch then we should be in the case of a `NormalExit` (as there are only two enum entries in `QProcess::ExitStatus`).
I doubt this enum would change in QtBase for future releases.

Vogtinator: It's still weird to read. Especially if it says that `exitCode` is only valid for `NormalExit`, the code should check for `NormalExit` explicitly.

ItsShamed: I'll rearrange a bit to make it clearer then

src/helper/HelperApp.cpp
```diff
@@ -34,6 +34,7 @@
 #include <QtNetwork/QLocalSocket>
 
 #include <iostream>
+#include <qobject.h>
```
Vogtinator: Wrong kind of include. What is this for? Maybe `#include <QtCore/QObject>` but that should be there already.

Vogtinator (commented):
LGTM otherwise. FWICT this is effectively a noop in the "happy path", so not that likely to cause regressions.

src/helper/HelperApp.cpp
ItsShamed: oops, probably some LSP shenanigans that I forgot to monitor

ItsShamed:
Actually, hold off merging this one. I might've actually introduced new issues where SDDM would restart for seemingly no reason (opposite problem). I'm poking around to see what the root cause might be.

ItsShamed:
Okay, so these are the logs that made me realize there is a problem.
image

A process cannot "crash" with `si_status = 0`, since UNIX signals start at 1 with `SIGHUP`.

I went back on Qt Base to see the behavior of waiting for a child process and
https://github.com/qt/qtbase/blob/46dcf3abc2ab4c6fbf14b5e108c585ba65060806/src/corelib/io/qprocess_unix.cpp#L1312-L1316
is absolute ~~dogshit~~ nightmare. There are no checks of the return code and `errno` except for their case of waiting for `EINTR` (`do {} while (ret = -1 && errno = EINTR)`). Which means that this is total undefined behavior if `errno` is anything other than `EINTR` (for example `ECHILD`), and can lead to `QProcess::finished` passing around undefined values.

We can however prevent this by disconnecting the `QProcess::finished` signal when we receive the first one, since those undefined behaviors are more likely to occur after the child initially exited.

Vogtinator:
> I went back on Qt Base to see the behavior of waiting for a child process and
https://github.com/qt/qtbase/blob/46dcf3abc2ab4c6fbf14b5e108c585ba65060806/src/corelib/io/qprocess_unix.cpp#L1312-L1316
is absolute dogshit nightmare. There are no checks of the return code and errno except for their case of waiting for EINTR (do {} while (ret = -1 && errno = EINTR)). Which means that this is total undefined behavior if errno is anything other than EINTR (for example ECHILD), and can lead to QProcess::finished passing around undefined values.

If you have a concrete case where this can fail, please report this to Qt upstream.

src/helper/HelperApp.cpp
```diff
@@ -193,8 +198,28 @@ namespace SDDM {
         return;
     }
 
-    void HelperApp::sessionFinished(int status) {
-        exit(status);
+    void HelperApp::sessionFinished(int exitCode, QProcess::ExitStatus exitStatus) {
+        disconnect(m_session, QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished), this, &HelperApp::sessionFinished);
```
Vogtinator: This can never be reached unless something went really wrong: All paths in this method lead to `exit`, i.e. it will never return.

If `exit` somehow spawns an event loop again (exit handlers?), that's broken and will cause many "fun" issues due to accidental reentrancy.

ItsShamed: That's what I thought too, all paths lead to an `exit`, but…

```diff
diff --git a/src/helper/HelperApp.cpp b/src/helper/HelperApp.cpp
index f7aa306..6860535 100644
--- a/src/helper/HelperApp.cpp
+++ b/src/helper/HelperApp.cpp
@@ -199,7 +199,8 @@ namespace SDDM {
     }
 
     void HelperApp::sessionFinished(int exitCode, QProcess::ExitStatus exitStatus) {
-        disconnect(m_session, QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished), this, &HelperApp::sessionFinished);
+        static int calls = 0;
+        qWarning("Call #%d: HelperApp::sessionFinished(%d, %d)", ++calls, exitCode, exitStatus);
         if (exitStatus == QProcess::NormalExit) {
             if (exitCode != 0) {
                 qWarning("Session crashed (exit code %d).", exitCode);
```

image

… `HelperApp::sessionFinished` is in fact called multiples times here…

And that's most likely what happened [to me, above](https://github.com/sddm/sddm/pull/2103#issuecomment-3136871058) too.

I don't think this is in our control, again, the code in Qt Base is very sketchy, and I don't trust it to do things correctly. So we might as well take precautions to prevent such cases.

I'll try filling a bug report upstream.

ItsShamed: Oh well: https://bugreports.qt.io/browse/QTBUG-119081
At this point, I don't have hopes this will get fixed whatsoever…
The reasoning is kinda stupid imho…

Vogtinator: > The reasoning is kinda stupid imho…

The reasoning is:

> The fcntl can only fail if the file descriptor is invalid. That means QProcess passed a bad file descriptor, so the bug would be in QProcess.

i.e. if this actually happens, it's a bug in `QProcess` and you can report it.

Vogtinator: > So we might as well take precautions to prevent such cases.

IMO the root cause needs to be found and it's not in `QProcessPrivate::waitForDeadChild`.

ItsShamed: > The fcntl can only fail if the file descriptor is invalid. That means QProcess passed a bad file descriptor, so the bug would be in QProcess.

Leaving out the fact that `fcntl` is not the only reason that `forkfd_wait` can return -1: https://github.com/qt/qtbase/blob/46dcf3abc2ab4c6fbf14b5e108c585ba65060806/src/3rdparty/forkfd/forkfd_linux.c#L224-L229, and that maybe an actual `waitid(2)` syscall with an expired PID could happen.

> IMO the root cause needs to be found and it's not in `QProcessPrivate::waitForDeadChild`.

You might be right, but it still feels wrong to not check errors because of some assumptions…

Anyway, I can't reproduce the "crash with 0" bug anymore, but I'm still not confident since I still get double `sessionFinished` calls despite `exit`s…

Vogtinator: > Anyway, I can't reproduce the "crash with 0" bug anymore, but I'm still not confident since I still get double sessionFinished calls despite exits…

You can try to trace why each of those calls happen.

ItsShamed: So I just realized the `exit` we've been calling, is not the `exit(2)` syscall, but `QCoreApplication::exit` which does in fact returns, and I was able to reproduce the "crash with 0" bug. I concluded this had nothing to do with `QProcess`.

Sorry for this nothingburger, here is what I understood what happened:
- When `UserSession` `QProcessed::finished`, we catch it in `HelperApp::sessionFinished`
- If it's a `QProcess::NormalExit`, call **`QCoreApplication::exit`**:
  - This only signals the event loop that we requested an exit, and the return code we exit with
  - It actually returns in order to process more events
- This means we accidentally fall through our second case of `QProcess::CrashExit`, believing we stumbled into one:
  - By calling this `exit`, this sets the event loop return code to 3
- Our `HelperApp` then process all the remaining events
- Finally, the event loop returns to the main thread
- When `main()` reaches out of scope, it calls our destructor `~HelperApp()`
- Which in turns calls `UserSession::stop()`.
- However, the `UserSession` is already `QProcess::NotRunning` by this stage, so it emits another `QProcess::finished(Auth::HELPER_SESSION_ERROR, QProcess::NormalExit)`.
  - `HelperApp::sessionFinshed` gets called again…

<details>
<summary>Sequence diagram</summary>

```mermaid
sequenceDiagram
    participant main as ::main
    participant HelperApp as SDDM::HelperApp
    participant Backend as SDDM::Backend
    participant UserSession as SDDM::UserSession

    main->>HelperApp: exec()
    Note over HelperApp: setUp() and doAuth()
    HelperApp->>Backend: openSession()
    Backend->>UserSession: start()

    Note over UserSession: Session goes on…
    UserSession->>HelperApp: QProcess::finished()
    Note over HelperApp: sessionFinished()
    opt QProcess::NormalExit
        alt exitCode not 0
            HelperApp->>HelperApp: exit(Auth::HELPER_SESSION_ERROR)
        else
            HelperApp->>HelperApp: exit(Auth::HELPER_SUCCES)
        end
    end
    Note right of HelperApp: Those RETURN and fall through (not exit syscall)
    HelperApp->>HelperApp: exit(Auth::HELPER_SESSION_ERROR)
    HelperApp-->HelperApp: Finish processing events
    HelperApp-->>main: Event loop returns with 3
    main->>main: Return 3
    Note over main: At SCOPE FINISH
    main->>HelperApp: ~HelperApp() (call destructor)
    HelperApp->>UserSession: stop()
    Note over UserSession: state() is QProcess::NotRunning
    UserSession->>HelperApp: QProcess::finished(3, 0)
    Note over HelperApp: Repeat sessionFinished() a second time
```
</details>

Should be pretty simple to fix:
- We should either explicitly `return` in `HelperApp::sessionFinished`, or wrap the second case in an `else` branch
- In the destructor, call `m_session->stop()` only if `m_session->state() != QProcess::NotRunning`.

Vogtinator: > So I just realized the exit we've been calling, is not the exit(2) syscall, but QCoreApplication::exit which does in fact return

In that case it should probably be made a bit more explicit by calling `QCoreApplication::exit()` instead...

Vogtinator: > However, the UserSession is already QProcess::NotRunning by this stage, so it emits another QProcess::finished(Auth::HELPER_SESSION_ERROR, QProcess::NormalExit).

That always seemed odd to me, maybe we can just remove it?

ItsShamed: > In that case it should probably be made a bit more explicit by calling `QCoreApplication::exit()` instead...

Done

> That always seemed odd to me, maybe we can just remove it?

Replaced by a `qInfo()` log

basil:
@Vogtinator Friendly note that I tested this patch on Fedora 43 and can confirm it resolves #1908. Looking forward to seeing this PR merged and released!

PR #2119, opened 2025-09-13
Add option to disable autologin after systemd-softreboot

Lama-Thematique:
Hello,

## description
This MR add a option to keep the state of autologin through soft-reboot for `relogin=False`.

## How it  works
According to [freedesktop.org documentation on systemd-soft-reboot.service](https://www.freedesktop.org/software/systemd/man/latest/systemd-soft-reboot.service.html) : 

> The /run/ file system remains mounted and populated and may be used to pass state information between such userspace reboot cycles.

By adding lockfile in /run/sddm i can keep track of the state of autologin through soft-reboot and more (including crashes, and restart of the daemon).
The patch maintains backward compatibility, and has the new behaviour disabled by default.

## Why this patch was created
I found a unexpected way to bypass auth in my system from a "guest" user to my "main" user.

### My setup: 
- When i boot, i'm greeted by grub cryptdisk that ask for my LUKS password to access the /boot partition
- SDDM directly open my session without asking for password with the autologin

### Why i have this setup:
- When I boot I have only 1 password (LUKS in grub cryptdisk, qutologin get rid of the 2nd one)
- When I leave my computer, I have to enter my password to unlock the screen due to `relogin=False`

My intention was that i would always have EXACTLY one password to login to my user (LUKS xor SSDM).

### The problem: 
- I have a second (guest) user for when I lend my PC to a friend
- When connected to the guest user, he can use the terminal to do `systemctl soft-reboot` which reboot without reloading grub, the kernel or the initramfs, bypassing LUKS password
- after soft-reboot i am logged back into my "main" user WITHOUT any password, as the SDDM service is restarted in the process

This allow any user to login to the designated autologin session even when `relogin=False` should have prevented it.

## Other informations
This patch set have been tested in virtual machines before being tried on my laptop for 1 week, I found no issues with it in my testing.

No AIs have been used in the making of this MR (I don't know if it matters for this project).

### DIFF
--- a/src/common/Configuration.h
+++ b/src/common/Configuration.h
@@ -103,7 +103,6 @@ namespace SDDM {
             Entry(User,                QString,     QString(),                                  _S("Username for autologin session"));
             Entry(Session,             QString,     QString(),                                  _S("Name of session file for autologin session (if empty try last logged in)"));
             Entry(Relogin,             bool,        false,                                      _S("Whether sddm should automatically log back into sessions when they exit"));
-            Entry(FirstLoginLockFile,  bool,        false,                                      _S("Whether sddm should use a temporary lock file to keep track of first login (persistent through `systemd soft-reboot`)"));
         );
     );
 
--- a/src/common/Configuration.h
+++ b/src/common/Configuration.h
@@ -103,6 +103,7 @@ namespace SDDM {
             Entry(User,                QString,     QString(),                                  _S("Username for autologin session"));
             Entry(Session,             QString,     QString(),                                  _S("Name of session file for autologin session (if empty try last logged in)"));
             Entry(Relogin,             bool,        false,                                      _S("Whether sddm should automatically log back into sessions when they exit"));
+            Entry(FirstLoginLockFile,  bool,        false,                                      _S("Whether sddm should use a temporary lock file to keep track of first login (persistent through `systemd soft-reboot`)"));
         );
     );
 
--- a/src/daemon/DaemonApp.cpp
+++ b/src/daemon/DaemonApp.cpp
@@ -47,6 +47,9 @@ namespace SDDM {
         // log message
         qDebug() << "Initializing...";
 
+        // setup logfile
+        m_firstLockFile.setFileName(QStringLiteral(RUNTIME_DIR) + QStringLiteral("/firstLock"));
+
         // set testing parameter
         m_testing = (arguments().indexOf(QStringLiteral("--test-mode")) != -1);
 
@@ -119,6 +122,25 @@ namespace SDDM {
     int DaemonApp::newSessionId() {
         return m_lastSessionId++;
     }
+
+    bool DaemonApp::hasLock() const {
+        return m_firstLockFile.exists();
+    };
+
+    bool DaemonApp::getFirst() const {
+        return self->first;
+    };
+
+    void DaemonApp::consumeFirst() {
+        self->first = false;
+        if (hasLock()) {
+            return;
+        }
+        if(!m_firstLockFile.open(QIODevice::WriteOnly)) {
+            qFatal("Failed to create lock file");
+        }
+        m_firstLockFile.close();
+    };
 }
 
 int main(int argc, char **argv) {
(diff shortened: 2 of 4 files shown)


Vogtinator:
> When connected to the guest user, he can use the terminal to do systemctl soft-reboot which reboot without reloading grub, the kernel or the initramfs, bypassing LUKS password

IMO that's the main issue here. guest users should not be able to `systemctl soft-reboot`.

Lama-Thematique:
> IMO that's the main issue here. guest users should not be able to `systemctl soft-reboot`.

While i somewhat agree, as far as i know, in systemd `soft-reboot` and `reboot` requires the exact same autorisation and is given by default to active user for all desktop distros i can think off. 

Splitting the `soft-reboot` and reboot would likelly require systemd changes.

This is in my opinion a much easier and less disruptive change to make in sddm than in systemd.
It also have some added benefits as it prevent abusing the autologin (with `relogin=False`) in case someone find a consitent way to crash/restart the daemon.

src/common/Configuration.h
```diff
@@ -103,6 +103,7 @@ namespace SDDM {
             Entry(User,                QString,     QString(),                                  _S("Username for autologin session"));
             Entry(Session,             QString,     QString(),                                  _S("Name of session file for autologin session (if empty try last logged in)"));
             Entry(Relogin,             bool,        false,                                      _S("Whether sddm should automatically log back into sessions when they exit"));
+            Entry(FirstLoginLockFile,  bool,        false,                                      _S("Whether sddm should use a temporary lock file to keep track of first login (persistent through `systemd soft-reboot`)"));
```
Vogtinator: IMO this doesn't  need to be configurable

src/daemon/DaemonApp.cpp
```diff
@@ -119,6 +122,25 @@ namespace SDDM {
     int DaemonApp::newSessionId() {
         return m_lastSessionId++;
     }
+
+    bool DaemonApp::hasLock() const {
```
Vogtinator: IMO no need for this

src/daemon/DaemonApp.cpp
```diff
@@ -47,6 +47,9 @@ namespace SDDM {
         // log message
         qDebug() << "Initializing...";
 
+        // setup logfile
```
Vogtinator: Initialize `self->first` here?

src/daemon/DaemonApp.cpp
```diff
@@ -47,6 +47,9 @@ namespace SDDM {
...
+        m_firstLockFile.setFileName(QStringLiteral(RUNTIME_DIR) + QStringLiteral("/firstLock"));
```
Vogtinator: IMO name should be more descriptive - why `Lock`?

src/daemon/DaemonApp.h
```diff
@@ -57,6 +61,9 @@ namespace SDDM {
 
         int m_lastSessionId { 0 };
 
+        bool first { true };
+        QFile m_firstLockFile;
```
Vogtinator: No need for this

src/common/Configuration.h
Lama-Thematique: Did it in case the SDDM project had strong backward compatibility requirements, will be removed

src/daemon/DaemonApp.cpp
Lama-Thematique: self->first will be removed with the FirstLoginLockFile option as the existance of tyhe file in the filesystem is the value of the boolean.

src/daemon/DaemonApp.cpp
Lama-Thematique: Error on my part, i wanted to put it as file extension `.lock`, will be renamed to `firstlogin.lock`

according to https://en.wikipedia.org/wiki/File_locking#Lock_files : 
> Shell scripts and other programs often use a strategy similar to the use of file locking: creation of lock files, which are files whose contents are irrelevant (although often one will find the process identifier of the holder of the lock in the file) and whose sole purpose is to signal by their presence that some resource is locked.

PR #2135, opened 2025-11-07
GreeterApp: Use shared contexts like in plasmashell from KDE 6

SgAkErRu:
Fix bug for themes with QtWebEngine usage.

On Qt5 themes it will fix warning in journal: `WebEngineContext used before QtWebEngine::initialize() or OpenGL context creation failed.`

On Qt6 themes it will fix crash with message: `QtWebEngineQuick::initialize() must be called from the Qt gui thread.`

https://bugs.kde.org/show_bug.cgi?id=495758

https://invent.kde.org/plasma/plasma-workspace/-/commit/a23432f218f9a4094f31e31ea1c109464cb5bf04

### DIFF
--- a/src/greeter/GreeterApp.cpp
+++ b/src/greeter/GreeterApp.cpp
@@ -339,6 +339,17 @@ int main(int argc, char **argv)
         qDebug() << "High-DPI autoscaling not Enabled";
     }
 
+    // GreeterApp: Use shared contexts like in plasmashell from KDE 6
+    //
+    // This fixes loading Qt6 themes that using QtWebEngine, which otherwise
+    // will crash with message:
+    // "QtWebEngineQuick::initialize() must be called from the Qt gui thread."
+    //
+    // According to Qt's docs setting the Qt::AA_ShareOpenGLContexts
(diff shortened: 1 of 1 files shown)


Vogtinator:
> Fix bug for themes with QtWebEngine usage.

O.o

Shared contexts are known to cause some issues (that's why they're not the default), not sure if it's a good idea.

SgAkErRu:
> > Fix bug for themes with QtWebEngine usage.
> 
> O.o
> 
> Shared contexts are known to cause some issues (that's why they're not the default), not sure if it's a good idea.

Well, looks like in KDE 6 `plasmashell` and `plasmawindowed` took a chance on this and it didn't seem to lead to any significant problems.

In general, this proposal is due to the fact that there is a custom theme where the use of QtWebEngine is justified by the need to display a web page on the login screen. And without this patch, it just turns out to be a crash greeter with this theme.

Or maybe there is some other way to implement this, if it really can lead to some problems? For example, adding an attribute to the metadata file of the theme stating that it is going to use QtWebEngine. And then optionally enable this flag in Greeter, if this is theme with QtWebEngine that will be displayed in Greeter?

PR #2147, opened 2026-01-08
Updated country and region flags

megaminxwin:
The original flags, taken from fbxkb, were in desperate need of an update. These flags were created by downscaling the images of the flags from Wikipedia to fit in a 60x40 box, and then indexed to 16 colours.

Some notes:
- Georgia (GE), Iraq (IQ), Comoros (KM), Libya (LY), and Syria (SY) have changed their flags completely
- The United Nations (UN) has its colours swapped to reflect the actual UN flag
- The aspect ratio of all the flags is now as close as possible given the size limitations
- The colours are now far more accurate, given the assumption that we don't need to rely on safe colours in 2026

There are still some questions that need to be answered:
- Why these particular regions? There are other countries and languages with their own keyboards
- Why is the UN included?
- Do we need both GB and UK for the United Kingdom?
- Does it make sense to still include the flag of Yugoslavia, given the country no longer exists and none of its successor states use this flag?

However, I believe this is a good starting point.


megaminxwin:
Just noticed (seven months later, wow) that the Georgia flag was completely the wrong size. That's been fixed now.

PR #2159, opened 2026-03-05
Update Romanian translation

sbivol:


data/translations/ro.ts
```diff
@@ -1,73 +1,143 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version='1.0' encoding='utf-8'?>
```
Vogtinator, line 1: Single quotes here looks weird

sbivol: It's how Lokalize has saved the file. 
Shall I change that to double quotes?

Vogtinator: I see it's the same for es and hsb, so let's keep it as-is

Vogtinator (approved):

PR #2166, opened 2026-03-26
sddm.service: conflicts with kmsconvt@ttyX

kdj0c:
To prevent kmscon to start on the same tty as sddm, add kmsconvt to the conflicts list, like getty.

### DIFF
--- a/services/sddm.service.in
+++ b/services/sddm.service.in
@@ -1,8 +1,8 @@
 [Unit]
 Description=Simple Desktop Display Manager
 Documentation=man:sddm(1) man:sddm.conf(5)
-Conflicts=getty@tty${SDDM_INITIAL_VT}.service
-After=systemd-user-sessions.service getty@tty${SDDM_INITIAL_VT}.service plymouth-quit.service systemd-logind.service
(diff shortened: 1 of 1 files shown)


Vogtinator:
I wonder if there's a generic way to achieve this

Vogtinator:
(Semi OT) @Conan-Kudo Looks like dnf changed something and `dnf builddep` is no longer available? Does it need a plugin? Should it be pre-installed?

Vogtinator (approved):

kdj0c:
> I wonder if there's a generic way to achieve this

I opened an issue to systemd-logind about this problem:
https://github.com/systemd/systemd/issues/41087

Hopefully they will find a generic way to solve it.

Conan-Kudo:
> (Semi OT) @Conan-Kudo Looks like dnf changed something and `dnf builddep` is no longer available? Does it need a plugin? Should it be pre-installed?

For Fedora and openSUSE, we need to install `dnf5-command(builddep)` (which is provided by `dnf5-plugins`) and for CentOS, we need `dnf-command(builddep)` (which is provided by `dnf-plugins-core`).

PR #2171, opened 2026-04-14
Fix #2157: Enable keyboard input on all monitors

### LINKED ISSUE #2157: Login screen appears on each monitor, but only one of them accepts input.
(I am using KDE Plasma on Ubuntu, with sddm login screen)

Steps:
- Connect multiple monitors. (I have laptop + 2 external monitors)
- Suspend, then wake up. (or some other thing that causes login screen to appear)

Expected:
- Login screen with password field appears on each monitor. Either separate, or mirrored.
- I can choose _any_ of the monitors to type the password.
- If mirrored, the same (masked) password will appear on each monitor.
- If separate, the password field on the other monitors will appear as empty.

Actual:
- A separate login screen appears on each monitor, each with its own password field.
- Only one of them accepts keyboard input. (I guess it somehow determines which one is the "primary" screen)

### Notes / Motivation
This is especially bad if the "primary" display happens to be broken.

It happened to me in the past when my laptop screen was defunct and I connected an external monitor.
Depending on the context (that is, stuff that happened before), the system would decide that the laptop screen is the primary display, and I had to enter the password "blindly".
(the main issue is not typing the password but making sure that the password field is focused for input.)

tambiyusuf:
Fixes #2157

This PR enables keyboard input on all monitors in multi-screen setups.

**Problem:** Login screens appear on each monitor but only the primary screen accepts keyboard input. This causes serious usability issues, especially when the primary display is broken.

**Solution:**
- Auto-activate view when user clicks on non-primary screen
- Add MultiMonitor configuration option (primary-only/all-separate/all-mirrored)

**Testing Note:** 
Tested on Arch Linux (SDDM 0.21.0-6, Qt 5.15.18) - the issue did not reproduce on this setup, likely due to newer Qt version. However, this fix provides:
1. Preventive solution for distributions with older Qt versions
2. Explicit multi-monitor behavior control via configuration
3. No regression on systems where the issue doesn't occur

The fix is backwards compatible and maintains default behavior while allowing users to customize multi-monitor handling.

### DIFF
--- a/src/greeter/GreeterApp.cpp
+++ b/src/greeter/GreeterApp.cpp
@@ -167,6 +167,15 @@ namespace SDDM {
             view->setGeometry(r);
         });
 
+        // Fix for issue #2157: activate view when user clicks on it
+        // This allows keyboard input on any monitor, not just the primary one
+        connect(view, &QQuickView::activeFocusItemChanged, this, [view]() {
+            if (!view->isActive() && view->activeFocusItem()) {
+                qDebug() << "Auto-activating view on" << view->screen()->name() << "due to focus change";
+                view->requestActivate();
+            }
+        });
+
         view->engine()->addImportPath(QStringLiteral(IMPORTS_INSTALL_DIR));
 
         // connect proxy signals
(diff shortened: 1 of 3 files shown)


src/greeter/GreeterApp.cpp
```diff
@@ -167,6 +167,15 @@ namespace SDDM {
             view->setGeometry(r);
         });
 
+        // Fix for issue #2157: activate view when user clicks on it
```
Vogtinator, line 170: Not necessary - this already happens automatically.

src/greeter/GreeterApp.cpp
```diff
@@ -276,8 +285,27 @@ namespace SDDM {
...
+        
+        // Apply multi-monitor configuration
+        const auto multiMonitorMode = mainConfig.MultiMonitor.get();
+        
+        if (multiMonitorMode == MainConfig::MULTI_PRIMARY_ONLY) {
+            // Only show greeter on primary screen
+            qInfo() << "Multi-monitor mode: primary-only - showing greeter only on primary screen";
+            addViewForScreen(QGuiApplication::primaryScreen());
+        }
+        else if (multiMonitorMode == MainConfig::MULTI_ALL_MIRRORED) {
```
Vogtinator, line 297: IMO it does not make sense to mention this anywhere if it's not implemented.

src/greeter/GreeterApp.cpp
```diff
@@ -276,8 +285,27 @@ namespace SDDM {
 
         // Create views
         const QList<QScreen *> screens = qGuiApp->primaryScreen()->virtualSiblings();
-        for (QScreen *screen : screens)
-            addViewForScreen(screen);
+        
+        // Apply multi-monitor configuration
+        const auto multiMonitorMode = mainConfig.MultiMonitor.get();
+        
+        if (multiMonitorMode == MainConfig::MULTI_PRIMARY_ONLY) {
```
Vogtinator, line 292: This is the only change this PR actually implements and IMO it is of very limited use. The primary screen can't even be easily configured ATM.

Vogtinator (changes requested):
The changes don't really make sense in relation to the linked issue.

tambiyusuf:
Thanks for the review! You're right, let me clarify:

 The reporter couldn't input on non-primary monitors. I tested on Arch (SDDM 0.21.0-6, Qt 5.15.18) and couldn't reproduce - both monitors worked fine.

1. Should I close this PR since the auto-activation already works?
2. Is the issue still relevant on older Qt versions/distros?
3. Would a simpler approach (just the primary-only mode) be useful, or should I close this entirely?

I'm open to either improving this or closing if it's not needed.

Vogtinator:
IMO let's close this until there's some reporter feedback that this actually helps.

PR #2185, opened 2026-07-10
Added working flake.nix configuration

rocapp:
- **flake.nix, build.sh from master branch**
- **all tests now pass**
- **fixed version**

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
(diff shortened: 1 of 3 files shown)


Vogtinator:
- I don't think we need this, NixOS already has it packaged and we don't have .dsc, .spec or PKGBUILD for other distros either
- This gives off AI slop vibe
- Qt 5 only
- Has a lock file, that's a liability that needs to be maintained

PR #2193, opened 2026-07-15
Use variable for code simplify in sddm

GermanAizek:


GermanAizek:
@Vogtinator i squash all commits in one

PR #2194, opened 2026-07-15
PamBackend: use const reference variables for code simplify

GermanAizek:


GermanAizek:
Migrate to https://github.com/sddm/sddm/pull/2193

PR #2195, opened 2026-07-15
daemon: use variable for code simplify

GermanAizek:


GermanAizek:
Migrate to https://github.com/sddm/sddm/pull/2193

PR #2196, opened 2026-07-15
Display: use variable for code simplify

GermanAizek:


Vogtinator:
Hm, could you make a single PR with all commits instead of PRs for each commit? IMO that's a bit easier to handle from a review PoV.

GermanAizek:
> Hm, could you make a single PR with all commits instead of PRs for each commit? IMO that's a bit easier to handle from a review PoV.

in many projects, I'm asked to make different commits, that's why I'm used to pushing like this.

GermanAizek:
@Vogtinator squash only code simplify with vars commits?

Vogtinator:
Different commits yes, but a single PR would be nice, as they only simplify the code while not actually changing behaviour.

GermanAizek:
@Vogtinator migrate all commits to https://github.com/sddm/sddm/pull/2193

Issue #2134, opened 2025-11-04
Black screen on logout need to force shutdown the system


iamabdulaziz100:
When I logout of plasma on cachy os the screen goes black and I need to force shutdown the system. I tried other dm's but it only happen in sddm.

**System Information:**

sddm 0.21.0-8
sddm-kcm 6.5.1-1.1
Operating System: CachyOS Linux
KDE Plasma Version: 6.5.1
KDE Frameworks Version: 6.19.0
Qt Version: 6.10.0
Kernel Version: 6.17.6-2-cachyos (64-bit)
Graphics Platform: Wayland
Processors: 8 × AMD Ryzen 5 3500U with Radeon Vega Mobile Gfx
Memory: 8 GiB of RAM (5.7 GiB usable)
Graphics Processor: AMD Radeon Vega 8 Graphics

shunf4:
Possibly a duplicate of #1535

Issue #1535, opened 2022-03-30
Black screen with mouse cursor after logging out


ctrlaltmilk:
Running Arch Linux with KDE Plasma and SDDM 0.19.0.

At boot, SDDM works fine. However, when I log into Plasma then log out, SDDM fails to start and shows only a black screen with a mouse cursor.

The logs from a fresh boot look like this, from adding a new display to removing :0, look like this:
```
Mar 29 16:41:21 nitrogen sddm[529]: Adding new display on vt 1 ...
Mar 29 16:41:21 nitrogen sddm[529]: Loading theme configuration from ""
Mar 29 16:41:21 nitrogen sddm[529]: Display server starting...
Mar 29 16:41:21 nitrogen sddm[529]: Adding cookie to "/var/run/sddm/{d03ee0af-3e21-4ec8-b96c-3ece8d7e260c}"
Mar 29 16:41:21 nitrogen sddm[529]: Running: /usr/bin/X -nolisten tcp -background none -seat seat0 vt1 -auth /var/run/sddm/{d03ee0af-3e21-4ec8-b96c-3ece8d7>
Mar 29 16:41:21 nitrogen sddm[529]: Setting default cursor
Mar 29 16:41:21 nitrogen sddm[529]: Running display setup script  "/usr/share/sddm/scripts/Xsetup"
Mar 29 16:41:21 nitrogen sddm[529]: Display server started.
Mar 29 16:41:21 nitrogen sddm[529]: Socket server starting...
Mar 29 16:41:21 nitrogen sddm[529]: Socket server started.
Mar 29 16:41:21 nitrogen sddm[529]: Loading theme configuration from ""
Mar 29 16:41:21 nitrogen sddm[529]: Greeter starting...
Mar 29 16:41:21 nitrogen sddm[682]: WARNING: CPU random generator seem to be failing, disabling hardware random number generation
Mar 29 16:41:21 nitrogen sddm[682]: WARNING: RDRND generated: 0xffffffff 0xffffffff 0xffffffff 0xffffffff
Mar 29 16:41:21 nitrogen sddm-helper[682]: [PAM] Starting...
Mar 29 16:41:21 nitrogen sddm-helper[682]: [PAM] Authenticating...
Mar 29 16:41:21 nitrogen sddm-helper[682]: [PAM] returning.
Mar 29 16:41:21 nitrogen sddm-helper[682]: pam_unix(sddm-greeter:session): session opened for user sddm(uid=975) by (uid=0)
[...]

PR #2205, opened 2026-07-17
fix(daemon): close server socket before calling deleteLater

### LINKED ISSUE #2207: SDDM breaks on logout due to AMD RDRAND bug
Under Debian, whether using the Debian SDDM package or compiling SDDM from source,
SDDM will simply break on logout, leaving only a black screen and a mouse cursor. To recover from this, SDDM must be restarted.

The log would complain that:

```
(EE) GREETER: Socket error:  "QLocalSocket::connectToServer: Invalid name"
(EE) GREETER: Cannot connect to the daemon - is it running?
```
because the socket in /tmp/ simply doesn't exist at the moment when it tries to open it, and even after.
This is abnormal because the daemon itself said previously that it had opened a server socket.

I made the following pull request that fixes the issue in my case:
https://github.com/sddm/sddm/pull/2205

It seems like this is a bug that appears only when amdgpu is present, it happens on different releases of Debian, but also stops happening on releases which are unable to use my video card (no/cannot load amdgpu driver)

The following issues are probably related:
#1535 
#2134 
(this idea comes from the fact that one user describes an AMD CPU with integrated graphics and the other one got `CPU random generator seem to be failing, disabling hardware random number generation` which is also a message I get, but which, from my testing, didn't prevent SDDM from working correctly when logging out on older releases)

ExplodingBottle:
Fixes: #2207

There is a problem happening on certain machines where logging out from your session will systematically result in a broken SDDM (there will only be a black screen and a mouse cursor), and the log would shout that:
```
(EE) GREETER: Socket error:  "QLocalSocket::connectToServer: Invalid name"
(EE) GREETER: Cannot connect to the daemon - is it running?
```

After some investigation, it appears that the socket would be totally inexistent at the time the greeter tries to connect, even though the daemon said that:

```
(II) DAEMON: Socket server starting...
(II) DAEMON: Socket server started.
```

There weren't any apparent problems in the code opening the socket. So the problem was where the socket was closed. The fact that the deletion of the server socket was delayed could cause it to happen after the creation of the newer server socket.

After making sure that the server socket was properly closed before opening a new one, the problem entirely disappeared (on my machine at least).

This is what this pull request does: call close on the server before calling deleteLater to make sure there won't be any dangling socket that could be used by the newer server socket.

I cannot say for certain, however it seems like those two issues are related to this, because of the description:
[sddm/sddm#1535](https://github.com/sddm/sddm/issues/1535)
[sddm/sddm#2134](https://github.com/sddm/sddm/issues/2134)

### DIFF
--- a/src/daemon/SocketServer.cpp
+++ b/src/daemon/SocketServer.cpp
@@ -82,6 +82,9 @@ namespace SDDM {
         // log message
         qDebug() << "Socket server stopping...";
 
+        // stop listening first
+        m_server->close();
+
         // delete server
         m_server->deleteLater();
         m_server = nullptr;


Vogtinator:
> The fact that the deletion of the server socket was delayed could cause it to happen after the creation of the newer server socket.

That should not be possible, as the old and new sockets don't share a name. If they do, that's a bug.

Do you have the full log?

ExplodingBottle:
> Do you have the full log?

I do have a full log, and I can tell that my sockets are always sddm-:0-ffVDPB
[sddm-debug.log](https://github.com/user-attachments/files/30167915/sddm-debug.log)

And thanks to the message you sent, I now know that what I thought was a bug in SDDM is in fact well related to an AMD microcode bug.

I also get the
```
WARNING: CPU random generator seem to be failing, disabling hardware random number generation
WARNING: RDRND generated: 0xffffffff 0xffffffff 0xffffffff 0xffffffff
```
which is why the generated name for the socket is always the same...

I read too fast the code for the generation of the socket name: I thought that it was made to be unique per machine, but it is in fact made to be random.

However I don't think that this pull request is rendered useless by this. It could instead be a workaround for people who never got a BIOS update to fix this.

Another fix could possibly be another way to generate a random socket name.

Vogtinator:
> which is why the generated name for the socket is always the same...

Ouch. That's bad, that will break other security mechanisms as well, e.g. ssh keys.

If the kernel already detected it, it should've disabled rdrand also for sddm.

ExplodingBottle:
It seems like the message about RDRND is not made by the kernel but by Qt5 itself:
https://github.com/qt/qtbase/blob/c68297f66533200c9ff626b056d096c78b4b94db/src/corelib/tools/qsimd.cpp#L701

On the side of the kernel, it doesn't seem to rely on RDRND (maybe only once at the system boot to init the seed), so values in `/dev/urandom` and `/dev/random` are effectively random.

The `getrandom` system call is unaffected either, and generates correctly random values too.

More testing reveals that the issue with `random_device` is only present when using `libstdc++` (the default) but not when using `libc++`

So what can be done is either:
- Merge the PR so even when the socket name is the same, it doesn't cause a crash
- Change generateName to use QT random number generation or other RNG functions (like getrandom) https://github.com/sddm/sddm/blob/d7f8b0211f63e91195fa4f6560678c3c107bf3d4/src/daemon/Utils.h#L28
- Change generateName to add /dev/urandom as constructor parameter for random_device (not ideal because cpp docs state that the constructor parameter is implementation independant, but will still work with libstdc++)
- Change SDDM to use libc++ (really not ideal, could have many unexpected drawbacks, and such a big change for a small bug...)

Another option could also be to not do anything on SDDM and report it as a bug to the maintainers of libstdc++

Vogtinator:
> > which is why the generated name for the socket is always the same...
> 
> Ouch. That's bad, that will break other security mechanisms as well, e.g. ssh keys.
> 
> If the kernel already detected it, it should've disabled rdrand also for sddm.

Apparently it does that only for the older AMD APU errata, not the Zen ones.

> On the side of the kernel, it doesn't seem to rely on RDRND (maybe only once at the system boot to init the seed), so values in /dev/urandom and /dev/random are effectively random.

FWICT you should see a message like `RDRAND is not reliable on this platform; disabling` and then the kernel won't use it for its entropy pool and thus `/dev/(u)random`

> Merge the PR so even when the socket name is the same, it doesn't cause a crash

That actually still shouldn't happen: The old socket server should not delete the new socket under any circumstances. I wonder how this could happen.

> Change generateName to use QT random number generation or other RNG functions (like getrandom)

Kind of this. It's important that the name isn't trivially guessable, or anyone with write access to `/tmp` could break SDDM.

(IMO the current design is broken anyway, it should just use `/run/sddm/`).

However, as long as `rdrand` on your system is broken, any application may break or become insecure without further notice. So I highly suggest to check for BIOS and AMD microcode updates.

ExplodingBottle:
> FWICT you should see a message like RDRAND is not reliable on this platform; disabling and then the kernel won't use it for its entropy pool and thus /dev/(u)random

I confirm I see this warning in dmesg now (I had searched for RDRND previously)

> So I highly suggest to check for BIOS and AMD microcode updates.

I do get BIOS updates, but none of them ever fixed this bug :) thanks HP

> Apparently it does that only for the older AMD APU errata, not the Zen ones.

And I have the Zen one...

> That actually still shouldn't happen: The old socket server should not delete the new socket under any circumstances. I wonder how this could happen.

This is how I understand it: the old server socket is present because it is not deleted yet when the new one is created. On the docs, they say that the deletion is not done directly, but is rather scheduled (it is also written in the docs that QLocalServer also closes connections when deleted). That should mean that the `/tmp/sddm:0-...` socket is not deleted instantly.

But, deleteLater is nonblocking, so while the previous socket is scheduled to be deleted (`sddm-:0-ffVDPB`), SDDM continues and creates a new one. But because of my unfortunate hardware configuration, the socket to be created happens to be `sddm-:0-ffVDPB` too.

To demonstrate this, here is a simple repository that you can clone:
https://github.com/ExplodingBottle/qt-socket-deletelater-demo

The same setup as SDDM is done to create a socket, close it and reopen it using the same name (`test-socket-sddm`). As you can see, by using deleteLater, at the end there is no socket in /tmp. However, if you uncomment the line with the close, the socket is there when checking /tmp

One could argue that my test is invalid: all of this runs before the Qt loop is started. However we can safely assume that during the SDDM logout process the Qt loop is not called enough to process the deletion of the socket (else there wouldn't be any bug, and no pull request there)

Vogtinator:
> This is how I understand it: the old server socket is present because it is not deleted yet when the new one is created.

Yeah, and creation of the new one should fail because the old one is still there.

ExplodingBottle:
> Yeah, and creation of the new one should fail because the old one is still there.

But it didn't crash in my test, with or without deleteLater.
However it did crash when using QLocalServer::NoOptions for the server options (able to create the first socket but not the one after)
Strange... Especially because there is that in the docs:

> Note: On Unix if the server crashes without closing listen will fail with AddressInUseError. To create a new server the file should be removed. On Windows two local servers can listen to the same pipe at the same time, but any connections will go to one of the server.

However, there is this:
https://qt-project.atlassian.net/browse/QTBUG-48635
which explains why there is no failure when creating the new socket.
Note that the bug has been there for a stunning 11 years...

However, in my testing, under Debian stretch I had no problem, but it did already include a buggy version of Qt (i.e. Qt 5.7.1)
That must mean that whether the socket creation fails or not has no impact on the crash on logout and that calling deleteLater is not sufficient to prevent the old socket from still being present at the time the new socket is created.

ExplodingBottle:
Hello,
I would like to request the removal of the `needinfo` tag because what's happening has been clarified and can be summarized as this TLDR:

```

CPU RDRAND bug -> Socket of the same name is created before the other one is deleted -> The older socket gets deleted and causes the newer one to be deleted too -> SDDM crash
                                           |
                                           + <- QTBUG-48635
                                           |
                                           v
                                          No error
```
Solution: make sure the older socket is deleted before the new one is created.

PR #2212, opened 2026-08-10
greeter: Don't quit when the last window closes

### LINKED ISSUE #2211: Greeter exits when the last monitor is unplugged
Unplugging the only monitor while the greeter is up kills the greeter, leaving
a blinking cursor on the console and no way back in without restarting sddm
from ssh or another VT. sddm.service stays active with NRestarts=0.

sddm 0.21.0 and current develop, Qt 6.11.1, KWin 6.7.4, amdgpu with a single
HDMI output:

```
sddm-greeter-qt6[55467]: There are no outputs - creating placeholder screen
sddm-greeter-qt6[55467]: Adding view for "" QRect(0,0 0x0)
sddm-helper-start-wayland[55448]: wayland greeter finished 0 QProcess::NormalExit
sddm-helper-start-wayland[55448]: wayland compositor finished 15 QProcess::NormalExit
sddm[55406]: Greeter stopped. SDDM::Auth::HELPER_SUCCESS
```

The Qt 5 greeter survives that behind Qt's fake screen and exits on replug
instead, when the real output replaces the placeholder. Same result.

GreeterApp creates a QQuickView per QScreen and destroys it again in
removeViewForScreen(), but setQuitOnLastWindowClosed(false) is never called. So
when the last screen goes away the window count reaches zero, Qt quits the
process with status 0 and the daemon reads that as HELPER_SUCCESS.

#1847 looks like the same bug. It was closed as a kwin issue, but in the log
there the greeter exits first and kwin is only stopped afterwards.

dnlbtz:
GreeterApp creates a QQuickView per QScreen and destroys it again in
removeViewForScreen(). With a single display, unplugging the monitor drops the
window count to zero and QGuiApplication's default quitOnLastWindowClosed makes
the greeter exit with status 0.

The daemon takes that for a successful exit and does not start a new greeter,
so the seat is left on a bare console while sddm.service is still active.

Tested with the Qt 5 and Qt 6 greeters on 0.21.0 and develop, the greeter now
survives unplug and replug.

Fixes #2211

### DIFF
--- a/src/greeter/GreeterApp.cpp
+++ b/src/greeter/GreeterApp.cpp
@@ -361,6 +361,10 @@ int main(int argc, char **argv)
         qputenv("QT_IM_MODULE", inputMethod.toLocal8Bit());
 
     QGuiApplication app(argc, argv);
+
+    // Don't quit when the last screen and thus the last view disappears
+    app.setQuitOnLastWindowClosed(false);
+
     SDDM::SignalHandler s;
     QObject::connect(&s, &SDDM::SignalHandler::sigtermReceived, &app, [] {
         QCoreApplication::instance()->exit(-1);


Vogtinator (approved):
I think this is fine: There is no code in the sddm daemon for handling the state that the greeter just quits on its own (no matter the seat), it's not supposed to happen. If all monitors for a seat != seat0 disappear, that should probably still be handled by the seat management and not be triggered by the greeter.

dnlbtz:
The two rawhide jobs fail in `dnf builddep`, before anything is compiled. `fedora-cisco-openh264` is still serving an fc45-signed openh264 while rawhide's keyring has moved to the Fedora 46 key, so the signature check fails with "Import of the key didn't help, wrong key?". It comes in as a hard dependency by way of ImageMagick -> libheif -> libavcodec-free, so it isn't something the build can just do without.

The same two jobs fail the same way on develop at 52efe99e (run 31901958029, 15 August), so it isn't coming from this change. The other eleven jobs are green, including both qt6 builds.

dnlbtz:
This is tracked on the Fedora side as releng ticket 13466 (https://forge.fedoraproject.org/releng/tickets/issues/13466): after each rawhide branch the openh264 RPMs have to be rebuilt and republished by Cisco, and until that happens the rawhide repo is deliberately pointed at the older f45 one, so the packages it serves are signed with the wrong key.

The f46 build has been uploaded and Cisco confirmed receipt on 17 August but hasn't published it yet, so the jobs should go green again on their own with no change needed here.

Issue #1418, opened 2021-06-10
last AD user is not remembered


majojoe:
All Computers in our Company are connected to a domain like this is the case in most Enterprise environments. This is also true for our Kubuntu Computers. But a very annoying thing is, that the last AD user is not remembered when using sddm. One must click onto the Search Button and enter the Username, followed by the password every time one needs to login.
Wouldn't it be possible that the last AD user is also remembered in standard Themes?
For now we have helped ourselfs using a very old text only theme.

Btw: Another issue I have noticed when logging in especially in virtual machines: When logging in the screen with sddm the screen keeps often black until the screen of the VM is resized or otherwise changed, showing then the normal screen after the login. One would expect that sddm would show up this screen automatically after the login animation. The login animation is also not shown then.

x09:
I think there is no solution for this problem.
This problem has existed for many years - there is still no solution.
MaximumUid, RememberLastUsers - do not work.

plop28:
I also have this problem. Domain UID are quite large in our case. the only solution I found is to use elarun sddm theme, it works with RememberLastUsers

Vogtinator:
Sounds like a theme bug. Which one is broken?

majojoe:
I have noticed this issue with the standard Kubuntu Theme "Breeze". I also tried several other themes (about halve a dozen). None of them worked  - but to be honest I cannot remember which ones I have tried in - sorry. 
To mitigate the problem I use the theme "Expose Blue" which is text based and remembers just the text until next login.
@plop28: Thank you for the hint. Which distribution do you use? In Kubuntu 22.04 I can find 3 themes which are elarun based, but do not come with the distribution.

plop28:
Yes, same problem on all "modern" themes I tried. 

@majojoe  Only text based (with no user dropdown list or avatar list) are working like you said.
I'm on Debian 12.4, and i have installed elarun theme via apt : sddm-theme-debian-elarun and sddm-theme-elarun

Issue #826, opened 2017-06-08
Remote usernames not displayed when working with sssd.


sedros:
Hi.
Recently I switched to LDAP and SSSD on my home computers. With LDAP only everything was ok, but after enabling SSSD SDDM stopped displaying remote usernames. User can login but has to input the username by hand rather than picking it from the list.

Distro: Gentoo
x11-misc/sddm-0.14.0-r3::gentoo was built with the following:
USE="consolekit pam -elogind -systemd" ABI_X86="(64)"

sys-auth/sssd-1.13.1::gentoo was built with the following:
USE="acl locator manpages nls ssh sudo -augeas -autofs -netlink -nfsv4 -python -samba (-selinux) -test" ABI_X86="(64) -32 (-x32)" PYTHON_TARGETS="python2_7 python3_4"

Best regards

dirks:
I see the same with sddm on kde neon. I guess this is due to a high uid. My users come from an AD and have uids in the hundreds of millions range, which in my experience is quite normal when [mapping AD SIDs to uids](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Windows_Integration_Guide/sssd-ad-integration.html) with sssd.

There is the configuration option MaximumUid in the [Users] part of sddm.conf. For me setting this to 999,999,999 did reveal uid 65534 (nobody) on the login screen, but did not reveal the users from the AD. I guess there is an explicit or implicit cutoff value somewhere. In the KDE GUI tool for sddm `system settings -> startup and shutdown -> login screen (sddm) -> advanced` the maximum uid value is displayed as 999,999 instead of 999,999,999.

Just to clarify, of course sddm should not display all users that are available from an AD/LDAP by default. But showing those that logged in last on that machine is a valid use case, e.g. if you login to a laptop (see lightdm, windows and probably others). This is somewhat related to #559

sedros:
I have users with UID of about 10000 which is well in configured sddm's range of 1000 to 65000.

dspadial:
Hi.
Thank them for the good work they do.
In my case, the same thing happens, and I can not find the solution.

System: "Linux Mint 18.2 Sonya"
Sssd-ad: "version 1.13.4-1ubuntu1"
Theme: "breeze"

Both the ID mapping range in sssd-ad and the sddm MaximumUid is set to work between 2000 and 60000.

The problem is that if you do sssd-ad work enumerating suarios, it works but all AD users are shown (over 1000), not being appropriate.

[...]

PR #2216, opened 2026-08-14
Greeter: Always try to load last user data

TimoTegtmeier:
When using SSSD the last username is remembered, but the user cannot be found when using a recent version (as enumeration has been disabled by default and a re-compile would be necessary) and so the remember function does not work as expected in such a scenario.

To overcome that limitation, I changed the code to always try to load the remembered user data. Under very specific conditions, this could also happen previously, but the solutions is a bit more generalized.

This PR fixes bug #1418
See also: #826

### DIFF
--- a/src/greeter/UserModel.cpp
+++ b/src/greeter/UserModel.cpp
@@ -108,16 +108,14 @@ namespace SDDM {
             }
         }
 
-        // Always try to add the last user if not already found.
-        if (!lastUserFound) {
-            struct passwd *lastUserData;
-            if(!lastUserFound && (lastUserData = getpwnam(qPrintable(lastUser())))) {
-                d->users << UserPtr(new User(lastUserData, iconURI));
-                lastUserFound = true;
-            }
-        }
-
         endpwent();
+
+        // Always try to add the last user if not already found
+        if (!lastUser().isEmpty() && !lastUserFound) {
+            struct passwd *lastUserData;
+            if((lastUserData = getpwnam(qPrintable(lastUser()))))
+                d->users << UserPtr(new User(lastUserData, iconURI));
+        }
 
         // sort users by username
         std::sort(d->users.begin(), d->users.end(), [&](const UserPtr &u1, const UserPtr &u2) { return u1->name < u2->name; });
--- a/src/greeter/UserModel.cpp
+++ b/src/greeter/UserModel.cpp
@@ -103,13 +103,17 @@ namespace SDDM {
                 lastUserFound = true;
 
             if (!needAllUsers && d->users.count() > mainConfig.Theme.DisableAvatarsThreshold.get()) {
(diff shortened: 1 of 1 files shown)


src/greeter/UserModel.cpp
```diff
@@ -103,16 +103,20 @@ namespace SDDM {
...
-
                 d->containsAllUsers = false;
                 break;
             }
         }
 
+        // Always try to add the last user if not already found.
+        if (!lastUserFound) {
+            struct passwd *lastUserData;
+            if(!lastUserFound && (lastUserData = getpwnam(qPrintable(lastUser())))) {
```
Vogtinator:
```suggestion
            if(lastUserData = getpwnam(qPrintable(lastUser()))) {
```

src/greeter/UserModel.cpp
```diff
@@ -103,16 +103,20 @@ namespace SDDM {
...
-                if(!lastUserFound && (lastUserData = getpwnam(qPrintable(lastUser()))))
-                    d->users << UserPtr(new User(lastUserData, themeDefaultFace));
-
                 d->containsAllUsers = false;
                 break;
             }
         }
 
+        // Always try to add the last user if not already found.
+        if (!lastUserFound) {
```
Vogtinator: Let's use the opportunity and fix a long standing issue too:

```suggestion
        if (!lastUser().isEmpty() && !lastUserFound) {
```

src/greeter/UserModel.cpp
```diff
@@ -103,16 +103,20 @@ namespace SDDM {
...
+            struct passwd *lastUserData;
+            if(!lastUserFound && (lastUserData = getpwnam(qPrintable(lastUser())))) {
+                d->users << UserPtr(new User(lastUserData, iconURI));
+                lastUserFound = true;
```
Vogtinator: Never read

src/greeter/UserModel.cpp
```diff
@@ -103,16 +103,20 @@ namespace SDDM {
...
+            }
+        }
+
         endpwent();
```
Vogtinator: Move this before the moved if block

Vogtinator (changes requested):
i.e. getpwent is unable to iterate through all users?

Some minor issues, otherwise LGTM.

Vogtinator:
One possible issue is that previously with disabled enumeration, Themes won't see any users and fall back to a username input field. With this, the user model would contain lastUser only and possibly only show that one. Needs manual adjustment of the avatar threshold.

TimoTegtmeier:
@Vogtinator I updated the code according to your suggestions.
Honestly, I'm not sure regarding your objection though, but you might be right.

Vogtinator (approved):

